PR #1, opened 2019-02-26
Summary: simple plugin which got out of hand.

rustyrussell:
We should add pretty-printing to lightning-cli (with an option to disable), and remove it from lightningd.  That would be faster, too, and avoid messy plugin output.

### DIFF
--- a/summary/summary.py
+++ b/summary/summary.py
@@ -100,17 +100,18 @@ def summary(plugin):
                 to_them = Millisatoshi(0)
             avail_in += to_them
             reply['num_channels'] += 1
-            chans.append((c['total_msat'], to_us, to_them, p['id'], c['private']))
+            chans.append((c['total_msat'], to_us, to_them, p['id'], c['private'], p['connected']))
 
     reply['avail_out'] = avail_out.to_btc_str()
     reply['avail_in'] = avail_in.to_btc_str()
 
     if plugin.fiat_per_btc:
-        reply['utxo_amount'] += ' = {}'.format(to_fiatstr(utxo_amount))
-        reply['avail_out'] += ' = {}'.format(to_fiatstr(avail_out))
-        reply['avail_in'] += ' = '.format(to_fiatstr(avail_in))
+        reply['utxo_amount'] += ' ({})'.format(to_fiatstr(utxo_amount))
+        reply['avail_out'] += ' ({})'.format(to_fiatstr(avail_out))
+        reply['avail_in'] += ' ({})'.format(to_fiatstr(avail_in))
 
     if chans != []:
+        reply['channels_key'] = 'P=private O=offline'
         reply['channels'] = []
         biggest = max(max(int(c[1]), int(c[2])) for c in chans)
         for c in chans:
@@ -138,13 +139,20 @@ def summary(plugin):
                 right = "{:23}".format(draw.bar * (their_len - 1) + draw.right)
 
             s = left + mid + right
+
+            extra = ''
+            if c[4]:
+                extra += 'P'
+            if not c[5]:
+                extra += 'O'
+            if extra != '':
+                s += '({})'.format(extra)
+
             node = plugin.rpc.listnodes(c[3])['nodes']
             if len(node) != 0:
                 s += ':' + node[0]['alias']
             else:
                 s += ':' + c[3][0:32]
-            if c[4]:
-                s += ' (priv)'
             reply['channels'].append(s)
 
     return reply
--- a/summary/summary.py
+++ b/summary/summary.py
@@ -54,6 +54,7 @@ def summary(plugin):
 
     avail_out = Millisatoshi(0)
     avail_in = Millisatoshi(0)
+    chans = []
     reply['num_channels'] = 0
     reply['num_connected'] = 0
     reply['num_gossipers'] = info['num_peers']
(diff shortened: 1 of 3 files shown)


summary/summary.py
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env python3
+from lightning import Plugin, Millisatoshi
+import json
+import requests
```
cdecker: The dependency on `requests` should be in a `requirements.txt` file. Probably also should list `pylightning` as a dependency.

summary/summary.py
```diff
@@ -0,0 +1,120 @@
...
+
+plugin = Plugin(autopatch=True)
+
+
+class PriceThread(threading.Thread):
+    def run(self):
+        try:
+            r = requests.get('https://apiv2.bitcoinaverage.com/convert/global'
+                             '?from=BTC&to={}&amount=1'.format(plugin.currency))
+            plugin.fiat_per_btc = json.loads(r.content)['price']
```
cdecker: Since you're wrapping the thread in its own class, it might be nice to have `plugin` as an instance field, instead of referring to a global. Alternatively you can forgo the class and run a function in a thread using

```python
t = Thread(target=my_function, args=[plugin])
t.daemon = True  # Make sure to terminate thread if main-thread exits
t.start()
```

summary/summary.py
```diff
@@ -0,0 +1,120 @@
...
+        except Exception:
+            pass
+        # Six hours is more than often enough for polling
+        time.sleep(6*3600)
```
cdecker: Wait, we don't get direct integration with the ICE datafeed? I'm disappointed :wink:

summary/summary.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def to_fiatstr(msat: Millisatoshi):
+    return "{}{:.2f}".format(plugin.currency_prefix,
+                              int(msat) / 10**11 * plugin.fiat_per_btc)
```
cdecker: Should probably be `float(msat)` otherwise the type coercion might bite us.

summary/summary.py
```diff
@@ -0,0 +1,120 @@
...
+    if info['network'] != 'bitcoin':
+        reply['network'] = info['network'].upper()
+
+    if not plugin.my_address:
+        reply['warning_no_address'] = "NO PUBLIC ADDRESSES"
+    else:
+        reply['my_address'] = plugin.my_address
+
+    utxo_amount = Millisatoshi(0)
+    reply['num_utxos'] = 0
```
cdecker: List comprehension might be nice here, and then `sum` and `len` of that list :-)

summary/summary.py
```diff
@@ -0,0 +1,120 @@
...
+            if c['their_reserve_msat'] < to_them_msat:
+                avail_in += to_them_msat - c['their_reserve_msat']
+            reply['num_channels'] += 1
+    reply['avail_out'] = avail_out.to_btc_str()
+    reply['avail_in'] = avail_in.to_btc_str()
+
+    if plugin.fiat_per_btc:
+        reply['utxo_amount'] += ' = ' + to_fiatstr(utxo_amount)
+        reply['avail_out'] += ' = ' + to_fiatstr(avail_out)
+        reply['avail_in'] += ' = ' + to_fiatstr(avail_in)
```
cdecker: Prefer formatting strings whenever possible, otherwise we create a lot of intermediate strings (not an issue here, just saying :-p)

summary/summary.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+@plugin.init()
+def init(options, configuration, plugin):
+    plugin.currency = options['summary-currency']
+    plugin.currency_prefix = options['summary-currency-prefix']
+    info = plugin.rpc.getinfo()
+
+    # Try to grab conversion price
+    PriceThread().start()
```
cdecker: This thread will keep the process running despite the main thread exiting (resulting in these sticking around for longer than necessary).

```python
t = PriceThread()
t.daemon = True
t.start()
```

Alternatively you can add a constructor `PriceThread.__init__` that sets the `.daemon` field. It is important that this is set before calling `.start()` since that field cannot be changed once started.

summary/summary.py
rustyrussell: msat contains an __int__ function so I think this is right.  fiat_per_btc is float?

summary/summary.py
rustyrussell: I guessed >=0.0.6 for pylightning, but I guess that's OK?

requests>=2.0.0 seemed reasonable, too.

summary/summary.py
rustyrussell: Stackoverflow said make it its own class if it's refreshing the same timer, so I copied.  WIll make it daemon though.

summary/summary.py
rustyrussell: We should ask, yes!

summary/summary.py
rustyrussell: Thanks, done.

summary/summary.py
rustyrussell: Hmm, not sure it's simpler, but it is shorter...

summary/summary.py
rustyrussell: Hmm, `+= " = {}".format(to_fiatstr(utxo_amount))` seems a bit marginal, and I like the += for clarity.  I've changed though...

rustyrussell:
OK, now I've done some more neatening, and handle with and without UTF-8.  Let's stop playing with this :)

cdecker:
ACK https://github.com/lightningd/plugins/pull/1/commits/ff67b268417f5956fad4a585f865f985ce16208e

PR #3, opened 2019-04-08
raw response fix

Johnstedt:
Not sure if this is a universal issue, price thread crashes while setting fiat_per_btc  as r.content is raw data. Running: packaging==19.0, requests==2.9.1 and pylightning==0.0.7.

### DIFF
--- a/summary/summary.py
+++ b/summary/summary.py
@@ -36,7 +36,7 @@ class PriceThread(threading.Thread):
         try:
             r = requests.get('https://apiv2.bitcoinaverage.com/convert/global'
                              '?from=BTC&to={}&amount=1'.format(plugin.currency))
-            plugin.fiat_per_btc = json.loads(r.content)['price']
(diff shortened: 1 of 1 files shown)


summary/summary.py
```diff
@@ -36,7 +36,7 @@ def run(self):
         try:
             r = requests.get('https://apiv2.bitcoinaverage.com/convert/global'
                              '?from=BTC&to={}&amount=1'.format(plugin.currency))
-            plugin.fiat_per_btc = json.loads(r.content)['price']
+            plugin.fiat_per_btc = json.loads(r.content.decode())['price']
```
cdecker:
```suggestion
            plugin.fiat_per_btc = json.loads(r.content.decode('UTF-8'))['price']
```

It's probably better to mention which encoding should be used here. UTF-8 is what we use everywhere else, and supports a superset of most other encodings. More importantly it also includes the EUR symbol :wink:

PR #6, opened 2019-04-16
rebalance plugin in python

gallizoltan:
- this plugin helps to move some liquidity between your channels using circular payments
- previous conversations here: ElementsProject/lightning#2567


cdecker:
ACK https://github.com/lightningd/plugins/pull/6/commits/f22615ed658573414fd50021c00cb6a4f0b0afec

PR #7, opened 2019-04-20
created git module for c-lightning-api plugin

nettijoe96:
submodule for c-lightning-api

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "sitzprobe"]
 	path = sitzprobe
 	url = https://github.com/niftynei/sitzprobe
+[submodule "c-lightning-graphql"]
+	path = c-lightning-graphql
+	url = https://github.com/nettijoe96/c-lightning-graphql.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "sitzprobe"]
 	path = sitzprobe
 	url = https://github.com/niftynei/sitzprobe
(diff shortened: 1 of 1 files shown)


cdecker:
Hi @nettijoe96 thanks for adding the project as a submodule. The name is a bit confusing imho since it is first and foremost a graphql adaptor, so maybe we should call it `graphql` or something to that end?

nettijoe96:
just updated the name

cdecker:
Why the `c-lightning` prefix? These are all `c-lightning` plugins :wink:

nettijoe96:
in case someone comes across the repo outside of this plugin repo

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "sitzprobe"]
 	path = sitzprobe
 	url = https://github.com/niftynei/sitzprobe
+[submodule "c-lightning-graphql"]
```
cdecker:
```suggestion
[submodule "graphql"]
```

.gitmodules
```diff
@@ -1,3 +1,6 @@
...
+	path = c-lightning-graphql
```
cdecker:
```suggestion
	path = graphql
```

nettijoe96:
wow that is cool! didn't know you could do that!

cdecker:
ACK https://github.com/lightningd/plugins/pull/7/commits/3ff173b405355995cf5c31b83b003b5fdafa29a9

PR #8, opened 2019-04-23
feat: add scid to summary output

m-schmoock:
Improves some minor output issues:

 - Adds short channel id, so it can be copy and pasted for rebalancing
 - renames `channels_key` to `channels_flags`
 - replace `=` with `:` in `channels_flags`, looks better in -H or JSON output.
 - changes flags round rackets (PO) to squace brackets [PO]
 - Always outputs flags even if not set `[__]` so things are always aligned
 - Removes the colon ':' and always prints a space
   this also helps when copy and pasting in a terminal

TLDR: Output will now look like this (with correct special char alignment which GitHub mess up):
```
channels_flags=P:private O:offline
channels=

    ├──────────────────┼───┤                    1511185x45x0   [__] testnet1-lnd.zaphq.io
├──────────────────────╢                        1511189x47x0   [P_] CoinGate
 ├─────────────────────┼┤                       1511205x23x0   [_O] lnd.fun
```

### DIFF
--- a/summary/summary.py
+++ b/summary/summary.py
@@ -101,7 +101,7 @@ def summary(plugin):
                 to_them = Millisatoshi(0)
             avail_in += to_them
             reply['num_channels'] += 1
-            chans.append((c['total_msat'], to_us, to_them, p['id'], c['private'], p['connected']))
+            chans.append((c['total_msat'], to_us, to_them, p['id'], c['private'], p['connected'], c['short_channel_id']))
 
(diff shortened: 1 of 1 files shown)


m-schmoock:
Since its only output formatting I did it all in one commit.
If you want it separately, I can forcepush update the PR.

regargs,
Michael

cdecker:
ACK c58cbd6e68309d5c08503fc2cd441b564c11d3f2

PR #19, opened 2019-05-08
probe: Actually add a sensible return value to the probe method

cdecker:
It was returning `null` and we'd have to check the logs or the DB for the
result. This makes it just a bit easier to see what happens.


m-schmoock:
ACK 14a9227

PR #20, opened 2019-05-08
autoreload: Add an autoreload plugin to restart plugins on change

cdecker:
A simple plugin that can be used to speed plugin development by having the
plugin under development be reloaded whenever its executable changes, or when
triggered automatically.

### DIFF
--- a/autoreload/README.md
+++ b/autoreload/README.md
@@ -1,0 +1,85 @@
+# Autoreload Plugin
+
+> You write the code, we do the rest!
+
+The autoreload plugin aims to help plugin developers to reload the plugin
+inside of `lightningd` without having to restart the entire daemon. It watches
+the executable file and will restart the plugin when a change is detected.
+
+The plugin adds the following command line options:
+
+ - `--autoreload-plugin=path/to/plugin.py`: specifies which plugin you'd like
+   `autoreload` to manage. The syntax was chosen so you can just prefix any
+   existing `--plugin` with `autoreload` and start hacking.
+   
+The plugin adds the following RPC methods:
+
+ - `autoreload-restart`: triggers a manual restart of the plugin under
+   development, useful if you want to reload because of a dependency change or
+   you have a script that knows when a reload is needed. We will still
+   autoreload whenever we detect a change in the executable.
+
+
+So in order to have a plugin `path/to/plugin.py` autoreload on changes you'll
+need to call `lightningd` with the following arguments:
+
+```bash
+lightningd --plugin=path/to/autoreload.py --autoreload-plugin=path/to/plugin.py
+```
+
+The first argument loads the `autoreload` plugin, and the second argument
+tells `autoreload` to load, watch and restart the plugin under development.
+
+## Install
+
+This plugin relies on the `pylightning` and the `psutil` libraries. You should
+be able to install these dependencies with `pip`:
+
+```bash
+pip3 install -r requirements.txt
+```
+
+You might need to also specify the `--user` command line flag depending on
+your environment.
+
+## How does it work?
+
+In order to hide the restarts from `lightningd` the autoreload plugin will
+insert itself between `lightningd` and the plugin that is to be reloaded,
+acting as an almost transparent proxy between the two. The only exception to
+this are the two calls to register and initialize the plugin:
+
+ - `getmanifest` is captured and the autoreload RPC methods and options are
+   injected, so `lightningd` knows about them and tells us when we get
+   initialized.
+ - `init` is captured and the options we returned above are stripped again,
+   otherwise we might upset the plugin under development. We also cache it, so
+   we can tell the plugin under development again when we restart it.
+
+Upon restarting we call `getmanifest` on the plugin under development just to
+be safe (though you wouldn't do any initialization before being told so with
+`init` would you? :wink:) and we then call `init` with the parameters we
+cached when we were initialized ourselves. After this initialization dance is
+complete, we will simply forward any calls directly to the plugin under
+development and forward any output it produces to `lightningd`.
+
+We watch the modification time of the executable file and automatically
+restart the plugin should it change. You can trigger this by changing the file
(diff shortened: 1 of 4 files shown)


autoreload/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+We watch the modification time of the executable file and automatically
+restart the plugin should it change. You can trigger this by changing the file
+or even just `touch`ing it on the filesystem. You can also trigger a manual
+restart using the `autoreload-restart` RPC method:
+
+```bash
+lightning-cli autoreload-restart
+```
+
```
m-schmoock: Maybe add a dependencies section here to point out the psutil dependency.
It is rather standard, but I did not have it installed (yet).

```
## Dependencies
This plugin uses the `psutil` python library (https://psutil.readthedocs.io). It is pretty common, you can install it with `pip` or your system's packet manager.
```

autoreload/autoreload.py
```diff
@@ -0,0 +1,227 @@
...
+        """
+        raw = {
+            'jsonrpc': '2.0',
+            'method': request.method,
+            'params': request.params,
+            'id': request.id,
+        }
+
+        self.init = raw
+        del self.init['params']['plugin']
```
m-schmoock: You need to check of key `'plugin'` and `'request'`. Otherwise it didnt work (for me):

```Python
        if 'plugin' in self.init['params']:
            del self.init['params']['plugin']
        if 'request' in self.init['params']:
            del self.init['params']['request']
```

Note: `my self.init['params']` looked like this:
```
{'options': {'summary-currency': 'USD', 'summary-currency-prefix': 'USD $', 'autoreload-plugin': '/home/will/projects/lightning.plugins.git/summary/summary.py'}, 'configuration': {'lightning-dir': '/home/will/.lightning', 'rpc-file': 'lightning-rpc'}
```

autoreload/autoreload.py
```diff
@@ -0,0 +1,227 @@
...
+            'params': request.params,
+        }
+        self.send(raw)
+
+
+@plugin.init()
+def init(options, configuration, plugin, request):
+    if options['autoreload-plugin'] in ['null', None]:
+        print("Cannot run the autoreload plugin on its own, please specify --autoreload-plugin")
+        plugin.rpc.stop()
```
m-schmoock: I think you need to return here and not fall thru and start the watch_thread below, right?

autoreload/autoreload.py
cdecker: Right, I was working with my parameter binding patchset already applied, which behaves a bit differently. Made it conditional.

autoreload/autoreload.py
cdecker: Good point, by the time `stop()` returns we're already dead, but always good to make sure we don't do useless work.

autoreload/README.md
cdecker: Added the appropriate comment in an Install section. Since we require `pylightning` anyway, I don't think the system package manager will really work for now, but `pip install --user` should work if not in a `virtualenv` and without `--user` will work inside a `virtualenv` :wink:

m-schmoock:
@cdecker I tried around a bit with various plugin errors, here is what I observed:

* the autoreload can handle runtime errors within `init` just fine. 
* the autoreoad cannot handle syntax errors currently, it will raise a syntax error one time in the logs, but wont restart the plugin anymore when changed afterwards

```
SyntaxError: invalid syntax
2019-05-08T18:28:38.162Z plugin-autoreload.py Killing plugin: JSON-RPC message does not contain "jsonrpc" field
```

cdecker:
> the autoreloader cannot handle syntax errors currently, it will raise a syntax error one time in the logs, but wont restart the plugin anymore when changed afterwards

Yes, I might be able to catch those non-JSON messages (which happen before we monkey patch `stdout`) but I'm wondering how much of a hack that'd be.

cdecker:
Ok, found the reason we were being killed. While I expected it to be the error message being spewed onto `stdout` it turns out to be our own error when attempting to read an object from the plugin, but failing. We'd then be passing that error up, but since it isn't a valid JSON-RPC error it'd cause us to get killed.

I also addressed the other issues, so this should be much better now. I'll autosquash once we have preliminary acks in place :-)

cdecker:
How would you surface that error? It's an automated restart, so no JSON-RPC call that we could return the error to. Logs seems like the only place we can currently do anything. One thing I'm considering for future improvements is that we should fail RPC calls if they arrive while the child plugin is not running since otherwise they'll hang indefinitely. But one step at a time :-)

m-schmoock:
Since autorestart Plugin acts as proxy, we can return/raise the Syntax error on any registered Plugin rpc methods, right?

Its hacky, but in the end autorestart is a development tool. Telling the dev in a quick but maybe unclean way is still helpful.

cdecker:
Hm, I'd rather return a generic error and have them check the output of `lightningd`. We currently don't grab a plugin's `stderr` exactly so that neither the autoreload nor `lightningd` can swallow any runtime errors, and I would like to keep it that way, otherwise behavior changes substantially between running under `autoreload` or running directly.

But let's discuss that in a separate issue :wink:

m-schmoock:
Okay, if there are no other issues, we can merge this. Can I?  :D

cdecker:
Sure, go ahead :-)

cdecker:
Squashed and rebased :-)

m-schmoock:
Hm, Github doesnt like me yet. Do you know this:

- On mobile it says I can merge and gives me a 'merge' button, but fails with;
> Pull requests that don’t have one approved review, with no changes requested can’t be merged on a phone.

- on desktop view it tells me im not allowed, and dies not render the 'merge' button.

m-schmoock:
@cdecker It doesnt work for me, you have to merge. On mobile he keeps preventing it because "it doesnt have a recent approved review". On Desktop it tells me im not authorized. Mergin is Blocked and restricted to authorized users °_°

PR #22, opened 2019-05-09
drain and fill channel plugin

m-schmoock:
This plugin offers three new commands: `drain`, `fill` and `setbalance`.
It works similar to the `rebalance` plugin but with some key differences:

- Does not need a second `scid` parameter but figures out the other one by routing.
- Can have chunked payments that increases routing probability but can result in partial completion. If `cunks` is not set, number of required chunks is autodetected.
- Does not consume or calculate msat amounts, but uses percentages. Default `100` for `drain` and `fill` and `50` for `setbalance`.

Note: If draining 100%, the plugin guesses the correct HTLC fee by try and error. For new servers that report HTLC fee via exception (https://github.com/ElementsProject/lightning/pull/2691) it takes the exact amount.

Issues:
- Can run into "route not found" errors afters several "drain"s on multiple channels. Reason yet unknown.

### DIFF
--- a/drain/README.md
+++ b/drain/README.md
@@ -45,27 +45,47 @@ installation as executables within these directories will be loaded as plugins.
 
 ## Usage
 
-Once the plugin is active you can use it to `drain` liquidity on one of your
-channels by:
+Once the plugin is active you can use it to `drain` a given percentage of
+liquidity (default 100%) on one of your channels by:
 
 ```
 lightning-cli drain scid [percentage] [chunks] [maxfeepercent] [retry_for] [exemptfee]
 ```
 
 The plugin has also a `fill` command that does excactly the opposite. You
-can use it to fill up your side of the channel:
+can use it to fill up a given percentage of liquidity (default 100%) on your
+side of a channel:
 
 ```
 lightning-cli fill scid [percentage] [chunks] [maxfeepercent] [retry_for] [exemptfee]
 ```
 
+Another useful command is the `setbalance` that will fill up or drain your side
+of a channels balance to a given total percentage (default 50%). It will do all
+the math for you, so that you do not need to care for current channel balance:
+
+```
+lightning-cli setbalance scid [percentage] [chunks] [maxfeepercent] [retry_for] [exemptfee]
+```
+
+
+
 ### Parameters
 
 - The `scid` is the short_channel_id of the channel to drain or fill.
-- OPTIONAL: The `percentage` paramter tells the plugin how much of the channel
-  capacity should be drained/filled. Default: 100. Likewise a 'drain 50' will
-  balance a channel that is above 50% of its capacity same as 'fill 50' will
-  balance a channel that is below 50% of its capacity.
+- OPTIONAL: The `percentage` parameter tells the plugin how much of a channels
+  total capacity should be `drain`ed or `fill`ed (default: 100%).
+  For the `setbalance` command this sets the target percentage and it defaults
+  to 50% in this case. Resulting over or under capacity will be limited
+  to 100% (full) or 0% (empty) automatically. Examples:
+  - A 'drain 10' will send out 10% of the channels total (not current) capacity.
+  - A 'drain 100' will send out 100% of the channels total capacity, the channel
+  will be empty after this.
+  - A 'fill 10' will increase your side of a channels balance by 10% from total.
+  - A 'fill 100' will increase will fill up your channel.
+  - A 'setbalance' will balance out a channel.
+  - A 'setbalance 70' will bring a channel in a state where your side will hold
+    70% of total capacity.
 - OPTIONAL: The `chunks` parameter tells the plugin to try breaking down the
   payment into several smaller ones. In this case it may happen that the
   operation will only be partially completed. The parameters value is the
@@ -79,13 +99,6 @@ lightning-cli fill scid [percentage] [chunks] [maxfeepercent] [retry_for] [exemp
   allows the `maxfeepercent` check to be skipped on fees that are smaller than
   exemptfee (default: 5000 millisatoshi).
 
-NOTE: The `percentage` relates to resulting target capacity. A 'drain 1%' will
-      not try to send 1% of channel balance away, but try to reduce its balance
-      to 'just' 99%, if channel is above 99% total capacity.
-
-NOTE: Automatically converting a 'drain 70' to a 'fill 30' if the channel is
-      already below 30% capacity is not implemented yet.
-
 
 ## Tips and Tricks
 
@@ -98,3 +111,11 @@ NOTE: Automatically converting a 'drain 70' to a 'fill 30' if the channel is
   side has funds; but the protocol ensures that there is always progress toward
   meeting this reserve, and once met, [it is maintained.](https://github.com/lightningnetwork/lightning-rfc/blob/master/02-peer-protocol.md#rationale)
   Therefore you cannot drain or fill a channel to be completely empty or full.
+
+
+## TODOs
+ - fix: occasionally strange route errors. maybe try increasing chunks on route errors.
+ - fix: sometimes, if we ran in error, not all chunk results are returned, i.e. [2/4, error] but not 1/4.
+        This maybe relate to the waitsendpay timed out race condition. Can be solved by a new plugin hook.
+ - feat: set HTLC_FEE MIN/MAX/STP by feerate
+ - chore: reconsider use of listchannels
--- a/drain/README.md
+++ b/drain/README.md
@@ -1,0 +1,100 @@
+# Drain and Fill Plugin
+
+This plugin drains or fills up the capacity of one of your channel using
+circular payments to yourself. This can be useful for:
+
+- liquidity management
+- cleaning up or reducing channels before closing
+    - keeping capacity: pushing remaining balance into other lightning channels
+    - reducing capacity: filling up a channel before closing to reduce capacity
+- accumulating dust before closing multiple channels
+- ...
+
+## Installation
+
+This plugin relies on the `pylightning` library. As with most plugins you should
+be able to install dependencies with `pip`:
+
+```bash
+pip3 install -r requirements.txt
+```
+
+You might need to also specify the `--user` command line flag depending on
+your environment. If you dont want this and your plugin only uses `pylightning`
+as the only dependency, you can also start `lightningd` with the `PYTHONPATH`
+environment variable to the `pylightning` package of your `lightningd`
+installation. For example:
+
+```
+PYTHONPATH=/path/to/lightning.git/contrib/pylightning lightningd --plugin=...
+```
+
+## Startup
+
+The plugin can be started with `lightningd` by adding the `--plugin` option.
+Remember that all `lightningd` plugins have to have executable permissions.
+
+```
+lightningd --plugin=/path/to/plugin/drain.py
+```
+
+Alternatively, you can also symlink or copy the plugins executable to the
+`.lightning/plugins` folder or the `plugins` folder of your c-lightning
+installation as executables within these directories will be loaded as plugins.
+
+
+## Usage
+
+Once the plugin is active you can use it to `drain` liquidity on one of your
+channels by:
+
+```
+lightning-cli drain scid [percentage] [chunks] [maxfeepercent] [retry_for] [exemptfee]
+```
+
+The plugin has also a `fill` command that does excactly the opposite. You
+can use it to fill up your side of the channel:
+
+```
+lightning-cli fill scid [percentage] [chunks] [maxfeepercent] [retry_for] [exemptfee]
+```
+
+### Parameters
+
+- The `scid` is the short_channel_id of the channel to drain or fill.
+- OPTIONAL: The `percentage` paramter tells the plugin how much of the channel
+  capacity should be drained/filled. Default: 100. Likewise a 'drain 50' will
+  balance a channel that is above 50% of its capacity same as 'fill 50' will
+  balance a channel that is below 50% of its capacity.
+- OPTIONAL: The `chunks` parameter tells the plugin to try breaking down the
+  payment into several smaller ones. In this case it may happen that the
+  operation will only be partially completed. The parameters value is the
+  number of chunks to use. Default: auto-detect based on capacities, max 16.
+- OPTIONAL: `maxfeepercent` is a perecentage limit of the money to be paid in
+  fees and defaults to 0.5.
+- OPTIONAL: `retry_for` defines the number of seconds the plugin will retry to
+  find a suitable route. Default: 60 seconds. Note: Applies for each chunk.
+- OPTIONAL: The `exemptfee` option can be used for tiny payments which would be
+  dominated by the fee leveraged by forwarding nodes. Setting `exemptfee`
+  allows the `maxfeepercent` check to be skipped on fees that are smaller than
+  exemptfee (default: 5000 millisatoshi).
+
+NOTE: The `percentage` relates to resulting target capacity. A 'drain 1%' will
+      not try to send 1% of channel balance away, but try to reduce its balance
+      to 'just' 99%, if channel is above 99% total capacity.
+
+NOTE: Automatically converting a 'drain 70' to a 'fill 30' if the channel is
+      already below 30% capacity is not implemented yet.
+
+
+## Tips and Tricks
+
+- To find the correct channel IDs, you can use the `summary` plugin which can
+  be found [here](https://github.com/lightningd/plugins/tree/master/summary).
+- After some failed attempts, may worth checking the `lightningd` logs for
+  further information.
+- Channels have a `channel_reserve_satoshis` value, which is usually 1% of the
+  channel's total balance. Initially, this reserve may not be met, as only one
+  side has funds; but the protocol ensures that there is always progress toward
+  meeting this reserve, and once met, [it is maintained.](https://github.com/lightningnetwork/lightning-rfc/blob/master/02-peer-protocol.md#rationale)
+  Therefore you cannot drain or fill a channel to be completely empty or full.
--- a/drain/drain.py
+++ b/drain/drain.py
@@ -394,7 +394,8 @@ def drain(plugin, scid: str, percentage: float=100, chunks: int=0, maxfeepercent
         retry_for: int=60, exemptfee: Millisatoshi=Millisatoshi(5000)):
     """Draining channel liquidity with circular payments.
 
-    Percentage defaults to 100. Chunks defaults to 0 (auto-detect).
+    Percentage defaults to 100, resulting in empty channel. Chunks defaults to 0 (auto-detect).
+    Use 'drain 50' to balance a channel that is above 50% capacity.
     """
     return execute('drain', scid, percentage, chunks, maxfeepercent, retry_for, exemptfee)
 
@@ -404,7 +405,8 @@ def fill(plugin, scid: str, percentage: float=100, chunks: int=0, maxfeepercent:
         retry_for: int=60, exemptfee: Millisatoshi=Millisatoshi(5000)):
     """Filling channel liquidity with circular payments.
 
-    Percentage defaults to 100. Chunks defaults to 0 (auto-detect).
+    Percentage defaults to 100, resulting in full channel. Chunks defaults to 0 (auto-detect).
+    Use 'fill 50' to balance a channel that is below 50% capacity.
     """
     return execute('fill', scid, percentage, chunks, maxfeepercent, retry_for, exemptfee)
 
--- a/drain/drain.py
+++ b/drain/drain.py
@@ -17,10 +17,10 @@ HTLC_FEE_MAX = Millisatoshi('100000sat')
 HTLC_FEE_PAT = re.compile("^.* HTLC fee: ([0-9]+sat).*$")
 
 
-def setup_routing_fees(plugin, payload, route, substractfees: bool=False):
+def setup_routing_fees(plugin, payload, route, amount, substractfees: bool=False):
     delay = int(plugin.get_option('cltv-final'))
 
-    amount_iter = payload['amount']
+    amount_iter = amount
     for r in reversed(route):
         r['msatoshi'] = amount_iter.millisatoshis
         r['amount_msat'] = amount_iter
@@ -36,7 +36,7 @@ def setup_routing_fees(plugin, payload, route, substractfees: bool=False):
     # amounts have to be calculated the other way when being fee substracted
     # we took the upper loop as well for the delay parameter
     if substractfees:
-        amount_iter = payload['amount']
+        amount_iter = amount
         first = True
         for r in route:
             channels = plugin.rpc.listchannels(r['channel'])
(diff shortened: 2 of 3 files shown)


m-schmoock:
I removed the "work in progress" TAG, as the plugin is very usable aready and I dont have good clues about the remaining TODOs and optimizations.

I could need a review now... volunteers?

drain/drain.py
```diff
@@ -0,0 +1,418 @@
...
+    test_or_set_chunks(plugin, payload, my_id)
+    plugin.log("%s  %s  %d%%  %d chunks" % (payload['command'], payload['scid'], payload['percentage'], payload['chunks']))
+
+    # iterate of chunks, default just one
+    for chunk in range(payload['chunks']):
+        # we discover capacities each chunk, as fees affect reserves
+        spendable, receivable = spendable_from_scid(plugin, payload, payload['scid'])
+        spendable_bak = spendable
+        total = spendable + receivable
+        #current_percentage = 100 * int(spendable) / int(total)
```
darosior: Trailing comment ? :)

drain/drain.py
```diff
@@ -0,0 +1,418 @@
...
+
+    return cleanup(plugin, payload)
+
+
+@plugin.method("drain")
+def drain(plugin, scid: str, percentage: int=100, chunks: int=0, maxfeepercent: float=0.5,
+        retry_for: int=60, exemptfee: Millisatoshi=Millisatoshi(5000)):
+    """Draining channel liquidity with circular payments.
+
+    Percentage defaults to 100. Chunks deaults to 0 (auto-detect).
```
darosior:
```suggestion
    Percentage defaults to 100. Chunks defaults to 0 (auto-detect).
```

drain/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+The plugin can be started with `lightningd` by adding the `--plugin` option.
+Remember that all `lightningd` plugins have to have executable permissions.
+
+```
+lightningd --plugin=/path/to/plugin/drain.py
+```
+
+Alternatively, you can also symlink or copy the plugins executable to the
+`plugins` folder of your c-lightning installation, as all of the executables
```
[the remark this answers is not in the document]
m-schmoock: ack

drain/drain.py
```diff
@@ -0,0 +1,418 @@
...
+
+
+def spendable_from_scid(plugin, payload, scid):
+    # only fetch funds once to reduce RPC load
+    if not "funds" in payload:
+        payload['funds'] = plugin.rpc.listfunds().get('channels')
+
+    try:
+        channel_funds = next(c for c in payload['funds'] if 'short_channel_id' in c and c['short_channel_id'] == scid)
+    except StopIteration as e:
```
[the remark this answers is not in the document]
m-schmoock: +1 moar nits, keep them coming ;)

drain/drain.py
```diff
@@ -0,0 +1,418 @@
...
+    their_reserve = Millisatoshi(channel_peer['their_reserve_msat'])
+    their = total - our
+
+    # reserves maybe not filled up yet
+    if our < our_reserve:
+        our_reserve = our
+    if their < their_reserve:
+        their_reserve = their
+
+    #spendable = our - our_reserve
```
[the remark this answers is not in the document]
m-schmoock: yup

drain/drain.py
```diff
@@ -0,0 +1,418 @@
...
+    channels = plugin.rpc.listchannels(short_channel_id).get('channels')
+    for ch in channels:
+        if ch['source'] == my_id:
+            return ch['destination']
+    raise RpcError(payload['command'], payload, {'message': 'Cannot find peer for channel: ' + short_channel_id})
+
+
+def find_worst_channel(route):
+    if len(route) < 4:
+        return None
```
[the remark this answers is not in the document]
m-schmoock: hmmm, good point. this one was copy&pasta from @gallizoltan rebalance plugin. I think this was done to prevent removing/excluding too many hops if route is already too short or something.

Also, it does not seem to be related to the remaining drain 'Could not find a route' bug.

m-schmoock:
> I also got an error when trying `drain` :
> 
> ```
> Error while processing drain: TypeError('__int__ returned non-int (type float)')
> ```

Which version of pylightnting are you using. Also can you give me a stacktrace or something on this? Would be very helpful to fix this...

m-schmoock:
@darosior 
> I also got an error when trying `drain` :
> 
> ```
> Error while processing drain: TypeError('__int__ returned non-int (type float)')
> ```

Which version of pylightnting are you using. Also can you give me a stacktrace or something on this? Would be very helpful to fix this...

drain/drain.py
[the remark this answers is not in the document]
m-schmoock: Hm, the `start_id` is 2, meaning it is the third entry in the route...

@gallizoltan can you elaborate on this version of the `find_worst_channel` function maybe (rebalance has the same) ?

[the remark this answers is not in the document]
gallizoltan: Okay, it is partly because of historical reasons, and maybe obsolete.

When I wrote this I assumed we need at least 3 channels/hops for a rebalance:
1) The outgoing channel to the outgoing node
2) A middle channel to the incoming node
3) The incoming channel to my node

Number 1 and 3 are fixed. We could only exclude number 2.
But when I wrote this, there was no `setchannelfee` function. A node applied the same fee for all its channels. So there was no point to exclude channel number 2, because its fee set by the outgoing node, so we could not find a cheaper channel instead of that.
This is why `start_id` is 2: possible excludes starts with the third entry in the route.

darosior:
@m-schmoock sorry for the delayed answer.

> Which version of pylightnting are you using.

Last released one (0.7 I think), should I try with master's one ?

> Also can you give me a stacktrace or something on this?

Actually I don't have any. I tried to quickly go back through the execution but did not found from where it comes (didn't search that much though).

darosior:
Hi @m-schmoock, at which stage are you with this ? Is still work in progress because of the TODOs, or are there still bugs ? If this is only about the TODOs maybe we can merge it and other contributors could help improve this (definitely wanted :-)) plugin ?

m-schmoock:
Hi @darosior , kinda glad you asked! 

The plug-in works almost like a charm. The reason it is not merged yet is, that it raises a reproducible 'route not found' error after doing several 'drain' on varoius channels in a row.  I'm not 100% sure if this is caused by the plug-in or lightningd. If you then stop, remove gossip store, restart, wait and retry, the issue will reappear only after doing some drains again. If you have time and ambitions to debug it feel free. Also, this does not seem to happen on 'fill' which is why I am more unsure what causes it.

F.y.i. I will have more time on from November, since I reduced my regular job to just 25hrs a week :). This was going to be my first task anyway.

Technically and code wise, it is in a mergable state beside that.

darosior:
> The plug-in works almost like a charm. The reason it is not merged yet is, that it raises a reproducible 'route not found' error after doing several 'drain' on varoius channels in a row. I'm not 100% sure if this is caused by the plug-in or lightningd. If you then stop, remove gossip store, restart, wait and retry, the issue will reappear only after doing some drains again. If you have time and ambitions to debug it feel free. Also, this does not seem to happen on 'fill' which is why I am more unsure what causes it.

Hmmm interesting, we might leave this as an open issue ? Don't know what the repo captain thinks about that @cdecker :-) ?

> F.y.i. I will have more time on from November, since I reduced my regular job to just 25hrs a week :). This was going to be my first task anyway.

Great ! To work on Lightning ?

darosior:
Btw if you think it's mergeable you should remove the WIP tag

m-schmoock:
@darosior 

I removed the WIP flag, still Im going to fix this one way or another, now that we talked about this issues it is at least known. If you are fine we can merge, I lost my commit access because I have been inactive for a while, so we need someone else to merge it.

> > F.y.i. I will have more time on from November, since I reduced my regular job to just 25hrs a week :). This was going to be my first task anyway.

> Great ! To work on Lightning ?

Yes !

darosior:
> If you are fine we can merge, I lost my commit access because I have been inactive for a while, so we need someone else to merge it.

I don't have one so we'll have to wait for ackbot ;-)

drain/requirements.txt
```diff
@@ -0,0 +1 @@
+pylightning==0.0.7
```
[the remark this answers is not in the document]
m-schmoock: thx. will go for ` >= 0.0.7.3 `

drain/drain.py
```diff
@@ -0,0 +1,417 @@
...
+        if payload['command'] == 'drain' and recv < payload['amount']:
+            excludes += [channel['short_channel_id']+'/0', channel['short_channel_id']+'/1']
+        if payload['command'] == 'fill' and spend < payload['amount']:
+            excludes += [channel['short_channel_id']+'/0', channel['short_channel_id']+'/1']
+
+    while int(time.time()) - start_ts < payload['retry_for']:
+        if payload['command'] == 'drain':
+            r = plugin.rpc.getroute(my_id, payload['amount'], riskfactor=0,
+                    cltv=9, fromid=peer_id, fuzzpercent=0, exclude=excludes)
+            route_out = {'id': peer_id, 'channel': payload['scid'], 'direction': int(not my_id < peer_id)}
```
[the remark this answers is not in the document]
m-schmoock: can change, it was "more readable" to me, dunno why :D

drain/drain.py
```diff
@@ -0,0 +1,417 @@
...
+
+def test_or_set_chunks(plugin, payload, my_id):
+    scid = payload['scid']
+    cmd = payload['command']
+    spendable, receivable = spendable_from_scid(plugin, payload, scid)
+    total = spendable + receivable
+    target = total * (0.01 * (100 - payload['percentage']))
+
+    if cmd == "drain":
+        if target >= spendable:
```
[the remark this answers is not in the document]
m-schmoock: Hm, because if they are equal the next line `amount = spendable - target` will set 0 amount which will raise errors later on..

drain/drain.py
```diff
@@ -0,0 +1,417 @@
...
+        "exemptfee": exemptfee,
+
+        "labels" : [],
+        "success_msg" : [],
+    }
+
+    # check parameters
+    if command != 'drain' and command != 'fill':
+        raise RpcError(payload['command'], payload, {'message': 'Invalid command. Must be drain or fill.'})
+    if percentage < 0 or percentage > 100:
```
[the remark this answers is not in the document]
m-schmoock: We can change back, once the plugin converts a "fill 0%" to a "drain 100%". Currently it will trow an error as the target is above/below spendable/receivable.

drain/drain.py
```diff
@@ -0,0 +1,417 @@
...
+
+def test_or_set_chunks(plugin, payload, my_id):
+    scid = payload['scid']
+    cmd = payload['command']
+    spendable, receivable = spendable_from_scid(plugin, payload, scid)
+    total = spendable + receivable
+    target = total * (0.01 * (100 - payload['percentage']))
+
+    if cmd == "drain":
+        if target >= spendable:
```
darosior: FWIW after testing if I drain a channel at exactly 50% then these conditions are mutually exclusive ;-)

Ex:
```bash
$ lnreg1 drain 366x1x0 50 50
[
   "987762msat sent over 3 hops to drain 987740msat [1/50]",
   "997392msat sent over 3 hops to drain 997370msat [2/50]",
   "1007783msat sent over 3 hops to drain 1007759msat [3/50]",
   ....

$ lnreg1 drain 366x1x0 1 20
"Error while processing drain: RpcError(\"RPC call failed: ..... error: {'message': 'channel already below target value'}\")"
$ lnreg1 fill 366x1x0 1 20
"Error while processing fill: RpcError(\"RPC call failed: .......  error: {'message': 'channel already above target value'}\")"
```
```

Comment (darosior) [reply drain/drain.py]: What's fun is that rebalanced all the regtest btw ^^

Comment (darosior) [reply drain/drain.py]: Hmm think I'll have to test on the only viable test network : mainnet

File: drain/drain.py [hunk 12]
```diff
@@ -0,0 +1,417 @@
...
+
+            if erring_channel == payload['scid']:
+                raise RpcError(payload['command'], payload, {'message': 'Error with selected channel: %s' % erring_message})
+
+            plugin.log("RpcError: " + str(e))
+            if erring_channel is not None and erring_direction is not None:
+                excludes.append(erring_channel + '/' + str(erring_direction))
+
+
+def execute(command: str, scid: str, percentage: int,
```
darosior: Trying this for real, I think it could be great if it could be a float as you may often need to not drain / fill a significant percentage of the channel funds

drain/drain.py
```diff
@@ -0,0 +1,417 @@
...
+    scid = payload['scid']
+    cmd = payload['command']
+    spendable, receivable = spendable_from_scid(plugin, payload, scid)
+    total = spendable + receivable
+    target = total * (0.01 * (100 - payload['percentage']))
+
+    if cmd == "drain":
+        if target >= spendable:
+            raise RpcError(payload['command'], payload, {'message': 'channel already below target value'})
+        amount = spendable - target
```
darosior: Trying this some more I think I don't get the logic (or don't understand the params) :
I tried to drain some sats from a channel with `293600000`msat (`287728000` spendable), setting the percentage to 30 I effectively drained `33807374`msat, which is about 11%of the channel capacity:
```python
>>> 33807374 / 287728000 * 100
11.749768531390758
>>> 33807374 / 293600000 * 100
11.514773160762942
```
Hmm trying this again I then got an error that the target was above the capacity (or below to fill), even with 1% as parameter ! Is this the error you were talking about ?

Back to the highlighted lines I think you compute the target as the final channel capacity:
```python
target = total * (0.01 * (100 - payload['percentage']))
```
Then you compare it to the spendable amount
```python
if target >= spendable:
```
Which will likely be inferior.
I think that the target should be "how much do I want to drain" and thus the target be :
```python
target = total * payload['percentage'] / 100
```
Or 
```python
    # The amount to either fill or drain
    amount = total * payload['percentage'] / 100

    if cmd == "drain":
        if amount >= spendable:
             # Hmm problem here
    elif cmd == "fill":
        if amount >= receivable:
             # Hmmmm
```

But maybe I'm just misunderstanding and misusing the whole thing

m-schmoock: Yes, the percentage amount is the resulting final channel capacity. The error can be reproduced when quickly draining and filling several channels (i.e. 100% default) until the error is raised.

m-schmoock: Note: I edited my last response to be more precise.

darosior: After more testing I think that's the latter case.. :
> But maybe I'm just misunderstanding and misusing the whole thing

darosior: But I still don't understand why I could not drain (or fill) 1% of a channel that is completely balanced (nor the target logic)..

m-schmoock: That's easy, a balanced channel is by definition 50% filled and drained 'at the same time'. You cannot be able to 'drain' to 1% as that would equate to a fill 99% (but command inversion is a Todo and not implemented yet.). Equally a fill to 1% is impossible as it has already 50% which is greater than 1%.

The percentage do not related to current balance but to target amount.

m-schmoock: Do you think we should implement command handling differently, if yes how exactly?

Current pros:
- do not need to know current percentage l
- no amount calculation needed
- Able to rebalance by using drain or fill 50%

Cons:
- counter intuitive? :D
- 'not often possible' to do stuff like drain 1% as that would require to have the channel above 99%.

m-schmoock:
I updated the documentation a bit to make clear how it is currently supposed to work.

drain/drain.py
darosior: Thanks for the explanation !

> Do you think we should implement command handling differently, if yes how exactly?

It was counter intuitive *to me*, maybe because I haven't read "mastering english" yet :smile: 
However I'd argue to maybe change the command name ? According to my understanding, `drain <amount> <scid>` implies `I want to drain <amount>% of sats from <scid>` : actually I tried to use it as a rebalance. Maybe `set_ours <percent>` `set_them <percent>`, they can still be additional commands though.

>  I think from a UX requirements perspective it is important to be able to get a channel balance full or empty (hence 100% default) or some to fractional value without doing math calculations. It is not a useful feature to send 1% or X% of a channel elsewhere.

I totally agree. Just asking as many questions as possible to make this as perfect as possible ;-) Next step (once merged) is the htlc fee finding optimisation

m-schmoock: @darosior can you come to irc? Or can we have a quick talk somewhere?

darosior: Yep I pinged you on IRC

darosior: > How about this:

    drain X will send X% total (or current?) channel balance away to some other channel, if possible.
    fill X will receive X% total (or current?) channel balance from some other channel, if possible.
    reach X will try to achieve X% our capacity, if possible. This command will actually do the auto conversion (drain/fill) if channel is above or below.

Sounds great

darosior: Ok so I have a channel with 60% theirs, 40% ours. I think the wording is misleading as if I want to drain, say 30 percent of the channel (so that the balance results in 90% theirs, 10% ours), I dont
```
cli drain <scid> 30
```
But
```
cli drain <scid> 90
```
So effectively passing to the drain command the expected percent on `their` side. That's why I proposed to change the wording to :
```
cli set_theirs <scid> 90
```
And, if we keep the `drain` command, to pass it the expected percent to __drain from__ my side.

The same goes for `fill` :
Now that they effectively have 90% of the channel capacity, and I want my side to have 90% capacity (so passing from `ours: 10%; theirs: 90%` to `ours: 90%; theirs: 10%`) I would expect to drain __from them__ 80% of the capacity (hence `fill` 80% on my side) but in order to do that with the current wording I have to :
```
cli fill <scid> 90
```
That's why I propose to rename it to 
```
cli set_ours <scid> 80
```

drain/drain.py
```diff
@@ -0,0 +1,417 @@
...
+                fee += (amount_iter * ch['fee_per_millionth'] + 10**6 - 1) // 10**6    # integer math trick to round up
+                if fee > amount_iter:
+                    raise RpcError(payload['command'], payload, {'message': 'cannot cover fees to %s %s' % (payload['command'], payload['amount'])})
+                amount_iter -= fee
+            first = False
+            r['msatoshi'] = amount_iter.millisatoshis
+            r['amount_msat'] = amount_iter
+
+
+def get_channel(plugin, payload, peer_id, scid, check_state: bool=False):
```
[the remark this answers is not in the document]
m-schmoock: thx. this was historic code. removed the check_state parameter all together and to the check or raise error.

drain/drain.py
```diff
@@ -0,0 +1,417 @@
...
+        "success_msg" : [],
+    }
+
+    # check parameters
+    if command != 'drain' and command != 'fill':
+        raise RpcError(payload['command'], payload, {'message': 'Invalid command. Must be drain or fill.'})
+    if percentage < 0 or percentage > 100:
+        raise RpcError(payload['command'], payload, {'message': 'Percentage must be between 0 and 100'})
+    if chunks < 0:
+        raise RpcError(payload['command'], payload, {'message': 'Negative chunks do not make sense. Try a positive value or use 0 (default) for auto-detection.'})
```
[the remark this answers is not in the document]
m-schmoock: Chunk size is problematic from a UX, as it would require to do some math before invoking the operation. The chunk parameter is auto detected per default and should only be touched when automatic fails, as using more chunks than required increases fees...

drain/drain.py
m-schmoock: thx, addressed by 26e8717

drain/drain.py
```diff
@@ -0,0 +1,417 @@
...
+    spendable = channel_peer['spendable_msat']
+    receivable = their - their_reserve
+    return spendable, receivable
+
+
+def peer_from_scid(plugin, payload, short_channel_id, my_id):
+    channels = plugin.rpc.listchannels(short_channel_id).get('channels')
+    for ch in channels:
+        if ch['source'] == my_id:
+            return ch['destination']
```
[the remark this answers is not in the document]
m-schmoock: address by d9c462a

drain/drain.py
m-schmoock: will be address by implementing this TODO 8d555a46dbd6dba32485ce4ea5a78955bae224ff

drain/drain.py
```diff
@@ -0,0 +1,417 @@
...
+    return execute('fill', scid, percentage, chunks, maxfeepercent, retry_for, exemptfee)
+
+
+@plugin.init()
+def init(options, configuration, plugin):
+    plugin.options['cltv-final']['value'] = plugin.rpc.listconfigs().get('cltv-final')
+    plugin.log("Plugin drain.py initialized")
+
+
+plugin.add_option('cltv-final', 10, 'Number of blocks for final CheckLockTimeVerify expiry')
```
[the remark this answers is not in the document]
m-schmoock: Yes,

this one is copy&paste from rebalance and sendinvoiceless plugins.
It declares the `cltv-final` option the plugin is going to be using and also reads out lightningd's  `cltv-final` value in the `init` method. Maybe this is duplicated and we should always use `plugin.rpc.listconfigs().get('cltv-final')` but this increases RPC load.

So unless there is no other reason than duplication, I think we can keep it.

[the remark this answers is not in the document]
m-schmoock: @darosior okay?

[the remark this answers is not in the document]
darosior: > @darosior okay?

Yeah, tbh I don't use plugin options that much. But it's interesting that we actually can override `lightningd` options.

drain/drain.py
```diff
@@ -0,0 +1,417 @@
...
+import time
+import uuid
+
+# TODOs:
+#  - fix: try increasing chunks on route errors
+#  - fix: sometimes, if we ran in error, not all chunk results are returned, i.e. [2/4, error] but not 1/4.
+#  - fix: occasionally strange route errors
+#  - feat: set HTLC_FEE MIN/MAX/STP by feerate
+#  - feat: convert a 'drain 0%' to a 'fill 100%' and vice versa instead of throwing error
+#  - chore: reconsider use of listchannels
```
cdecker: Might be worth putting these in a markdown README in the subdirectory. Makes it clearer that there are some TODOs and allows you to tick them off visibly :-)

m-schmoock: I will get this done in the next two weeks, I am about to have more time for all this... promise :D

m-schmoock: @cdecker  I want to change the meaning of `drain` and `fill` to not mean 'total target percentge' but rather just 'total percentage', as it caused confusions for @darosior for being unintuitive. Ignore over draining/filling (replacing it by whats possible). And add a `setbalance` command that does all the auto stuff.

Regarding the routing errors, I will have to check. Maybe we want this merged first.

cdecker:
Sorry for the delay, I seem to have missed the non-WIP window :wink: 

I'm quite happy to merge the plugin even if it has a edge-case that may fail some times. fwiw I have a c-lightning PR pending that should allow us to start testing plugins, so pinning down the issue and reproducing it will become easier: https://github.com/ElementsProject/lightning/pull/3218

drain/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+The plugin has also a `fill` command that does excactly the opposite. You
+can use it to fill up a given percentage of liquidity (default 100%) on your
+side of a channel:
+
+```
+lightning-cli fill scid [percentage] [chunks] [maxfeepercent] [retry_for] [exemptfee]
+```
+
+Another useful command is the `setbalance` that will fill up or drain your side
```
darosior: :tada: ! :grin:

drain/drain.py
m-schmoock: resolved by c8ac5cf `drain: document setbalance and todo`

m-schmoock:
Hi,

After my first two days of my new 'Altersteilzeit' I managed to get this ready.

@darosior  I added the `setbalance` command and rewrote the `drain` `fill` to behave more intuitive. thanks for the feedback. The new command works like a charm.
@cdecker  I finally removed the WIP tag. Can we make final reviews and merge?

cheers,
Michael

m-schmoock:
You guys also might want to test this, but for my setup and testnet it works smoothly

darosior:
> testnet and mainnet it works smoothly.

So you don't have the spontaneous errors you were talking about ?

m-schmoock:
From time to time, which is why I think it's not plug-in but daemon related.

PR #23, opened 2019-05-10
fix: a race condition in the summary plugin

m-schmoock:
When the summary plugin starts, it runs a thread to discover the current
`fiat_per_btc` price. When this was not finished yet, the `if` statement
in line 112 fails, because this in not JavaScript and the property has
to be checked with `hasattr(obj, key)` method in order to not throw an
exception.

### DIFF
--- a/summary/summary.py
+++ b/summary/summary.py
@@ -109,7 +109,7 @@ def summary(plugin):
     reply['avail_out'] = avail_out.to_btc_str()
     reply['avail_in'] = avail_in.to_btc_str()
 
-    if plugin.fiat_per_btc:
+    if plugin.fiat_per_btc > 0:
         reply['utxo_amount'] += ' ({})'.format(to_fiatstr(utxo_amount))
         reply['avail_out'] += ' ({})'.format(to_fiatstr(avail_out))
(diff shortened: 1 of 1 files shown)


summary/summary.py
```diff
@@ -109,7 +109,7 @@ def summary(plugin):
     reply['avail_out'] = avail_out.to_btc_str()
     reply['avail_in'] = avail_in.to_btc_str()
 
-    if plugin.fiat_per_btc:
+    if hasattr(plugin, 'fiat_per_btc'):
```
cdecker: We really should initialize this in `__init__` since this kind of static method call on an object is generally frowned upon.

m-schmoock: @cdecker like this 858de14 ?

m-schmoock:
@cdecker I used the plugins `init` and not the modules `__init__.py` because I'm not sure if `__init__.py` is supported when linking the executable `.py` simply to cln `./plugins` directory to skip `--plugin=`

cdecker:
ACK 858de14dd49eecc228a6c73a902d62464f9552f4

PR #28, opened 2019-05-15
Add lightning-qt as a submodule and to the README list

darosior:


m-schmoock:
I like it :+1: 

@darosior  p.s.: If you use RPC socket forwarding through ssh, I think your are currently not able to do stuff like your plugin registering for `lightningd´ events.  So you (currently) have to poll or use X forwarding (which sucks).

darosior:
> I like it 
> 
> @darosior p.s.: If you use RPC socket forwarding through ssh, I think your are currently not able to do stuff like your plugin registering for `lightningd` events. So you (currently) have to poll or use X forwarding (which sucks).

Thanks :D 
Yeah I finally noticed it, but Qt X forwarding is quite an experiment x)

Issue #29, opened 2019-05-20
Cannot Rebalance without msatoshi argument


agnjkafgh:
I would like to rebalance my channels 50/50 which should work if I do not provide the `msatoshi` argument. Looks like msatoshi should be a whole number and not have a decimal point.
```
"Error while processing rebalance: 
RpcError('RPC call failed: method: invoice, payload: 
{
\\'msatoshi\\' : 8388607500.0msat, 
\\'label\\': \\'Rebalance-23a20dc6-356f-423e-8a0c-6a835a061847\\', 
\\'description\\ ': \\'1516934x52x0 to 1517537x39x0\\', 
\\'expiry\\': 120
}, 
error: {
\\'code\\': -32602, 
\\'message\\': \ "\\'msatoshi\\' should be millisatoshis or \\'any\\', not \\'8388607500.0msat\\'\"
}',)"

```

m-schmoock:
@agnjkafgh did the error occur when not supplying a msat amount and let the plugin do the math? Or did you try to balance `8388607500.0msat`, because there is obviously no sub-millisatoshi...

agnjkafgh:
Yup. I submitted two channels which have 16777215 sats each and the plugin did the math correctly. I believe the plugin attempted to create an invoice with the **8388607500.0msat** in the `msatoshi` field and it failed. If I am not mistaken if the plugin created an invoice with  8388607500msat instead it should have worked.

m-schmoock:
Thx. Can you also supply logs ? I never had this...

agnjkafgh:
`cli/lightning-cli rebalance 1516934x52x0 1517537x39x0`
```
"Error while processing rebalance: 
RpcError('RPC call failed: method: invoice, payload: 
{
\\'msatoshi\\' : 8388607500.0msat, 
\\'label\\': \\'Rebalance-23a20dc6-356f-423e-8a0c-6a835a061847\\', 
\\'description\\ ': \\'1516934x52x0 to 1517537x39x0\\', 
\\'expiry\\': 120
}, 
error: {
\\'code\\': -32602, 
\\'message\\': \ "\\'msatoshi\\' should be millisatoshis or \\'any\\', not \\'8388607500.0msat\\'\"
}',)"
```

m-schmoock:
I meant the logfile... Not cmd output

m-schmoock:
Nvm, have it. Thx. will fix quickly...

m-schmoock:
[...]

PR #30, opened 2019-05-20
rebalance: fix issue 29

m-schmoock:
This should solve #29 where the auto amount calculation created decimal `123.0msat` amounts.
I think this issue happens on older version of pylightning only.

### DIFF
--- a/rebalance/rebalance.py
+++ b/rebalance/rebalance.py
@@ -85,8 +85,8 @@ def calc_optimal_amount(out_ours, out_total, in_ours, in_total, payload):
     in_ours, in_total = int(in_ours), int(in_total)
 
     in_theirs = in_total - in_ours
-    vo = out_ours - (out_total/2)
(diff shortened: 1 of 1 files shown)


rebalance/rebalance.py
```diff
@@ -85,8 +85,8 @@ def calc_optimal_amount(out_ours, out_total, in_ours, in_total, payload):
     in_ours, in_total = int(in_ours), int(in_total)
 
     in_theirs = in_total - in_ours
-    vo = out_ours - (out_total/2)
-    vi = (in_total/2) - in_ours
+    vo = int(out_ours - (out_total/2))
+    vi = int((in_total/2) - in_ours)
```
cdecker: You can use `//` to have the division round to `int` in future :-)

m-schmoock: gracias

PR #32, opened 2019-05-21
doc: auto plugin directories. main repo issue 2647

m-schmoock:
This documents the recently merged https://github.com/ElementsProject/lightning/pull/2647
that points the users to organize their plugins not by a growing cascade of `--plugin` arguments, but to put them in `~/.lightning/plugins/...`


cdecker:
ACK https://github.com/lightningd/plugins/pull/32/commits/e82bfb2ab2f8ef850058184719fecf720df470ae

PR #33, opened 2019-05-23
Cleanup rebalance sendinvoiceless

m-schmoock:
This is a **work in progress** PR on refinements I discovered when implementing the `drain` plugin (#22)

Affects `rebalance`, `sendinvoiceless` and `summary`.

### DIFF
--- a/rebalance/rebalance.py
+++ b/rebalance/rebalance.py
@@ -9,15 +9,23 @@ plugin = Plugin()
 def setup_routing_fees(plugin, route, msatoshi):
     delay = int(plugin.get_option('cltv-final'))
     for r in reversed(route):
-        r['msatoshi'] = r['amount_msat'] = msatoshi
+        r['msatoshi'] = msatoshi.millisatoshis
+        r['amount_msat'] = msatoshi
         r['delay'] = delay
         channels = plugin.rpc.listchannels(r['channel'])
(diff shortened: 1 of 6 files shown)


m-schmoock:
> cdecker added work in progress  and removed work in progress  labels 8 hours ago

Whats that supposed to mean? ( •_•)

rebalance/rebalance.py
```diff
@@ -109,11 +117,10 @@ def calc_optimal_amount(out_ours, out_total, in_ours, in_total, payload):
 
 @plugin.method("rebalance")
 def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi=None,
-              maxfeepercent="0.5", retry_for="60", exemptfee: Millisatoshi=Millisatoshi(5000)):
+              maxfeepercent: float=0.5, retry_for: int=60, exemptfee: Millisatoshi=Millisatoshi(5000)):
```
cdecker: Notice that these type annotations will not really result in any type-coercion, so the casts below should probably stay in.

m-schmoock: Thx for the hint. I added step that force casts the arithmetic arguments, so we do not need to worry about casting later on.
I allow myself the liberty to merge the commit, as its just that and you ACKed.

PR #34, opened 2019-05-28
Summary: autopilot: add requirements.txt

jsarenik:


jsarenik:
Tested with virtual environment:

```
pip3 install --user virtualenv

cd autopilot
PYTHON3=$(which python3)
VENV=$PWD/venv

virtualenv -p $PYTHON3 $VENV
. $VENV/bin/activate
# manually install pylightning
# pip install ../../lightning/contrib/pylightning
pip install -r requirements.txt

lightningd --plugin=./autopilot.py
```

jsarenik:
> `pylightning>=0.0.7`
> 
> Should be sufficient. Not sure about networkx and numpy if we need to nail it down to a specific version.

Good point. I don't think any of them need to be nailed down. Force-pushing in a minute...

m-schmoock:
@jsarenik thx :D

PR #36, opened 2019-06-03
rebalance: fix amount None

m-schmoock:
This was really stupid, before merging the last one, I added hard parameter casting at the top of the function without testing the amount `msatoshi` parameter for default (None) that was then used to call `Millisatoshi(None)`.


cdecker:
ACK https://github.com/lightningd/plugins/pull/36/commits/1495c501fc745d29c6056fd219740661b2003b5b

PR #37, opened 2019-06-10
add remoteRPC plugin

rsbondi:
This is an http/https passthrough to/from the unix socket allowing remote rpc calls


cdecker:
ACK https://github.com/lightningd/plugins/pull/37/commits/4add2e20de15f0285a11af3553a3ffe2e41a259c

PR #38, opened 2019-06-10
rebalance: fix route direction

m-schmoock:
This will set the correct `direction` on manual route hops first and last (in/out).
Before this was not set, resulting in WIRE_FEE_INSUFFICIENT errors,
if fee was calculated (correctly) but inconsistently with `direction`, defaulting perhaps to `0`.

The error occurred on `sendpay` doing checks for the route I guess. I noticed this when having WIRE_FEE_INSUFFICIENT errors on last hop, where my peer had a higher fee setting than me and direction should have been 1.


cdecker:
ACK https://github.com/lightningd/plugins/pull/38/commits/deaf7e94c8ce456859556d53f98de27d7e47fa08

PR #40, opened 2019-07-13
dbbackup:  A minimal database backup plugin

SimonVrouwe:
Here a simple plugin that keeps a synchronized backup of CL's database.

I heard rumors that similar (more advanced ) backup plugins are being developed. But I needed this now to sleep better, please comment if that is justified or not.

Some tests are included that can be run in c-lightnings `pytest` framework by linking this plugin directory into CL's `/tests`. It is not ideal, but couldn't find another solution.

Attempts to fix [#2400](https://github.com/ElementsProject/lightning/issues/2400)

**Edit**: Last `test_dbbackup_plugin_kill` requires a logline in CL that's not there yet (will create PR)


cdecker:
Awesome work, I'll review this asap ^^

fiatjaf:
Just tested in production and it works. Now yeah, someone must review it.

cdecker:
> Some tests are included that can be run in c-lightnings pytest framework by linking this plugin directory into CL's /tests. It is not ideal, but couldn't find another solution.

That's something that I have been wanting to look into. Basically the `pylightning` library should contain the testing framework that we use for c-lightning as well, so that you can just add that as a dependency in your `requirements.txt` file and, providing you have c-lightning installed somewhere where we can find it, it should be possible to just run against it.

I'm sorry you had to jump through these hoops to get testing working, but great work getting testing working :+1:

SimonVrouwe:
Modified README.md (nothing else) and squashed everything into two commits.

The '/tests` inside plugin directory appears harmless, because it doesn't contain any executable files.
Not a python expert, but I assume that `__pycache__` is also harmless.

As a side note, I run this plugin with Toshiba 16GB usb flash drive for the backup-file. Little concerned about flash (p/e cycles) lifetimes, `iostat -d sdb` shows an average ~150MB write per day over past 2 weeks for my not-very-busy node. Using a conservative estimate lifetime of 250 writes per cell, this should last for about 16GB * 250 / (0.150 * 365) = 73 year?

dbbackup/README.md
```diff
@@ -0,0 +1,41 @@
+## Summary plugin
+
+This plugin keeps a synchronized backup of c-lightning's (CL) sqlite3 database.
+It uses the db_write hook so that every commit (write) to CL's database is first
+written to the backup. This allows recovery of any committed-to channel state,
+including HTLCs. This plugin does not backup the seed and is not a complete
+node-backup.
```
cdecker: Excellent clarification :+1:

dbbackup/README.md
```diff
@@ -0,0 +1,41 @@
...
+For general plugin installation instructions see the repos main
+[README.md](https://github.com/lightningd/plugins/blob/master/README.md#Installation)
+
+## Options:
+
+* `--db-backup-file`: path of the backup file
+
+## Usage
+
+If the given [`db-backup-file`] doesn't exist yet, it will be created from a
```
cdecker: Parentheses do not seem to be correct markup, was this intended to be a link to the option above?

dbbackup/dbbackup.py
```diff
@@ -0,0 +1,96 @@
...
+        if os.path.isfile(backup):
+            plugin.log('Found existing db-backup-file: {} comparing...'.format(backup))
+            backup_copy = shutil.copy(backup, backup + '.tmp')
+            db1 = sqlite3.connect(backup_copy, isolation_level=None)
+            db2 = sqlite3.connect('file:{}?mode=ro'.format(db), uri=True)   # open in read-only
+            for c in plugin.sqlite_pre_init_cmds:
+                db1.execute(c)
+
+            # If it then matches orignal db, replace backup with copy ... else abort
+            dbs_match = [x for x in db1.iterdump()] == [x for x in db2.iterdump()]
```
cdecker: Interesting, I was not aware of `iterdump()`. You materialize the entire dump in memory twice here. Would it be possible to not do that, and instead use something like this:

```python
for a, b in zip(db1.iterdump(), db2.iterdump()):
    if a != b:
         return False
```

This creates a new iterator from the two iterators and spits out one comparable tuple per entry, hence avoiding having both dumps in memory and saving multiple iterations.

dbbackup/tests/test_dbbackup.py
```diff
@@ -0,0 +1,161 @@
...
+def test_dbbackup_write_fail(node_factory, executor):
+    """When the plugin cannot write to backup file for some reason"""
+
+    backup = os.path.join(node_factory.directory, "lightningd.sqlite3-backup")
+    l1 = node_factory.get_node(allow_broken_log=True,
+                               options={'plugin': 'tests/dbbackup/dbbackup.py',
+                                        'db-backup-file': backup})
+
+    # rename backup file will cause db_write failure and crash lightningd
+    os.rename(backup, backup + '_')
```
cdecker: How does this cause a failure writing to the DB? I thought sqlite3 keeps an fd open to the db file, which is unaffected by moves.

fiatjaf:
How can we use this to backup the db to another computer?

fiatjaf:
Since I've started to run this plugin I've been seeing hundreds of crashes:

```
+213.639880031 plugin-dbbackup.pyBROKEN: Failed to write to backup: near \"channels\": syntax error
+213.641676833 lightningd(13527):BROKEN: Plugin returned failed db_write: ??{"jsonrpc": "2.0", "id": 231, "result": false}??U.
```

cdecker:
> How can we use this to backup the db to another computer?

You'd likely have to mount a remote disk over the network and write the backup file to that disk.

cdecker:
> Since I've started to run this plugin I've been seeing hundreds of crashes per day:
> 
> ```
> +213.639880031 plugin-dbbackup.pyBROKEN: Failed to write to backup: near \"channels\": syntax error
> +213.641676833 lightningd(13527):BROKEN: Plugin returned failed db_write: ??{"jsonrpc": "2.0", "id": 231, "result": false}??U.
> ```

Can you add a print of the executed SQL command before it is being applied to the replica? Might be a difference in the sqlite3 version used by `lightningd` and the plugin itself.

cdecker:
> Since I've started to run this plugin I've been seeing hundreds of crashes per day:
> 
> ```
> +213.639880031 plugin-dbbackup.pyBROKEN: Failed to write to backup: near \"channels\": syntax error
> +213.641676833 lightningd(13527):BROKEN: Plugin returned failed db_write: ??{"jsonrpc": "2.0", "id": 231, "result": false}??U.
> ```

@fiatjaf  Can you add a print of the executed SQL command before it is being applied to the replica? Might be a difference in the sqlite3 version used by `lightningd` and the plugin itself.

dbbackup/dbbackup.py
SimonVrouwe: Good idea!

> I was not aware of iterdump()

I found it in c-lightning's `test_db_hook`

dbbackup/tests/test_dbbackup.py
SimonVrouwe: I have no idea, but it works . The error returned by sqlite3: `attempt to write a readonly database`

Any other ideas how to emulate a disk-full scenario? (maybe loop device)

dbbackup/README.md
SimonVrouwe: fixed

dbbackup/dbbackup.py
```diff
@@ -0,0 +1,96 @@
...
+        else:
+            plugin.log('Initialization failed: {}'.format(e), 'error')
+
+        plugin.rpc.stop()   # stop lightningd
+
+
+@plugin.hook('db_write')
+def db_write(plugin, writes):
+    if not plugin.initted:
+        plugin.sqlite_pre_init_cmds += writes
```
ZmnSCPxj: There is a potential (but unlikely) race condition where `db_write` is invoked before `init`, then C-lightning crashes before it is able to `init` the plugins.  It may be safer to save these in a temporary file that you clear out once `init` is invoked.

dbbackup/dbbackup.py
```diff
@@ -0,0 +1,96 @@
...
+
+            # If it then matches orignal db, replace backup with copy ... else abort
+            dbs_match = [x for x in db1.iterdump()] == [x for x in db2.iterdump()]
+            db1.close()
+            db2.close()
+            if dbs_match:
+                os.rename(backup_copy, backup)
+                plugin.log("Existing db-backup-file OK and successfully synced")
+            else:
+                plugin.log("Existing db-backup-file differs from original database, i.e. applying"
```
ZmnSCPxj: A better handling of this would be to just outright copy the original over the backup.  Though it would not be safe to do so in the `init` as that is async.  Instead, on the *first* `db_write` *after* `init`, that is when you do your overwriting the backup with the original.  Though this is sensitive to the order in which C-Lightning writes to its own original copy (whether it does the write before `db_write` or after `db_write`).  Might be more complicated than what you are willing to implement and I suppose the operator can manually do the copy in this edge case while C-Lightning is not running.

dbbackup/dbbackup.py
cdecker: Good catch, though writes before the `init` call will all be migrations iirc, so these are not catastrophic.

ZmnSCPxj: Not keeping up with C-lightning upgrades is "not catastrophic"?  Upgrades getting lost in the stored database are arguably worse.  Further, upgrades are a risk by themselves as they represent changed code, and the user might be running on some obscure combination of system components that happen to work on older releases of C-lightning but not on newer ones, so a new release (which is when we are likely to have migrations running) is at increased risk of crashes before `init`... *shrug*

SimonVrouwe: @ZmnSCPxj Thanks, there is also a potential dead-lock when the plugin makes the `rpc.stop()` call from `init` while at the same time `lightningd` is waiting for a `db_write` hook to return. TBH I was totally unaware that plugin with (sync) hooks shouldn't make **any** (sync) RPC calls, at least not from the same threat that also handles the hook.

If I understand correctly, RPC calls can be send after [jsonrpc_listen](https://github.com/ElementsProject/lightning/blob/804aee766e968d27fa117a3b21dca6269e78aa7d/lightningd/lightningd.c#L777) but are not be answered until the main [io_loop_with_timers](https://github.com/ElementsProject/lightning/blob/804aee766e968d27fa117a3b21dca6269e78aa7d/lightningd/lightningd.c#L834)

To solve these and similar issues, I am exploring a _sync init_ in [plugins_config](https://github.com/ElementsProject/lightning/blob/804aee766e968d27fa117a3b21dca6269e78aa7d/lightningd/plugin.c#L1066) where `lightningd` waits for the `init` response of certain (perhaps all static?) plugins before continue startup, in a similar way as it does for `getmanifest`.  And maybe shutdown CL when a plugin's `init` returns False? Will make a PR if that doesn't break too much.

cdecker:
Just a heads up that with PR https://github.com/ElementsProject/lightning/pull/2924 a couple of things will change:

 - The DB initialization will no longer be sent as a `db_write` call. This is limited to the `PRAGMA foreign_keys = ON;` query that enables foreign keys (and by consequence enforcement of cascade constraints on which we rely heavily). So make sure to enable that upon opening the destination DB, otherwise some queries could hit constraints like this https://github.com/ElementsProject/lightning/issues/3007. 
 - We will also no longer send the `BEGIN TRANSACTION;`  and `COMMIT;` statements that initialize and commit a DB transaction. This is mainly due to them also being the first and last statement respectively, and were mostly just a waste of payload. You can check whether you need to wrap the statements in the `db_write` payload by checking whether the first and last statement match the above TX statements. A single call to `db_write` will _always_ correspond to a single DB transaction, so that is safe.

This was done partially because it allows us to safe some data transferred, simplifies the flush logic (empty list of statements means we don't call `db_write`, whereas before we had to check whether we only have TX begin and commit statements), and unifies the logic for other DBs that are supported, that may or may not have SQL statements to initialize and complete DB transactions, or just C API calls with no SQL representation.

SimonVrouwe:
I am closing this PR, because I am not really happy about the design, mostly the use of `sqlite_pre_init_cmds` buffer, which feels like a hack. To make it really bullet-proof, I think we need something like [https://github.com/ElementsProject/lightning/pull/3013](https://github.com/ElementsProject/lightning/pull/3013).

Or maybe a `plugin` is just not the right choice for this or even having a sync `db_write` hook for plugins was a mistake because it just doesn't fit in well enough.

Issue #41, opened 2019-07-17
autopilot breaking when no peers present. Missing dns requirement


dfresh613:
Hi, When autopilot is run with no peers present -
https://github.com/lightningd/plugins/blob/899f0a0867d2ab899ab5b31407a20d5223fe178d/autopilot/autopilot.py#L43

This call fails, due to a missing import of the random module. Upon importing random, the process fails at a further point due to no module "dns" being present.

https://github.com/lightningd/plugins/blob/899f0a0867d2ab899ab5b31407a20d5223fe178d/autopilot/autopilot.py#L25

It does not appear that there is any such dependency specified in the requirements of the autopilot plugin.

What is the "dns" module autopilot is expecting to use? I believe adding this requirement and corresponding import statement should make initial seed key retrieval complete. Allowing autopilot to run with a node with no peers connected.

I'd be glad to open a PR to resolve this, I will attempt to find which dns module we are expecting as well.

dfresh613:
Found the missing module was dnspython, and the import needed was `import dns.resolver`. 

Hit another problem further along in the process though. Looks like get_seed_keys is returning None. 

I'll look into this further when I can.

PR #42, opened 2019-07-18
autopilot functional peer searching when no peers connected

dfresh613:
- Resolves issue #41 by adding fixing the failing call attempting to iterate over `random.shuffle()`. 
- some additional error handling so that we continue onto other peers in the event one fails to connect
- missed imports encountered on the path when no peers present
- missed dependency in requirements file

### DIFF
--- a/autopilot/autopilot.py
+++ b/autopilot/autopilot.py
@@ -2,11 +2,12 @@
 
 from bech32 import bech32_decode, CHARSET, convertbits
 from lib_autopilot import Autopilot, Strategy
-from lightning import LightningRpc, Plugin
-
+from lightning import LightningRpc, Plugin, RpcError
+import random
 import math
 import networkx as nx
-
+import dns.resolver
+import time
 
 class CLightning_autopilot(Autopilot):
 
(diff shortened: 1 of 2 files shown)


autopilot/autopilot.py
```diff
@@ -2,11 +2,12 @@
 
 from bech32 import bech32_decode, CHARSET, convertbits
 from lib_autopilot import Autopilot, Strategy
-from lightning import LightningRpc, Plugin
+from lightning import LightningRpc, Plugin, RpcError
 import random
 import math
 import networkx as nx
 import dns.resolver
+import time
```
cdecker: `time` seems to be unused in the diff?

dfresh613: Time is used to add a delay when peering with nodes (currently on line 50). The call is not in this diff because it was there originally, and the necessary import was missing.

autopilot/autopilot.py
```diff
@@ -46,7 +49,7 @@ def __connect_to_seeds(self):
                 # FIXME: better strategy than sleep(2) for building up
                 time.sleep(2)
         except:
-            pass
+            raise
```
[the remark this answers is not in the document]
dfresh613: This comment seems to be on an outdated commit. This is no longer the behavior, the latest version specifically attempts to catch RPCError. I will squash the commits so that it is easier to view the change, and gets rid of unnecessary history.

dfresh613:
Commits have been squashed

cdecker:
ACK https://github.com/lightningd/plugins/pull/42/commits/d26ae7ae360cc6a1a07616c9061a50b322e028b9

PR #44, opened 2019-07-26
helpme: new plugin.

rustyrussell:
This plugin autoconnects to a few nodes, and offers some useful advice.  It's an eternal Work In Progress, but the idea is that it will expand to become all-encompassing and Know Everything!

### DIFF
--- a/helpme/helpme.py
+++ b/helpme/helpme.py
@@ -1,0 +1,1049 @@
+#! /usr/bin/env python3
+from lightning import Plugin, Millisatoshi, RpcError
+from collections import defaultdict
+from os import path
+import random
+import threading
+
+plugin = Plugin()
+
+# How many peers we want
+DESIRED_PEERS = 3
+
+# These about 1 tenth of the nodes which have been around for about 6 months.
+known_nodes_mainnet = {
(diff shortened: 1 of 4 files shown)


helpme/helpme.py
```diff
@@ -0,0 +1,1049 @@
...
+    [{"type": "ipv4", "address": "2.238.196.161", "port": 9735}],
+    '030444f1a4807f97612c8c6d5f9e4c9e75b96059dd99ccb14b6b39bb962104cf4d':
+    [{"type": "ipv4", "address": "0.0.0.0", "port": 9735}],
+    '0207a77f9012103a9e557f14e80f445946002b2752390b86554d7ca13eb1e6e0e9':
+    [{"type": "ipv4", "address": "5.102.145.58", "port": 9735}],
+    '032720f82414c8ce006d67004398b99697ae3f076c090745d6440ae3c3ade2af45':
+    [{"type": "ipv4", "address": "66.70.135.170", "port": 19735}],
+    '02b0ee0e06112db6929b1bc861d3659c608c61e5ab5899e02d417d1986ae0ab485':
+    [{"type": "ipv4", "address": "204.48.26.93", "port": 9735}]
+}
```
cdecker: Would probably be best to have this as a separate JSON file, instead of inline. That way it could get regenerated on a semi-regular basis.

PR #47, opened 2019-07-29
Add a 'requirements.txt' to probe plugin

darosior:
And remove blank lines from monitor's and persistant-channels' requirements.


SimonVrouwe:
Thanks for taking care of this, just some housekeeping comment.
I think it is helpful if commit messages have a format like:
```
plugin name: One subject line
         (empty line)
more detailed description (if any)
```

PR #50, opened 2019-07-30
helpme: Wrap in `simple` format-hint, so lightning-cli prints it as -H

SimonVrouwe:
A tiny fix to wrap `response` in `{'text': response, 'format-hint': 'simple'}`, so it looks better in lightning-cli.

### DIFF
--- a/helpme/helpme.py
+++ b/helpme/helpme.py
@@ -1,6 +1,7 @@
 #! /usr/bin/env python3
 from lightning import Plugin, Millisatoshi, RpcError
 from collections import defaultdict
+from functools import wraps
 from os import path
 import random
 import threading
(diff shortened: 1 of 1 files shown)


helpme/helpme.py
```diff
@@ -896,7 +897,16 @@ def give_history_advice(plugin):
 Senior Code Monkey."""
 
 
+# Format hint `simple` makes lightning-cli print it as (-H) human readable
+def format_simple(fn):
```
cdecker: Nice one, we might want to add that to pylightning directly so it can be used by all plugins :+1:

cdecker:
ACK https://github.com/lightningd/plugins/pull/50/commits/f14974d4d6c9dbbb94adfc97f95d579e8219b3dc

PR #55, opened 2019-09-02
fix: making sum() of funds iteration did not work on non-numeric

m-schmoock:
Small fix:

Using `sum()` on Millisatoshi does not work and throws a TypeError.
We have to do the ugly `Millisatoshi(1000 * sum([value]))` way if we want to use `sum()`.

Note: also the plugin config detection does not seem to work as it should when using the default `~/.lightning/plugins` directory. The output from the **plugin** does not make sense in this context: 
```
...
STAGE 6 (adding bling): You have not added plugins.  Try 'helpme plugins'
```

### DIFF
--- a/helpme/helpme.py
+++ b/helpme/helpme.py
@@ -451,7 +451,7 @@ node!  Be prepared to lose your funds (but please report a bug if you do!)
     if len(funds) == 0:
         r += "INCOMPLETE: No bitcoins yet.  Try 'helpme funds'"
     else:
-        funds = Millisatoshi(sum([funds['amount_msat'] for f in funds
+        funds = Millisatoshi(1000 * sum([f['value'] for f in funds
                                   if f['status'] == 'confirmed']))
         r += "COMPLETE ({} a.k.a {})".format(funds, funds.to_btc_str())
         stages['funds'] = True


helpme/helpme.py
```diff
@@ -451,7 +451,7 @@ def give_general_advice(plugin):
     if len(funds) == 0:
         r += "INCOMPLETE: No bitcoins yet.  Try 'helpme funds'"
     else:
-        funds = Millisatoshi(sum([funds['amount_msat'] for f in funds
+        funds = Millisatoshi(1000 * sum([f['value'] for f in funds
```
cdecker: Would this work if we were to define `__radd__` to the `Millisatoshi` (and `Satoshi`) type? That should add support for these kind of numeric operations :-)

m-schmoock: Yes, but we can stick with this for backwards compatiblity. I can make an additional PR on the on contrib/pylightning...

cdecker: That'd be very much appreciated, thanks ^^

cdecker:
ACK 9ccaa14fb919f35b98b03635b0c1a98a80ef23f3

helpme/helpme.py
m-schmoock: okay. see https://github.com/ElementsProject/lightning/pull/3022

PR #56, opened 2019-09-02
sendinvoiceless: round HTLC fees up

m-schmoock:
This will apply the same fee rounding error correction as done to rebalance to the sendinvoiceless plugin:

```
477bdd8 rebalance: round HTLC fees up
```


cdecker:
Feels a bit hacky, but the only alternative I can think of is converting to `float`, divide and then `ceil()`.

ACK https://github.com/lightningd/plugins/pull/56/commits/2e3c44adf17a599fa6d776a10a6c9c02b5b8aa1b

PR #57, opened 2019-09-04
rebalance: restore Python3.5 compat

darosior:
This is less readable, but compatible with Python3.5


m-schmoock:
Hm, if we do this, we need to change sendinvoiceless as well. I will do it for drain (still not ready yet).

I would recomment to use `10**6` instead of `1000000`. But what about the `999999`? :D

darosior:
> I would recomment to use 10**6 instead of 1000000. But what about the 999999? :D

`10**6 - 1` ? :-)

cdecker:
ACK https://github.com/lightningd/plugins/pull/57/commits/e31899fc8d7a16cf437ca69d0510f3d72356eef3

PR #63, opened 2019-09-30
sendinvoiceless: restore Python3.5 compat

m-schmoock:
Applies the same  py3.5 compat fix to sendinvoiceless:

94c641e rebalance: restore Python3.5 compat


cdecker:
ACK https://github.com/lightningd/plugins/pull/63/commits/e3b24b4741e3a249d3f8323221581629c40060ac

PR #66, opened 2019-11-04
jitrebalance: First version of the JITrebalance plugin

cdecker:
A simple implementation of a jit-rebalance plugin based in the idea of
@renepickhardt. Upon receiving an HTLC that we are supposed to forward check
whether the channel has sufficient capacity to forward. If that's not the case
we need to rebalance. For this we compute a circular route through some other
nodes, returning funds on the edge we are supposed to use when
forwarding. This means we rebalance the channel just enough to make forwarding
possible.

This only implements the simple version, without FOAF advertisements of free
rebalancings, but it already works rather nicely.

Suggested-by: Rene Pickhardt <@renepickhardt>
Signed-off-by: Christian Decker <@cdecker>

### DIFF
--- a/jitrebalance/jitrebalance.py
+++ b/jitrebalance/jitrebalance.py
@@ -1,17 +1,101 @@
 #!/usr/bin/env python3
 from math import ceil
-from pprint import pprint
-from pyln.client import Plugin, Millisatoshi
+from pyln.client import Plugin, Millisatoshi, RpcError
 import binascii
 import hashlib
 import secrets
+import threading
 import time
 
 plugin = Plugin()
 
 
-def get_circular_route(route, scid):
-    pass
+def get_circular_route(scid, chan, amt, peer, exclusions, request):
+    """Compute a circular route with `scid` as last leg.
+
+    """
+    # Compute the last leg of the route first, so we know the parameters to
+    # traverse that last edge.
+    reverse_chan = plugin.rpc.listchannels(scid)['channels']
+    assert(len(reverse_chan) == 2)
+    reverse_chan = [
+        c for c in reverse_chan if c['channel_flags'] != chan['direction']
+    ][0]
+
+    if reverse_chan is None:
+        print("Could not compute parameters for the last hop")
+        request.set_result({"result": "continue"})
+        return
+
+    last_amt = ceil(float(amt) +
+                    float(amt) * reverse_chan['fee_per_millionth'] / 10**6 +
+                    reverse_chan['base_fee_millisatoshi'])
+    last_cltv = 9 + reverse_chan['delay']
+
+    route = plugin.rpc.getroute(
+        node_id=peer['id'],
+        msatoshi=last_amt,
+        riskfactor=1,
+        exclude=exclusions,
+        cltv=last_cltv,
+    )['route']
+
+    # Append the last hop we computed manually above
+    route += [{
+        'id': plugin.node_id,
+        'channel': scid,
+        'direction': chan['direction'],
+        'msatoshi': amt,
+        'amount_msat': '{}msat'.format(amt),
+        'delay': 9
+    }]
+
+    return route
+
+
+def try_rebalance(scid, chan, amt, peer, request):
+    # Exclude the channel we are trying to rebalance when searching for a
+    # path. We will manually append it to the route and bump the other
+    # parameters so it can be used afterwards
+    exclusions = [
+        "{scid}/{direction}".format(scid=scid, direction=chan['direction'])
+    ]
+
+    # Try up to 5 times to rebalance that last leg.
+    for i in range(0, 5):
+        route = get_circular_route(scid, chan, amt, peer, exclusions, request)
+
+        # We're about to initiate a rebalancing, we'd better remember how we can
+        # settle it once we see it back here.
+        payment_key = secrets.token_bytes(32)
+        payment_hash = hashlib.sha256(payment_key).hexdigest()
+        plugin.rebalances[payment_hash] = {
+            "payment_key": binascii.hexlify(payment_key).decode('ASCII'),
+            "payment_hash": payment_hash,
+            "request": request,
+        }
+
+        # After all this work we're finally in a position to judge whether a
+        # rebalancing is worth it at all. The rebalancing is considered worth it
+        # if the fees we're about to pay are less than or equal to the fees we get
+        # out of forwarding the payment.
+        plugin.log("Sending rebalance request using payment_hash={}, route={}".format(
+            payment_hash, route
+        ))
+        try:
+            plugin.rpc.sendpay(route, payment_hash)
+            # If the attempt is successful, we acknowledged it on the
+            # receiving end (a couple of line above), so we leave it dangling
+            # here.
+            plugin.rpc.waitsendpay(payment_hash)
+            return
+        except RpcError as e:
+            error = e.error['data']
+            erring_channel = error['erring_channel']
+            exclusions.append(erring_channel)
+            plugin.log("Excluding {} due to a failed attempt".format(erring_channel))
+
+    request.set_result({"result": "continue"})
 
 
 @plugin.async_hook("htlc_accepted")
@@ -37,7 +121,6 @@ def on_htlc_accepted(htlc, onion, plugin, request, **kwargs):
         return
 
     # Check to see if the next channel has sufficient capacity
-
     scid = onion['short_channel_id'] if 'short_channel_id' in onion else '0x0x0'
 
     # Are we the destination? Then there's nothing to do. Continue.
@@ -66,11 +149,8 @@ def on_htlc_accepted(htlc, onion, plugin, request, **kwargs):
     # TODO If we are the funder we need to take the cost of an HTLC into
     # account as well.
     #funder = chan['msatoshi_to_us_max'] == chan['msatoshi_total']
-
     forward_amt = Millisatoshi(onion['forward_amount'])
-    incoming_amt = Millisatoshi(htlc['amount'])
-    fee = incoming_amt - forward_amt
-    pprint(fee)
+
     # Compute the amount we need to rebalance, give us a bit of breathing room
     # while we're at it (25% more rebalancing than strictly necessary) so we
     # don't end up with a completely unbalanced channel right away again, and
@@ -84,69 +164,9 @@ def on_htlc_accepted(htlc, onion, plugin, request, **kwargs):
         request.set_result({"result": "continue"})
         return
 
-    # Compute the last leg of the route first, so we know the parameters to
-    # traverse that last edge.
-    reverse_chan = plugin.rpc.listchannels(scid)['channels']
-    assert(len(reverse_chan) == 2)
-    reverse_chan = [
-        c for c in reverse_chan if c['channel_flags'] != chan['direction']
-    ][0]
-
-    if reverse_chan is None:
-        print("Could not compute parameters for the last hop")
-        request.set_result({"result": "continue"})
-        return
-
-    last_amt = ceil(float(amt) +
-                    float(amt) * reverse_chan['fee_per_millionth'] / 10**6 +
-                    reverse_chan['base_fee_millisatoshi'])
-    last_cltv = 9 + reverse_chan['delay']
-
-    # Exclude the channel we are trying to rebalance when searching for a
-    # path. We will manually append it to the route and bump the other
-    # parameters so it can be used afterwards
-    exclusions = [
-        "{scid}/{direction}".format(scid=scid, direction=chan['direction'])
-    ]
-
-    route = plugin.rpc.getroute(
-        node_id=peer['id'],
-        msatoshi=last_amt,
-        riskfactor=1,
-        exclude=exclusions,
-        cltv=last_cltv,
-    )['route']
-
-    # Append the last hop we computed manually above
-    route += [{
-        'id': plugin.node_id,
-        'channel': scid,
-        'direction': chan['direction'],
-        'msatoshi': amt,
-        'amount_msat': '{}msat'.format(amt),
-        'delay': 9
-    }]
-
-    # We're about to initiate a rebalancing, we'd better remember how we can
-    # settle it once we see it back here.
-
-    payment_key = secrets.token_bytes(32)
-    payment_hash = hashlib.sha256(payment_key).hexdigest()
-    plugin.rebalances[payment_hash] = {
-        "payment_key": binascii.hexlify(payment_key).decode('ASCII'),
-        "payment_hash": payment_hash,
-        "request": request,
-    }
-
-    # After all this work we're finally in a position to judge whether a
-    # rebalancing is worth it at all. The rebalancing is considered worth it
-    # if the fees we're about to pay are less than or equal to the fees we get
-    # out of forwarding the payment.
-
-    plugin.log("Sending rebalance request using payment_hash={}".format(
-        payment_hash
-    ))
-    plugin.rpc.sendpay(route, payment_hash)
+    t = threading.Thread(target=try_rebalance, args=(scid, chan, amt, peer, request))
+    t.daemon = True
+    t.start()
 
 
 @plugin.init()
--- a/jitrebalance/jitrebalance.py
+++ b/jitrebalance/jitrebalance.py
@@ -37,7 +37,8 @@ def on_htlc_accepted(htlc, onion, plugin, request, **kwargs):
         return
 
     # Check to see if the next channel has sufficient capacity
-    scid = onion['per_hop_v0']['short_channel_id']
+
+    scid = onion['short_channel_id'] if 'short_channel_id' in onion else '0x0x0'
 
     # Are we the destination? Then there's nothing to do. Continue.
     if scid == '0x0x0':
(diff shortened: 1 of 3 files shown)


fiatjaf:
If we had https://github.com/ElementsProject/lightning/issues/3227 could we just send the payment through an alternative route directly instead of rebalancing and then sending through the initial route?

Maybe this is all what's needed for trampoline nodes: a node that advertises itself as capable of accepting an incomplete onion and then finds a valid route to send the received onion through.

cdecker:
> If we had [ElementsProject/lightning#3227](https://github.com/ElementsProject/lightning/issues/3227) could we just send the payment through an alternative route directly instead of rebalancing and then sending through the initial route?

There might be a bit we need to do there, since re-wrapping an onion is not really possible (by re-wrapping I mean prepending hops and encrypting them according to the onion spec). The issue is, [just like it is with rendezvous routing](https://github.com/lightningnetwork/lightning-rfc/wiki/Rendez-vous-mechanism-on-top-of-Sphinx), that the ephemeral key used to derive the `shared_secret` for the next hop is derived from the one we got using a one-way derivation. So there is no way for us to make them meet up.

In order to use the same trick we do in the [rendezvous proposal](https://github.com/lightningnetwork/lightning-rfc/wiki/Rendez-vous-mechanism-on-top-of-Sphinx) we'd need the ability to modify the next hop's payload to add the swap ephemeral key, but we can't modify that since it is encrypted with their shared secret, not ours.

There migh be way to do it with a dummy hop however. Need to think about this a bit more.

cdecker:
> Maybe this is all what's needed for trampoline nodes: a node that advertises itself as capable of accepting an incomplete onion and then finds a valid route to send the received onion through.

Yes, you don't really need anything else than `htlc_accepted` + `sendpay` for trampoline nodes other than a plugin that can "complete" a transfer. I don't see any issue here. The sender needs the ability to inject an onion however, which is ElementsProject/lightning#3227, and I'll spend a few cycles to get that implemented soon :wink

jitrebalance/jitrebalance.py
```diff
@@ -0,0 +1,160 @@
...
+
+    # After all this work we're finally in a position to judge whether a
+    # rebalancing is worth it at all. The rebalancing is considered worth it
+    # if the fees we're about to pay are less than or equal to the fees we get
+    # out of forwarding the payment.
+
+    plugin.log("Sending rebalance request using payment_hash={}".format(
+        payment_hash
+    ))
+    plugin.rpc.sendpay(route, payment_hash)
```
renepickhardt: what if sendpay still does not work because our circular onion does not have enogh liquidity on all edges? 
Wouldn't it make more sense to start a loop at line 111 (`getroute` api call)?

jitrebalance/jitrebalance.py
```diff
@@ -0,0 +1,160 @@
...
+
+    payment_key = secrets.token_bytes(32)
+    payment_hash = hashlib.sha256(payment_key).hexdigest()
+    plugin.rebalances[payment_hash] = {
+        "payment_key": binascii.hexlify(payment_key).decode('ASCII'),
+        "payment_hash": payment_hash,
+        "request": request,
+    }
+
+    # After all this work we're finally in a position to judge whether a
```
renepickhardt: while I understand the semantics of the comment I don't see this anywhere enforced in the code.  

I thought for a second that is done implicitly in line 124 by adding the `amt` field which adds the fees in line 99 to 102 but there is no comparison to the routing fees that we get on that channel.

jitrebalance/jitrebalance.py
```diff
@@ -0,0 +1,160 @@
...
+import secrets
+import hashlib
+import binascii
+import time
+
+plugin = Plugin()
+pdb = RemotePdb('127.0.0.1', 4444)
+
+
+def get_circular_route(route, scid):
```
renepickhardt: what does that empty function do? It is also not used anywhere in the code

jitrebalance/jitrebalance.py
```diff
@@ -0,0 +1,160 @@
...
+    pass
+
+
+@plugin.async_hook("htlc_accepted")
```
renepickhardt: so as the hook is async how do you make sure that the HTLC is only attempted to forwarded if rebalancing was succesfull? In particular: see my comment in line 148. Wouldn't it make sense to have more than 1 rebalancing attempt?

jitrebalance/jitrebalance.py
```diff
@@ -0,0 +1,160 @@
...
+    forward_amt = Millisatoshi(onion['per_hop_v0']['forward_amount'])
+    incoming_amt = Millisatoshi(htlc['amount'])
+    fee = incoming_amt - forward_amt
+    pprint(fee)
+    # Compute the amount we need to rebalance, give us a bit of breathing room
+    # while we're at it (25% more rebalancing than strictly necessary) so we
+    # don't end up with a completely unbalanced channel right away again, and
+    # to account for a bit of fuzziness when it comes to dipping into the
+    # reserve.
+    amt = ceil(int(forward_amt - chan['spendable_msat']) * 1.25)
```
renepickhardt: this will probably come with experience or by looking at past data of this channel. But I guess rebalancing the channel so that it is 50:50 afterwards might also be a plausible heuristic. 

you did not check in this case that your amt stays beyond 4.2M satoshi the max pay amt

jitrebalance/jitrebalance.py
```diff
@@ -0,0 +1,160 @@
...
+        'direction': chan['direction'],
+        'msatoshi': amt,
+        'amount_msat': '{}msat'.format(amt),
+        'delay': 9
+    }]
+
+    # We're about to initiate a rebalancing, we'd better remember how we can
+    # settle it once we see it back here.
+
+    payment_key = secrets.token_bytes(32)
```
renepickhardt: there was the idea floating around to make rebalancing to the same payment hash. but of course this would also confuse the peer as he might have two htlcs in oposing directions and it might make the logic more difficult here

jitrebalance/jitrebalance.py
darosior: This does not seem implemented yet.

I don't think that it would be profitable for now tbh.

jitrebalance/jitrebalance.py
darosior: > so as the hook is async how do you make sure that the HTLC is only attempted to forwarded if rebalancing was succesfull?

The async trick is only Pylightning side, to be able to "return" a result to `lightningd` without having to actually return in the plugin logic (see https://github.com/lightningd/plugins/pull/66/files#diff-78896cd8fe0da553087891ebc6efda26R30-R39)

> Wouldn't it make sense to have more than 1 rebalancing attempt?

Agree, even if I think it's intended. We could check all potential routes as in rebalance and drain plugins, and we could reduce the `1.25` factor on the rebalanced amount on capacity failure.

jitrebalance/test_jitrebalance.py
```diff
@@ -0,0 +1,78 @@
...
+    # Drain (l2, l3) so that a larger payment fails later on
+    chan = l2.rpc.listpeers(l3.info['id'])['peers'][0]['channels'][0]
+    pprint(chan)
+    # Send 9 million millisatoshis + reserve + a tiny fee allowance from l3 to
+    # l2 for the actual payment
+    inv = l2.rpc.invoice(
+        chan['our_channel_reserve_satoshis']*1000 + 9000000 + 100,
+        "imbalance", "imbalance"
+    )
+    time.sleep(1)
```
darosior: Why ?

jitrebalance/test_jitrebalance.py
```diff
@@ -0,0 +1,78 @@
...
+    channels = [(l1, l2), (l3, l2), (l3, l4), (l2, l5), (l5, l3)]
+    for src, dst in channels:
+        src.openchannel(dst, capacity=10**6)
+
+    # Drain (l2, l3) so that a larger payment fails later on
+    chan = l2.rpc.listpeers(l3.info['id'])['peers'][0]['channels'][0]
+    pprint(chan)
+    # Send 9 million millisatoshis + reserve + a tiny fee allowance from l3 to
+    # l2 for the actual payment
+    inv = l2.rpc.invoice(
```
darosior: There is something I don't get with this test : shouldn't we drain __from__ `l2` ?

If I follow the docstring :
```
l2 however will attempt to rebalance (l2,l3) using a circular route (l2, l5, l3, l2) to get the required funds back.
```
then l2 would try to send a payment through `l3 => l2` while it's already exhausted ?

darosior:
@renepickhardt 
> How difficult would it be to provide an API via c-lightning plugin system so that lightning messages could be send (maybe only odd ones!) in that way plugin developers could experiment with new features like the query_foaf_balances in that way we could have a network of nodes that have JIT enabled. also in case we don't have that we could probe channels for balance (technically by doing many get_route calls concurrently.

That's `pyln-proto` ! :-)
There is for now only connection messages (BOLTS 1 and 8) if I don't make a mistake implemented, but I've planned for some time to implement BOLTS 2 and 3. It's on my TODOs...

jitrebalance/jitrebalance.py
```diff
@@ -0,0 +1,160 @@
...
+    plugin.rebalances[payment_hash] = {
+        "payment_key": binascii.hexlify(payment_key).decode('ASCII'),
+        "payment_hash": payment_hash,
+        "request": request,
+    }
+
+    # After all this work we're finally in a position to judge whether a
+    # rebalancing is worth it at all. The rebalancing is considered worth it
+    # if the fees we're about to pay are less than or equal to the fees we get
+    # out of forwarding the payment.
```
m-schmoock: But we are not doing this decision yet, or do we?

Just thinking:
Additionally, even if we do this correctly, a mean node might know about our JIT rebalance feature node enough so he will get us to 'drain' our other 'cheap enough' channels, leaving all the capacity with the higher fee channels. If he controls these channels, he can then readjust the fees to force us overpaying the next time we want to use our channels, right?

darosior: > a mean node might know about our JIT rebalance feature node enough

Hmm .. Interesting but how would it ? If you don't disclose it in your alias that you use jitrouting.

>  If he controls these channels

It would have to control most of our channels, then ? Both ones to drain and ones used to rebalance. Here the Sybil att would have a certain cost if you are a minimum well connected ?

m-schmoock: I know this is all very hypothetical and complicated effort for steeling on fees, and I'm not even sure if this es economically viable at least since the victim can choose to close the extorted channel and the attacker has costs upfront on routing and such, but its a thought experiment:

1. Figuring out if a node uses JIT rebalancing is rather easy, at least by sending two circular payments that would not fit both through the same channel at the vitctim. if it works, its likely the node uses some rebalancing algo. There may also be other options, where the attacker has a testing channel with very low fees to the victim to bias rebalancing through this testchannel, ...

2. Sybil attacking this is also not hard, as the attacker overpays on fees until all the victims honest channels are empty, except his channels and one would be enough in theory.

cdecker:
Fixed the plugin to work against TLV onions. Will address the feedback by you guys asap.

jitrebalance/test_jitrebalance.py
```diff
@@ -0,0 +1,79 @@
...
+from pyln.testing.utils import wait_for
+from pprint import pprint
+import os
+import time
+
+
+plugin = os.path.join(os.path.dirname(__file__), 'jitrebalance.py')
+
+
+@unittest.skipIf(not DEVELOPER, "gossip is too slow if we're not in developer mode")
```
m-schmoock: As Travis points out, there is missing dependencies for this annotation:

https://travis-ci.org/lightningd/plugins/jobs/620908677
```
jitrebalance/test_jitrebalance.py:11: in <module>
    @unittest.skipIf(not DEVELOPER, "gossip is too slow if we're not in developer mode")
E   NameError: name 'unittest' is not defined
```

cdecker:
Finally got around to fix this thing up a bit, splitting it into functions and implement the retry logic :wink:

jitrebalance/jitrebalance.py
```diff
@@ -0,0 +1,160 @@
...
+    if not peer['connected'] or chan['state'] != "CHANNELD_NORMAL":
+        request.set_result({"result": "continue"})
+        return
+
+    # Need to consider who the funder is, since they are paying the fees.
+    funder = chan['msatoshi_to_us_max'] == chan['msatoshi_total']
+    forward_amt = Millisatoshi(onion['per_hop_v0']['forward_amount'])
+    incoming_amt = Millisatoshi(htlc['amount'])
+    fee = incoming_amt - forward_amt
+    pprint(fee)
```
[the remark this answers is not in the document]
cdecker: When you call `plugin.run()` it'll replace the `stdout` file-descriptor, so anything that writes to `stdout` will get intercepted and wrapped in valid `log` notifications. This works, but relies a bit on the magic introduced by `pyln-client` :-)

jitrebalance/jitrebalance.py
```diff
@@ -0,0 +1,160 @@
...
+
+@plugin.async_hook("htlc_accepted")
+def on_htlc_accepted(htlc, onion, plugin, request, **kwargs):
+    pdb.set_trace()
+    plugin.log("Got an incoming HTLC htlc={}, onion={}".format(htlc, onion))
+
+    # The HTLC might be a rebalance we ourselves initiated, better check
+    # against the list of pending ones.
+    rebalance = plugin.rebalances.get(htlc['payment_hash'])
+    if rebalance is not None:
```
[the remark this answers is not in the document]
cdecker: You caught a memory leak ^^

jitrebalance/jitrebalance.py
cdecker: With the recent addition of AMP this is not an issue with c-lightning anymore. But plugins may still get some things wrong. The same hash would leak the purpose and allow coalescing the rebalancing with the payment that caused it, but that's a minor issue imho.

jitrebalance/jitrebalance.py
cdecker: Added a retry of up to 5 attempts, with tracking of failing channels so we actually do try different routes.

jitrebalance/test_jitrebalance.py
cdecker: We just accepted our own rebalancing request, which still needs to be cleared back to us through the circular route, only after that the funds will be available for the forwarding request.

We could be more clever here, and wait for the `waitsendpay` to complete and allow the forwarding to continue. The `sleep(1)` is only a stopgap for now.

jitrebalance/test_jitrebalance.py
cdecker: We don't have capacity in the direction `(l2, l3)`, so we create a circular route `(l2,l5, l3, l2)` (opposite direction of what we are trying to rebalance). `(l2, l3)` is exhausted which is why we try the rebalancing, if `(l3, l2)` is also exhausted we can never rebalance correctly, so any attempt would fail.

darosior: thanks

cdecker:
Ping @renepickhardt & @darosior, how does this look now?

jitrebalance/jitrebalance.py
```diff
@@ -0,0 +1,184 @@
...
+        # Settle the rebalance, before settling the request that initiated the
+        # rebalance.
+        request.set_result({
+            "result": "resolve",
+            "payment_key": rebalance['payment_key']
+        })
+
+        # Now wait for it to settle correctly
+        # TODO Maybe be a bit smarter than having a fixed timeout here?
```
darosior: Maybe we could generate an invoice when initiating the rebalance, then listen for `invoice_payment` ? Or maybe it would be even  slower..

jitrebalance/jitrebalance.py
```diff
@@ -0,0 +1,184 @@
...
+    for p in peers:
+        for c in p['channels']:
+            if c['short_channel_id'] == scid:
+                chan = c
+                peer = p
+
+    # Check if the channel is active and routable, otherwise there's little
+    # point in even trying
+    if not peer['connected'] or chan['state'] != "CHANNELD_NORMAL":
+        request.set_result({"result": "continue"})
```
darosior: Aside: Could we test using another outgoing channel instead ? I'm planning to do a probe-then-forward plugin and that would be the exact case.

jitrebalance/jitrebalance.py
```diff
@@ -0,0 +1,184 @@
...
+        return
+
+    # Need to consider who the funder is, since they are paying the fees.
+    # TODO If we are the funder we need to take the cost of an HTLC into
+    # account as well.
+    #funder = chan['msatoshi_to_us_max'] == chan['msatoshi_total']
```
darosior: trailing debug line :)

jitrebalance/jitrebalance.py
cdecker: Uhm, we can't really redirect an HTLC to another node (and we don't have multi-channels with a single peer) since the onion is only decodeable by the correct next peer.

But we can probe and do any shenannigans we want before forwarding :-)

jitrebalance/jitrebalance.py
cdecker: Hm, we could also just wait using `waitsendpay` or listen to the `sendpay_success` notification. I was just lazy :wink:

jitrebalance/jitrebalance.py
cdecker: Yeah, I kept it there so I remember to eventually do something with the funder and compute the fees more precisely.

PR #67, opened 2019-11-11
drain: cleanups

m-schmoock:
... sorry, i just sit down and made the cleanup commit a little too late.

This fixes:
 - division by zero when gossip was not up to date yet.
 - uninitialized variable payload on read_params when asserting too early.


cdecker:
ACK https://github.com/lightningd/plugins/pull/67/commits/1a294bceca6c6e106e7c5b0c8fd83b70693d351c

PR #70, opened 2019-12-01
ZeroMQ publisher plugin

jarret:
A plugin that subscribes to notification and passes them through to be published to ZeroMQ endpoints. It utilizes Twisted and txZMQ to facilitate the publication. An additional script is provided to serve as an example for how to subscribe to the events.

The presentation was chosen to follow how `bitcoind` sets up its ZMQ publishing functionality.

NOTE: The plugin source file is given the name `cl-zmq.py` as opposed to `zmq.py` because the package dependencies import the official ZeroMQ python packages via `from zmq import ...` which is a namespace conflict with a source file named `zmq.py`.

### DIFF
--- a/zmq/cl-zmq.py
+++ b/zmq/cl-zmq.py
@@ -1,0 +1,206 @@
+#!/usr/bin/env python3
+# Copyright (c) 2019 lightningd
+# Distributed under the BSD 3-Clause License, see the accompanying file LICENSE
+
+###############################################################################
+# ZeroMQ publishing plugin for lightningd
+#
+# Using Twisted and txZMQ frameworks, this plugin binds to ZeroMQ endpoints and
+# publishes notification of all possible subscriptions that have been opted-in
+# for via lightningd launch parameter.
+#
+# This plugin doesn't interpret any of the content of the data which comes out
+# of lightningd, it merely passes the received JSON through as encoded UTF-8,
+# with the 'tag' being set to the Notification Type name (also encoded as
+# UTF-8). It follows that adding future possible subscriptions *should* be as
+# easy as appending it to NOTIFICATION_TYPE_NAMES below.
+#
+# The user-selectable configuration takes inspiration from the bitcoind ZeroMQ
+# integration. The endpoint must be explicitly given as an argument to enable
+# it. Also, the high water mark argument for the binding is set as an
+# additional launch option.
+#
+# Due to how the plugins must register via getmanifest, this will opt-in to all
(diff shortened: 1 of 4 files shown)


zmq/cl-zmq.py
```diff
@@ -0,0 +1,207 @@
...
+    reactor.callFromThread(publisher.publish_notification, "{}",
+                           *args, **kwargs)
+"""
+
+DEFAULT_HIGH_WATER_MARK = 1000
+
+for nt in NOTIFICATION_TYPES:
+    # subscribe to all notifications
+    subscribe = SUBSCRIBE_TEMPLATE.format(nt, nt, nt, nt)
+    exec(subscribe)
```
cdecker: This seems rather dangerous. I'd suggest creating a generic handler, use `functools.partial` to generate variants with the notification type bound and then using `plugin.add_subscription` to register them.

Something like the following would do I think:

```python
def on_notification(notification_type, plugin, *args, **kwargs):
    if len(args) != 0:
￼        plugin.log("got unexpected args: {}".format(args), level="warn")
￼    reactor.callFromThread(publisher.publish_notification, "{}",
￼                           *args, **kwargs)

for nt in NOTIFICATION_TYPES:
    f = functools.partial(on_notification, nt)
    plugin.add_subscription(nt, f)
```

This would not call the python interpreter with a variable payload. In this case it's likely ok with your mechanism, since it just fills in the notification types, but in other cases this could be a security vulnerability since you execute code that was provided from outside.

jarret:
Great suggestion. Defining code in a string also is kind of ugly and `functools.partial()` appears to be suited.

One issue I bumped into in the revised implementation is that `Plugin._coerce_arguments()` expects the callable object to have the `__annotations__` attribute set, for which the partial object does not. To fix, it had to be manually set to an empty dict:

```
on.__annotations__ = {} # needed to please Plugin._coerce_arguments()

```

cdecker:
Good catch, I'll add a check that should make that no longer necessary.

cdecker:
I have patched `pyln-client` to check for the existence of `__annotations__` before accessing it here https://github.com/ElementsProject/lightning/commit/dfa7f48d21e7c65afca47a75e65e148236892978. Sadly it didn't make the v0.8.0 release window, so it'll be available from 0.8.1 onwards :-)

cdecker:
It might be interesting to write a couple of tests by the way so that we don't accidentally break compatibility while working on c-lightning :wink:

PR #71, opened 2019-12-05
test drain plugin

m-schmoock:
@cdecker  This PR is (currently) for me if ln plugins and travis is working for me. Ignore it until I finished my code...

### DIFF
--- a/drain/test_drain.py
+++ b/drain/test_drain.py
@@ -1,0 +1,240 @@
+from pyln.testing.fixtures import *
+from pyln.testing.utils import DEVELOPER
+from pyln.client import RpcError, Millisatoshi
+from utils import *
+import unittest
+
+
+pluginopt = {'plugin': os.path.join(os.path.dirname(__file__), "drain.py")}
+
+
+@unittest.skipIf(not DEVELOPER, "slow gossip, needs DEVELOPER=1")
+def test_drain_and_refill(node_factory, bitcoind):
+    # SETUP: A basic circular setup to run drain and fill tests
+    #
+    #   l1---l2
+    #    |    |
+    #   l4---l3
+    #
+
+    l1, l2, l3, l4 = node_factory.line_graph(4, opts=pluginopt)
+    l4.rpc.connect(l1.info['id'], 'localhost', l1.port)
+
+    scid12 = l1.get_channel_scid(l2)
+    scid23 = l2.get_channel_scid(l3)
+    scid34 = l3.get_channel_scid(l4)
+    scid41 = l4.fund_channel(l1, 10**6)
+
+    # wait for each others gossip
+    bitcoind.generate_block(6)
+    for n in [l1,l2,l3,l4]:
+        for scid in [scid12,scid23,scid34,scid41]:
+            n.wait_channel_active(scid)
+
+
+    # do some draining and filling
+    ours_before = get_ours(l1, scid12)
+    assert(l1.rpc.drain(scid12))
+    ours_after = wait_ours(l1, scid12, ours_before)
+    assert(ours_after < ours_before * 0.05)  # account some reserves
+
+    # refill again with 100% should not be possible in a line_graph circle,
+    # as we also paid some fees earlier and lost a tiny bit of total capacity.
+    with pytest.raises(RpcError, match=r"Outgoing capacity problem"):
+        l1.rpc.fill(scid12)
+
+    # refill to 99.9% should work however
+    assert(l1.rpc.fill(scid12, 99.9))
+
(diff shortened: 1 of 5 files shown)


drain/test_drain.py
```diff
@@ -0,0 +1,42 @@
...
+
+    wait_for(lambda: [c['active'] for c in l3.rpc.listchannels(scid1)['channels']] == [True, True])
+    wait_for(lambda: [c['active'] for c in l3.rpc.listchannels(scid2)['channels']] == [True, True])
+    wait_for(lambda: [c['active'] for c in l3.rpc.listchannels(scid3)['channels']] == [True, True])
+    wait_for(lambda: [c['active'] for c in l3.rpc.listchannels(scid4)['channels']] == [True, True])
+
+    wait_for(lambda: [c['active'] for c in l4.rpc.listchannels(scid1)['channels']] == [True, True])
+    wait_for(lambda: [c['active'] for c in l4.rpc.listchannels(scid2)['channels']] == [True, True])
+    wait_for(lambda: [c['active'] for c in l4.rpc.listchannels(scid3)['channels']] == [True, True])
+    wait_for(lambda: [c['active'] for c in l4.rpc.listchannels(scid4)['channels']] == [True, True])
```
cdecker: You can likely check that `l1.rpc.listchannels() == n.rpc.listchannels()` for all `n in [l2,l3,l4]` and be done with it :-)

drain/test_drain.py
```diff
@@ -0,0 +1,42 @@
+from utils import wait_for
+from pyln.testing.fixtures import *
+
+pluginopt = {'plugin': os.path.join(os.path.dirname(__file__), "drain.py")}
+
+def test_drain(node_factory):
+    # build a simple circular setup to run tests on
+    # l1, l2, l3, l4 = node_factory.line_graph(4, announce_channels=True, wait_for_announce=True, opts=pluginopt)
+    # ValueError: ('Error waiting for {}', <function NodeFactory.line_graph.<locals>.<lambda> at 0x7f299deb67b8>)
+    l1, l2, l3, l4 = node_factory.line_graph(4, announce_channels=True, opts=pluginopt)
```
cdecker: This might require the `@unittest.skipIf(not DEVELOPER)` annotation on the test since otherwise the gossip sync is too slow ^^

drain/test_drain.py
m-schmoock: @cdecker 

thanks :D

Im just messing around... ln tests 'utils' module we have for main repo is not available, so no `wait_for` anyway...

Also I think I discovered a but in linear_graph together with wait_for_announce True..

m-schmoock: Also, Im not shure what `l1.rpc.listchannels() == n.rpc.listchannels()` actually tests. Is the `listchannels` output sorted equally for all nodes?

cdecker:
Sorry for the long radio silence on this one. Is this still WIP or can we rebase and merge it?

m-schmoock:
Hi @cdecker you don't need to be sorry, as I'm not yet happy with the test. It discloses what I assume to be a bug or an issue that I don't understand. See https://github.com/lightningd/plugins/pull/71/files#diff-88f5d9948c5e37be5dc85a7dfbfc4503R143 and following lines refill the channel to just 97% in order to have the test green.

If you want to have a look, feel free...

m-schmoock:
I really need to get the daemon to give me an estimation about something like `receivable_msat` analog to `spendable_msat`. Currently I can only fix this by not allowing to fill 100% but rather cap it at i.e. 97%.

This is technically not correct, which leads to all these issues I have:
https://github.com/lightningd/plugins/blob/ba5500ddd8f579b0860c31d5c1f7e378b7a86713/drain/drain.py#L100

m-schmoock:
Okay, now I finally know why testing chunk autodetection fails.
Will solve this soonish so we can merge and close this.

m-schmoock:
Note: There are other test failing (non-drain, i.e. backup)
Other than that, this PR should be done now. Sorry for the delay ;)

m-schmoock:
rebased to current master, maybe the failing backup tests are done now ;) lets wait and see

cdecker:
Yep, main reason for my weekend fixes was to get a better signal for PRs 🙂 I'll review this asap

cdecker:
ACK https://github.com/lightningd/plugins/pull/71/commits/2aeb22c0da8ebd1c9161a20c087b0d661215a4e8

Excellent work @m-schmoock, there are a couple of things that could be more pythonic, but I'll make a followup PR and fix them up :+1:

PR #72, opened 2019-12-11
Add an 'exclude' parameter to the summary plugin

darosior:


summary/summary.py
```diff
@@ -126,6 +128,8 @@ def summary(plugin):
             if c['state'] != 'CHANNELD_NORMAL':
                 continue
             active_channel = True
+            if c['short_channel_id'] in exclude:
```
darosior: Not part of the condition above as we need `active_channel = True`.

summary/summary.py
```diff
@@ -89,8 +92,8 @@ def append_header(table, max_msat):
             % (draw.left, short_str, draw.mid, short_str, draw.right))
 
 
-@plugin.method("summary")
-def summary(plugin):
+@plugin.method("summary", long_desc=summary_description)
+def summary(plugin, exclude=''):
```
cdecker: I was planning to split the docstring into first line for the short description and the lines after the blank lines for the long description. Do you think that's a good idea? It'd remove the need to provide the `long_desc` in an extra arg to `plugin.method()`

darosior: I like to keep the docstring to describe what the function does and the `desc`/`long_desc` to describe the command `lightningd`-side (I've even completely separated, for instance, the descriptions in [reckless](https://github.com/darosior/reckless/blob/master/descriptions.py)). But maybe I should use another function for the logic, and keep the decorated function's docstring as the command description..

cdecker: Maybe you're right. The docstring stuff is already quite magic and might not work well for everybody. Let's keep the magic minimal for now :+1:

PR #74, opened 2019-12-16
Basic starting tests for everyone

darosior:
I first wanted to test if all dynamic plugins really supported to be dynamic, and it finally got to be basic tests for all plugins: after a `pip install` party I can attest that all plugins are still working :grin:. They still need functional tests but that's at least a basis.

It's (really) redundant so I can stash at will.


cdecker:
ACK https://github.com/lightningd/plugins/pull/74/commits/982ae77933f89d5a2db15eac05ee8360ad1c9118

PR #77, opened 2019-12-24
Update to pyln-client

darosior:
--Still need to update all that don't have requirements yet.--

Updated to add a requirements.txt to plugins that didnt have one, and updated all plugins to use `pyln-client`.
(And a drive-by fix to add @0xB10C's  `csvexportpays` to the table :) )


cdecker:
ACK https://github.com/lightningd/plugins/pull/77/commits/c8ab37b33dc25d5d246590f56ccd96e2ea81017e

PR #78, opened 2019-12-24
Update lightning-qt (second take)

darosior:
Sorry for the double update in a row: I'm playing with reckless and lightning-qt is my guinea pig. This version fixes the final quirks.


cdecker:
ACK https://github.com/lightningd/plugins/pull/78/commits/7ced699f27a00e9a48d3af34bb0606901caa0de9

PR #82, opened 2020-01-12
WIP: A new plugin to find good peers | explore the LN 

darosior:
The algorithm is a bit sketchy but it can be improved over time, and I already made some good finds with the plugin.

### DIFF
--- a/goodpeer/descriptions.py
+++ b/goodpeer/descriptions.py
@@ -8,4 +8,8 @@ Get a list of good enough peers to connect to, reversely sorted.
 you plan to make small payments. It will tweak the score computation
 towards prefering small ppm fees (big) or small base fees (small). By
 default no tweak is applied and they have the same weight.
+
+{min_height} is useful to set because, by default, the fees stats will be
+calculated upon all the active channels of a given node, which may have
+changed its fees policy since.
 """
--- a/goodpeer/descriptions.py
+++ b/goodpeer/descriptions.py
@@ -1,0 +1,11 @@
+goodpeers_small_desc = "Get a list of good enough peers to connect to, reversely sorted."
+
+goodpeers_desc = \
+"""
+Get a list of good enough peers to connect to, reversely sorted.
+
+{bias} can be set to "big" if you plan to make big payments or "small" if
+you plan to make small payments. It will tweak the score computation
+towards prefering small ppm fees (big) or small base fees (small). By
+default no tweak is applied and they have the same weight.
+"""
--- a/goodpeer/goodpeer.py
+++ b/goodpeer/goodpeer.py
@@ -46,10 +46,10 @@ def filter_nodes(nodes):
     """
     Filter out nodes with <=15 non-dust channels.
 
-    20$ minimum for a bitcoin at 8000$ (250000000 msats).
+    55$ minimum for a bitcoin at 8500$ (650000000 msats).
     """
     filtered_nodes = {}
-    dust_limit = Millisatoshi(250000000)
+    dust_limit = Millisatoshi(650000000)
     for id, channs in nodes.items():
         if len(list(filter(lambda c: c["value"] > dust_limit, channs))) > 15:
             filtered_nodes[id] = channs
--- a/goodpeer/goodpeer.py
+++ b/goodpeer/goodpeer.py
@@ -56,9 +56,10 @@ def filter_nodes(nodes):
     return filtered_nodes
 
 
-def get_medians(channels):
+def get_stats(channels):
     """
-    Return the (median_base_fee, median_fees_ppm) of the given peer.
+    Return the (median_base_fee, median_fees_ppm, avg_base_fee, avg_fees_ppm)
+    of the given peer.
     """
     # Loop only once..
     base_fees = ppm_fees = []
@@ -68,13 +69,16 @@ def get_medians(channels):
     base_fees, ppm_fees = sorted(base_fees), sorted(ppm_fees)
     n_channs = len(channels)
     if n_channs % 2 == 0:
-        return base_fees[n_channs // 2], ppm_fees[n_channs // 2]
+        median_base = base_fees[n_channs // 2]
+        median_ppm = ppm_fees[n_channs // 2]
     else:
         median_base = (base_fees[n_channs // 2] +
                        base_fees[n_channs // 2 + 1]) / 2
         median_ppm = (ppm_fees[n_channs // 2] +
                       ppm_fees[n_channs // 2 + 1]) / 2
-        return median_base, median_ppm
+    avg_base = sum(base_fees) / len(base_fees)
+    avg_ppm = sum(ppm_fees) / len(ppm_fees)
+    return median_base, median_ppm, avg_base, avg_ppm
 
 
 def reverse_sort(nodes):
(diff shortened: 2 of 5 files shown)


goodpeer/goodpeer.py
```diff
@@ -0,0 +1,113 @@
...
+def get_nodes(plugin):
+    """
+    Get public Lightning network nodes.
+
+    :return: A dict with the node id as key, and a list of channels as values.
+    """
+    peers = {}
+    channels = plugin.rpc.listchannels()
+    for chan in channels["channels"]:
+        if chan["fee_per_millionth"] == chan["base_fee_millisatoshi"] == 0:
```
cdecker: Interesting, I didn't know that this sort of transitive equality works in python, I'd have expected the resolution to be the following:

```
0 == 0 == 0
True == 0
False
```

But it seems to work when I test it locally. Nice find :+1:

goodpeer/goodpeer.py
```diff
@@ -0,0 +1,113 @@
...
+            # distort the medians.
+            continue
+        if not chan["active"]:
+            # It could just be temporary, but we don't want zombies!
+            continue
+        id = chan["source"]
+        if id not in peers.keys():
+            peers[id] = [get_channel_infos(chan)]
+        else:
+            peers[id].append(get_channel_infos(chan))
```
cdecker: This can usually be simplified to this:

```python
if id not in peers:
    peers[id] = []
peers[id].append(get_channel_infos(chan))
```

Or even more concise:

```python
peers[id] = peers.get(id, []).append(get_channel_infos(chan))
```

though the second one is a bit less legible imho :wink:

goodpeer/goodpeer.py
```diff
@@ -0,0 +1,113 @@
...
+    dust_limit = Millisatoshi(250000000)
+    for id, channs in nodes.items():
+        if len(list(filter(lambda c: c["value"] > dust_limit, channs))) > 15:
+            filtered_nodes[id] = channs
+    return filtered_nodes
+
+
+def get_medians(channels):
+    """
+    Return the (median_base_fee, median_fees_ppm) of the given peer.
```
cdecker: Which direction are you considering here? Or does that not matter? I would have expected merchants to look for nodes with a small incoming fee, and end-users for a small outgoing fee.

goodpeer/goodpeer.py
```diff
@@ -0,0 +1,113 @@
...
+def reverse_sort(nodes):
+    """
+    Reversely sort the nodes, the lower the score the better.
+    """
+    # Python3.5 hasn't native dicts ordered.. So use OrderedDict
+    sorted_nodes = OrderedDict()
+    for k, v in sorted(nodes.items(),
+                       key=lambda item: item[1]["score"], reverse=True):
+        sorted_nodes[k] = v
+    return sorted_nodes
```
cdecker: Could be simplified to

```python
OrderedDict(sorted(nodes.items(), key=lambda item: -item[1]["score"]))
```

Though it's a bit less descriptive.

goodpeer/goodpeer.py
darosior: Thanks!

goodpeer/goodpeer.py
darosior: Ah right, I was only considering end-user here !

goodpeer/goodpeer.py
darosior: I prefer it your way, why did you add the minus sign before the score though ?

goodpeer/goodpeer.py
darosior: (Considered outgoing fees)

darosior:
> Seeing that this overlaps in scope with the autopilot plugin, would it make sense to merge the two?

Ah I've never read the autopilot. Since this is just part of it maybe it can just be duplicated ? I'd like to use this without autopilot, maybe so do other users.

goodpeer/goodpeer.py
darosior: Yep it's particularly cool for cases like
```python
if x < y < z:
```

goodpeer/goodpeer.py
darosior: I added another optional parameter

goodpeer/goodpeer.py
darosior: Fixed, I just kept the reverse=True instead of the minus :-)

goodpeer/goodpeer.py
darosior: Actually the `direction` argument doesn't make that much sense, as when scanning nodes I take the fees *this node* sets (`source`), taking the fees its peers set (`destination`) doesn't seem relevant.

> I would have expected merchants to look for nodes with a small incoming fee, and end-users for a small outgoing fee.

Re-thinking it I had outgoing fees in mind in any case, and assumed that if we fund a channel to a peer which always set somewhat small outgoing fees, it would set it to us too. So that we can route through it, and receive through it.

darosior:
*Removed the just added `direction` parameter*

cdecker:
> > Seeing that this overlaps in scope with the autopilot plugin, would it make sense to merge the two?
> 
> Ah I've never read the autopilot. Since this is just part of it maybe it can just be duplicated ? I'd like to use this without autopilot, maybe so do other users.

The autopilot doesn't really do anything unless you call the exposed RPC commands, and even most of the commands are just suggestions, kind of a discovery mechanism. This is pretty similar in scope to that, that's why I'm suggesting it.

It's just a bit confusing for users to have multiple options to achieve the same goal, I think.

darosior:
I had to (re) open channels recently, which permitted me to tweak the results:
- The filters were too low. 20$ isnt enough
- The average fees can be a secondary metric
- Add the possibility to filter out old channels, fees policy may have changed since

darosior:
Late reply: I need to read the autopilot to integrate it, but I've thought a bit more about that and I think I'm going to extend this to be more kind of an "explorer" as soon as I can. Marking WIP then.

goodpeer/descriptions.py
```diff
@@ -0,0 +1,15 @@
+goodpeers_small_desc = "Get a list of good enough peers to connect to, reversely sorted."
+
+goodpeers_desc = \
+"""
```
m-schmoock, line 4: Nit:  E122 continuation line missing indentation or outdented

goodpeer/goodpeer.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+    # Python3.5 hasn't native dicts ordered.. So use OrderedDict
+    return OrderedDict(sorted(nodes.items(),
+                       key=lambda item: item[1]["score"], reverse=True))
+
+
+@plugin.method("goodpeers", desc=goodpeers_small_desc,
+               long_desc=goodpeers_desc)
+def goodpeers(plugin, bias=None, min_height=None, **kwargs):
+    nodes = get_nodes(plugin)
```
m-schmoock, line 103: We should use the docstring here, maybe instead of the `desc_long=...` above.

m-schmoock (approved):
ACK. Tested and works as promised.
This has been dormant for a while, are we go with this?
Anyway some nits:

I have a docstring warning in the RPC method:
```
INFO    plugin-goodpeer.py: RPC method 'goodpeers' does not have a docstring.
```

Additional we have some W504 nits: 
```
goodpeer/goodpeer.py:55:45: W504 line break after binary operator
goodpeer/goodpeer.py:82:49: W504 line break after binary operator
goodpeer/goodpeer.py:84:47: W504 line break after binary operator
```
@cdecker I think we can allow W504 (maybe also in main repo) as I think it increases readability when used like in the above occurences.

darosior:
Thanks for giving it a look :-)

I have mixed feelings about this plugin:
- I think it should certainly not be in autopilot, now that we have CLBOSS i think we should focus here on smaller parts of what consists a comprehensive autopilot so people (and us actually) can choose to run only smaller parts.
- I think it's great and got me great finds. However, the best peers to connect to are.. A way more complicated matter. If we merge this i should probably rename it to "blackfridaypeers" :p
- I'm still thinking about the explorer i mentioned previously. It will probably be called "edges" and will allow research of peers sorted by new edges created, which could be weighted. It will be a completely new plugin however and should not condition the fortune of this one :)

m-schmoock:
@darosior   "blackfridaypeers"  :D

One thing that came to my mind is that it would be good to limit the output to just the best `20` peers or something, maybe a parameter that can be set to `100` or  `all` if one wants more output.

PR #90, opened 2020-02-12
request-invoice 

Alex-CodeLab:
This plugin starts a minimal, rate limited HTTP Server and returns a invoice.

### DIFF
--- a/backup/README.md
+++ b/backup/README.md
@@ -27,8 +27,12 @@ internal state, and which makes sure no instances are using the same backup.
 ```
 
 Notes:
- - Make sure to adjust the `~/.lightning/bitcoin` directory to your needs.
- - You should use some non-local SSH or NFS mount as destination.
+ - If you are not using the default lightning directory you'll need to
+   change `~/.lightning/bitcoin` in the command line to point to that
+   directory instead.
+ - You should use some non-local SSH or NFS mount as destination,
+   otherwise any failure of the disk may result in both the original
+   as well as the backup being corrupted.
  - Currently only the `file:///` URL scheme is supported.
 
 ## IMPORTANT note about hsm_secret


darosior:
Unrelated Travis error, I think #91 fixes that...

Could you add a basic test just to make sure we don't break your plugin's commands upstream ?

darosior:
You need to make your plugin executable... Did you test locally ?
```
pyln.client.lightning.RpcError: RPC call failed: method: plugin, payload: {'subcommand': 'start', 'plugin': '/home/travis/build/lightningd/plugins/request-invoice/requestinvoice.py'}, error: {'code': -32602, 'message': '/home/travis/build/lightningd/plugins/request-invoice/requestinvoice.py is not executable: Permission denied'}
```

request-invoice/requestinvoice.py
```diff
@@ -0,0 +1,143 @@
...
+jobs = {}
+
+
+def start_server(port):
+    if port in jobs:
+        return False, "server already running"
+
+    p = multiprocessing.Process(
+        target=worker, args=[port], name="server on port {}".format(port))
+    p.daemon = True
```
cdecker: Running in a separate process might make future things harder since we don't really have a way to communicate with the subprocess at all. I'd suggest using `threading.Thread()` instead.

request-invoice/requestinvoice.py
```diff
@@ -0,0 +1,143 @@
...
+    /invoice/<amount>/<description>  
+    where amount is in Satoshis.  
+    The plugin takes one of the following commands: 
+    {start/stop/restart} and {port} , default port: 8089.
+    """
+    commands = {"start", "stop", "restart", "list"}
+
+    # if command unknown make start our default command
+    if command not in commands:
+        command = "start"
```
cdecker: This kind of clobbering the command if unknown may be strange. For example if I call `invoiceserver STOP 1234` it'd start a new one, because the subcommand is not in the list. Better to return an error here.

request-invoice/requestinvoice.py
```diff
@@ -0,0 +1,143 @@
...
+
+    # if port not an integer make 8088 as default
+    try:
+        port = int(port)
+    except:
+        port = int(plugin.options['invoice-web-port']['value'])
```
cdecker: This is also rather unexpected, why not just return an error message telling the user that this is not a valid port number?

request-invoice/requestinvoice.py
```diff
@@ -0,0 +1,143 @@
...
+    IOLoop.instance().start()
+    return
+
+
+jobs = {}
+
+
+def start_server(port):
+    if port in jobs:
+        return False, "server already running"
```
cdecker: This check is duplicated in the `invoiceserver` RPC method.

request-invoice/requestinvoice.py
```diff
@@ -0,0 +1,143 @@
...
+            return "Server already running on port {}. Maybe restart the server?".format(port)
+        suc = start_server(port)
+        if suc:
+            return "started server successfully on port {}".format(port)
+        else:
+            return "Could not start server on port {}".format(port)
+
+    if command == "stop":
+        if stop_server(port):
+            return "stopped server on port{}".format(port)
```
cdecker:
```suggestion
            return "stopped server on port {}".format(port)
```

request-invoice/test_request-invoice.py
```diff
@@ -0,0 +1,44 @@
...
+    assert(r.status_code == 429)
+    assert("429 Too Many Requests" in r.text )
+
+    l1.rpc.plugin_stop(plugin_path)
+    l1.stop()
+
+    # terminate background process
+    try:
+        full_path = "python3 " + plugin_path
+        check_output(["pkill","-f",  full_path])
```
cdecker: Yep, that's going to be a problem. Hence my suggestion of using `threading.Thread()` instead of `multiprocessing.Process()`.

Alex-CodeLab:
care to explain?

cdecker:
My bad, I misinterpreted this as being a stale PR, and closed it prematurely. I'll review and merge if appropriate.

darosior:
Hi @Alex-CodeLab , what's the state of this ?

Alex-CodeLab:
@hi Daroior
I think it is ready since last summer.
(it is active on my node for months without issues.)

PR #95, opened 2020-03-30
Adapt Sauron to the new lightningd

darosior:
After https://github.com/ElementsProject/lightning/pull/3570 was merged.

I used a higher feerate for HTLC resolution and penalty transaction, as per https://github.com/ElementsProject/lightning/pull/3592. What do you think ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ Community curated plugins for c-lightning.
 | [prometheus][prometheus]           | Lightning node exporter for the prometheus timeseries server               |
 | [rebalance][rebalance]             | Keeps your channels balanced                                               |
 | [reckless][reckless]               | An **experimental** plugin manager (search/install plugins)                |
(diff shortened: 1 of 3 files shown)


darosior:
Hi @codecov ! Unfortunately I haven't figured out yet how to make Travis automatically test that !

Otherwise I just added a small README, too.

cdecker:
> Hi @codecov ! Unfortunately I haven't figured out yet how to make Travis automatically test that !
> 
> Otherwise I just added a small README, too.

Might be sufficient to just run against testnet, that's what I tried with my bitcoin-cli replacements. It does allow us to test some parts, except `sendrawtransaction` maybe.

darosior:
> Might be sufficient to just run against testnet, that's what I tried with my bitcoin-cli replacements. It does allow us to test some parts, except sendrawtransaction maybe.

Used mainnet because of blockchain.info line :/

PR #104, opened 2020-04-19
Prometheus: avoid hyphens in label names

sgeisler:
Prometheus 2.7.1+ds apparently doesn't accept hyphens in label names. Replacing them with underscores fixes the error I got otherwise:
```
error while linting: text format parsing error in line 3: expected '=' after label name, found '-'
```

### DIFF
--- a/prometheus/prometheus.py
+++ b/prometheus/prometheus.py
@@ -16,7 +16,7 @@ class BaseLnCollector(object):
 class NodeCollector(BaseLnCollector):
     def collect(self):
         info = self.rpc.getinfo()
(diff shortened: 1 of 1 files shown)


sgeisler:
The travis error seems to be a random failure in `backup/test_backup.py`. Since codecov is bugging me: should I add a test for my change? Seems a bit excessive tbh.

cdecker:
ACK https://github.com/lightningd/plugins/commit/4612a1266d9b5a6fe86e3db45f285521a3bbbc03

PR #106, opened 2020-05-05
helpme: fix out of range access in configfile parser

### LINKED ISSUE #59: helpme: returns `list assignment index out of range`
When I try to run `helpme:` "Error while processing helpme: IndexError('list assignment index out of range')"

theStack:
Within the if-block the list `parts` obviously only consists of one element, so the access `[1]` fails. 

Probably fixes #59


cdecker:
ACK https://github.com/lightningd/plugins/pull/106/commits/30a51d08d88cd9c33b03a9af5d55cb8e09618e85

PR #110, opened 2020-05-17
Tor support for Sauron

darosior:


sauron/README.md
```diff
@@ -23,3 +23,14 @@ Here is a fully reptilian example running against [blockstream.info](https://blo
...
+You can use also proxy your requests through [Tor](https://www.torproject.org/) by adding
+the `--sauron-use-tor` startup option.
+
+Hidden services are also supported :
+```
+lightningd --testnet --disable-plugin bcli --plugin $PWD/sauron.py --sauron-api-endpoint http://explorerzydxu5ecjrkwceayqybizmpjjznk5izmitf2modhcusuqlid.onion/testnet/api/ --sauron-use-tor
+```
+
+*Just to make it clear, this option doesn't make you automagically anonymous : if you are
+using a public endpoint then you are giving away all links between your coins.*
```
darosior, line 36: Should I drop this ? As I often see a lot of confusion about the effects of using Tor I thought it would be nice to have..

sauron/sauron.py
```diff
@@ -6,7 +6,8 @@
 from pyln.client import Plugin
 
 
-plugin = Plugin()
+plugin = Plugin(dynamic=False)
+SESSION = requests.session()
```
cdecker, line 10: Notice that adding a session here means that any cookie set by the server is sent back on subsequent requests, which might allow a server to collate the requests easily.

darosior: Yeah, but if I go this way I should also renew circuits after each call to `gettxout`, send tx, get tx as the exit node ip might allow to collate the requests too ?

darosior: This is not viable so I thought about some heuristics (keeping track of txids, timeouts), in the same vein maybe drop `requests` for a more low level solution (pycurl?) to avoid automagically leaking data.
At the end I considered this as a "avoid opening a public port and do port forwarding for my home Esplora instance" more than a solution to hit public APIs and added the above comment to the README..

I can still renew session though.
Do you think we should consider the run-against-public-APIs case ?

cdecker: I think it's ok, but we can generate a new session on each distinct action. But afaik Tor circuits are persistent for a time / destination, so not much to be gained here.

sauron/README.md
cdecker: The only really public information should be transactions we send out anyway, shouldn't it? Otherwise we process full blocks, grabbing owned outputs without revealing ownership. Behavior can leak that we're a LN node, due to short channel ID lookups, but that's not a huge leak imho.

darosior: There are also the `getutxout` calls we do with our coins iirc ?

cdecker: Only on `dev-rescan-outputs` iirc, which users usually don't do.

sauron/sauron.py
```diff
@@ -161,4 +173,23 @@ def getfeerate(plugin, **kwargs):
...
+    "localhost",
+    "Tor's SocksPort address. Requires --sauron-use-tor."
+)
+
+plugin.add_option(
+    "sauron-tor-port",
+    9050,
+    "Tor's SocksPort port. Requires --sauron-use-tor."
+)
+
```
cdecker, line 194: These could be collapsed to a single option `--sauron-tor-proxy`? If it is not set we don't use TOR, if it is it's an `address[:port]` string that tells us where the SOCKS5 port is.

darosior:
I removed `sauron-use-tor` but it now means we cannot use the handy `default` field and set it to `localhost:9050` :/

I also reworked it a bit.

cdecker:
> I removed `sauron-use-tor` but it now means we cannot use the handy `default` field and set it to `localhost:9050` :/
> 
> I also reworked it a bit.

We can always add an option later to disable sauron's Tor use if that's desired. We should probably detect if Tor is available and just use it if it is :-) Users have a choice to be less private by setting `sauron-tor-proxy` to a dummy value (e.g., `false` or `no` will do that).

cdecker:
ACK https://github.com/lightningd/plugins/pull/110/commits/61541285fa6769575702c22e18616ba56d02e467

PR #123, opened 2020-05-31
prometheus: give short channel id for scid fiels, not long one

darosior:
For example : `lightning_channel_htlcs{id="02065e25c272203440b66ea0ba66601d1248564554d0a68472b82511af54288120",scid="f93c2d2af0f1abf88b77bd2f014e06fbf064c18199ddb20e197195b2dc492b6c"} 1.0`

### DIFF
--- a/prometheus/prometheus.py
+++ b/prometheus/prometheus.py
@@ -162,7 +162,7 @@ class ChannelsCollector(BaseLnCollector):
         peers = self.rpc.listpeers()['peers']
         for p in peers:
             for c in p['channels']:
-                labels = [p['id'], c['channel_id']]
+                labels = [p['id'], c.get('short_channel_id', c.get('channel_id'))]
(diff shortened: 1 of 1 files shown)


prometheus/prometheus.py
```diff
@@ -162,7 +162,7 @@ def collect(self):
         peers = self.rpc.listpeers()['peers']
         for p in peers:
             for c in p['channels']:
-                labels = [p['id'], c['channel_id']]
+                labels = [p['id'], c.get('short_channel_id', c.get('channel_id'))]
```
cdecker, line 165: Doesn't this now mismatch label keys with label values? We probably need to either add the long ID to the gauge and counter declarations, or drop it here (I think `short_channel_id` is sufficient and more memorable?).

darosior: The label was `scid`, so it mismatched before (I just kept the long id in case the channel isn't yet confirmed and `short_channel_id` isn't a field of `listpeers` yet)

PR #124, opened 2020-06-06
G.node -> G.nodes

svewa:
Without this fix I get:  AttributeError: 'Graph' object has no attribute 'node' (freshly installed networkx)

https://networkx.github.io/documentation/stable/release/release_2.4.html#deprecations claims it's not G.node anymore, but G.nodes is to be used now. Worked for me.


cdecker:
ACK https://github.com/lightningd/plugins/pull/124/commits/fab99a8dc9466fdb1edf753eeedfd059e13457f1

PR #125, opened 2020-06-07
Limit channel size, don't require confirmations for channel funding

svewa:
This fixes two problems my autopilot kept running into:
1) trying to fund too large channels, which failed for this reason

2) having only one UTXO, which backfired when it tried to fund more than one channel via autopilot at once, as it would only accept UTXOs with several confirmations to fund a channel. So only the first channel got funded and then I got errors for all others as the new UTXOs were still unconfirmed.


cdecker:
ACK https://github.com/lightningd/plugins/pull/125/commits/607c5d01935727e3f3b33f91111f7cb16df57581

PR #127, opened 2020-07-20
feat: summary peer availability tracing

m-schmoock:
Adds:
 - peer availability tracing, to show the node operator which peers are reliable.
 - 72hr exponential smoothing window

Fixes:
 - fix broken fiat price API by switching to public (and reliable) bitstamp ticker API.
 - fix PriceThread actually perform a loop instead of fetching price only one time

### DIFF
--- a/summary/summary.py
+++ b/summary/summary.py
@@ -20,6 +20,7 @@ try:
 except Exception:
     pass
 
+Channel = namedtuple('Channel', ['total', 'ours', 'theirs', 'pid', 'private', 'connected', 'scid', 'avail'])
 Charset = namedtuple('Charset', ['double_left', 'left', 'bar', 'mid', 'right', 'double_right', 'empty'])
 if have_utf8:
     draw = Charset('╟', '├', '─', '┼', '┤', '╢', '║')
@@ -133,10 +134,10 @@ def summary(plugin, exclude=''):
                 to_them = Millisatoshi(0)
             avail_in += to_them
             reply['num_channels'] += 1
-            chans.append((
+            chans.append(Channel(
                 c['total_msat'],
                 to_us, to_them,
-                p['id'],
+                pid,
                 c['private'],
                 p['connected'],
                 c['short_channel_id'],
@@ -157,12 +158,12 @@ def summary(plugin, exclude=''):
     if chans != []:
         reply['channels_flags'] = 'P:private O:offline'
         reply['channels'] = ["\n"]
-        biggest = max(max(int(c[1]), int(c[2])) for c in chans)
+        biggest = max(max(int(c.ours), int(c.theirs)) for c in chans)
         append_header(reply['channels'], biggest)
         for c in chans:
             # Create simple line graph, 47 chars wide.
-            our_len = int(round(int(c[1]) / biggest * 23))
-            their_len = int(round(int(c[2]) / biggest * 23))
+            our_len = int(round(int(c.ours) / biggest * 23))
+            their_len = int(round(int(c.theirs) / biggest * 23))
             divided = False
 
             # We put midpoint in the middle.
@@ -186,28 +187,28 @@ def summary(plugin, exclude=''):
             s = left + mid + right
 
             # output short channel id, so things can be copyNpasted easily
-            s += " {:14} ".format(c[6])
+            s += " {:14} ".format(c.scid)
 
             extra = ''
-            if c[4]:
+            if c.private:
                 extra += 'P'
             else:
                 extra += '_'
-            if not c[5]:
+            if not c.connected:
                 extra += 'O'
             else:
                 extra += '_'
             s += '[{}] '.format(extra)
 
             # append 24hr availability
-            s += '{:4.0%}  '.format(c[7])
+            s += '{:4.0%}  '.format(c.avail)
 
             # append alias or id
-            node = plugin.rpc.listnodes(c[3])['nodes']
+            node = plugin.rpc.listnodes(c.pid)['nodes']
             if len(node) != 0 and 'alias' in node[0]:
                 s += node[0]['alias']
             else:
-                s += c[3][0:32]
+                s += c.pid[0:32]
             reply['channels'].append(s)
 
     # Make modern lightning-cli format this human-readble by default!
--- a/summary/summary.py
+++ b/summary/summary.py
@@ -79,7 +79,7 @@ class PeerThread(threading.Thread):
                         peerstate[pid]['connected'] = False
                         peerstate[pid]['availability'] = 0.0 * alpha + peerstate[pid]['availability'] * beta
             except Exception as ex:
-                plugin.log("[PeerThread] " + str(ex), 'error')
+                plugin.log("[PeerThread] " + str(ex), 'warn')
             time.sleep(interval)
 
 
@@ -94,7 +94,7 @@ class PriceThread(threading.Thread):
                 r = requests.get('https://www.bitstamp.net/api/v2/ticker/BTC{}'.format(plugin.currency))
                 plugin.fiat_per_btc = float(r.json()['last'])
             except Exception as ex:
-                plugin.log("[PriceThread] " + str(ex), 'error')
+                plugin.log("[PriceThread] " + str(ex), 'warn')
             # Six hours is more than often enough for polling
             time.sleep(6*3600)
 
--- a/summary/summary.py
+++ b/summary/summary.py
@@ -2,7 +2,7 @@
 from pyln.client import Plugin, Millisatoshi
 from packaging import version
 from collections import namedtuple
-from datetime import datetime
+from summary_avail import *
 import pyln.client
 from math import floor, log10
 import requests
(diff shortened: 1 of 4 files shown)


summary/summary.py
```diff
@@ -30,21 +31,77 @@
...
+                    pid = p['id']
+                    addpeer(p)
+
+                    if p['connected']:
+                        peerstate[pid]['last_seen'] = datetime.now()
+                        peerstate[pid]['connected'] = True
+                        peerstate[pid]['availability'] = peerstate[pid]['availability'] * alpha + 1.0 * beta
+                    else:
+                        peerstate[pid]['connected'] = False
+                        peerstate[pid]['availability'] = peerstate[pid]['availability'] * alpha + 0.0 * beta
```
niftynei, line 82: i've never seen this alpha/beta trick before, does it have a name?

summary/summary.py
```diff
@@ -30,21 +31,77 @@
...
+        interval = 5 * 60            # collect peer state once in a while
+        window   = 3 * 24 * 60 * 60  # 72hr hr availability window
+        count    = 0
+
+        # delay initial execution, so peers have a chance to connect on startup
+        time.sleep(interval)
+
+        while True:
+            count += 1
+            leadwin  = max(min(window, count * interval), count * interval)
```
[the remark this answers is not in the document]
m-schmoock: I guess your right, didn't catch that.
So for count being a bigger number, the `leadwin` should be set to `window` (which was my intention). Will fix

summary/summary.py
```diff
@@ -30,21 +31,77 @@
...
+            samples  = leadwin / interval
+            alpha    = 1.0 - 1.0 / samples
+            beta     = 1.0 - alpha
```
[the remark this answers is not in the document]
m-schmoock: sure, its just a more excessive way of expressing that there's a exponential smoothing going on in the following code.

summary/summary.py
m-schmoock: exponential smoothing

m-schmoock: and yes, `0.0 * beta` is pointless, but it would also work for values in between, which do not happen for bolean states :D

niftynei: so if i understand this simple exponential smoothing algo from [wikipedia](https://en.wikipedia.org/wiki/Exponential_smoothing)

exponential smoothing

What this would translate to in this case is 

```
alpha * (if p['connected'] 1.0 else 0.0) + beta * peerstate[pid]['availability']
```

- `alpha` is multiplied by the input at this window, in this case 'is_connected'
- `beta` is multiplied by the previous output, or `peerstate[pid]['availability']`

niftynei: looking at this more closely, your `alpha` and `beta` calculations are flipped with respect to the wikipedia example -- as is you've got it coded correctly.

summary/summary.py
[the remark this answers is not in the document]
m-schmoock: indeed , last `count * interval` should have been just `interval`

summary/summary.py
m-schmoock: I flipped alpha / beta to make it more regular :D

m-schmoock:
I think I can also rearrange code to make the calculation tetstable so codecov is more happy with me

m-schmoock:
Okay, I fixed the code coverage, even though it was just related to seemingly not profiling unittest thread itself.
By adding a test with a low `summary-availability-interval` value we get the PeerThread to execute quick enough, and this thread is audited by codecov.

summary/summary.py
```diff
@@ -326,4 +298,14 @@ def init(options, configuration, plugin):
     'USD $',
     'What prefix to use for currency'
 )
+plugin.add_option(
+    'summary-availability-interval',
+    300,
+    'How often in seconds the availability should be calculated. Causes RPC and CPU.'
```
niftynei, line 270: The meaning of `Causes RPC and CPU` isn't totally clear to me. May be clearer as "uses" or "causes usage"?

summary/summary.py
```diff
@@ -326,4 +298,14 @@ def init(options, configuration, plugin):
...
+)
+plugin.add_option(
+    'summary-availability-window',
+    72,
+    'How many hours the availability should be averaged.'
```
niftynei, line 275:
```suggestion
    'How many hours the availability should be averaged over.'
```

summary/test_summary.py
```diff
@@ -20,6 +22,26 @@ def get_stub():
     return plugin
 
 
+def test_summary_peer_thread(node_factory):
+    # in oder to give the PeerThread a chance in a unit test
```
niftynei, line 26:
```suggestion
    # in order to give the PeerThread a chance in a unit test
```

summary/summary.py
```diff
@@ -73,43 +73,9 @@ def to_fiatstr(msat: Millisatoshi):
                               int(msat) / 10**11 * plugin.fiat_per_btc)
 
 
-# This is part of pylightning, but its just merged,
```
niftynei, line 55: how long has the updated pylightning release been available for?

m-schmoock: It already was in the pyln-client version the requirements.txt suggested for this plugin anyway, but since we do not track plugin deps precisely, I put that in in the past just to be sure. The pyln-client `plugin.fiat_per_btc` was added to master with commit 0efd72d88 what is a bit more than one year ago.

summary/summary.py
m-schmoock: If your okay, I simply removed that silly addition. I thought it would be useful to tell the user that this is just local RPC and doesn't consume much resources or network stuff. But it doesn't add anything particular interesting.

Solved, rebased an force-pushed.

summary/summary.py
m-schmoock: Solved, rebased an force-pushed.

summary/test_summary.py
m-schmoock: Solved, rebased an force-pushed.

summary/summary.py
```diff
@@ -2,7 +2,7 @@
 from pyln.client import Plugin, Millisatoshi
 from packaging import version
 from collections import namedtuple
-from datetime import datetime
+from summary_avail import *
```
cdecker, line 5: If you're concerned that the path might not be unique you can explicitly import a local module by using the following:

```python
from .availability import *
```

This saves you from having to prefix everything. Alternatively you could also have the executable in the top-level and create a `summary` package containing all the sub-modules.

cdecker:
Regarding codecov, I'm really tempted to just remove it. It hasn't produced anything actionable, and has been outright misleading in some cases (some plugins flip-flopping despite not being touched...)

It might just be our setup, with tests running as unit-tests and also collecting coverage when run as plugins.

summary/summary.py
```diff
@@ -37,14 +37,14 @@ def __init__(self):
...
-                             '?from=BTC&to={}&amount=1'.format(plugin.currency))
-            plugin.fiat_per_btc = json.loads(r.content.decode('UTF-8'))['price']
-        except Exception:
-            pass
-        # Six hours is more than often enough for polling
-        time.sleep(6*3600)
+        while True:
+            try:
+                r = requests.get('https://www.bitstamp.net/api/v2/ticker/BTC{}'.format(plugin.currency))
+                plugin.fiat_per_btc = float(json.loads(r.content.decode('UTF-8'))['last'])
```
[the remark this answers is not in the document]
m-schmoock: thx. done

summary/summary.py
```diff
@@ -30,6 +31,60 @@
...
+            'last_seen' : datetime.now() if p['connected'] else None,
+            'availability' : 1.0 if p['connected'] else 0.0
+        }
+
+
+class PeerThread(threading.Thread):
+    def __init__(self):
+        super().__init__()
+        self.daemon = True
+        self.start()
```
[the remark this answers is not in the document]
m-schmoock: thx. done

summary/summary.py
m-schmoock: Wont do, since plugin API or autoreload plugin gets me this:

```
    from .availability import *
ImportError: attempted relative import with no known parent package
```

summary/summary.py
```diff
@@ -147,7 +204,15 @@ def summary(plugin, exclude=''):
...
-            chans.append((c['total_msat'], to_us, to_them, p['id'], c['private'], p['connected'], c['short_channel_id']))
+            chans.append((
+                c['total_msat'],
+                to_us, to_them,
+                p['id'],
+                c['private'],
+                p['connected'],
+                c['short_channel_id'],
+                peerstate[pid]['availability']
+            ))
```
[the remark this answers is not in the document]
m-schmoock: thx. done

summary/summary.py
```diff
@@ -30,6 +31,60 @@
...
+        self.daemon = True
+        self.start()
+
+    def run(self):
+        interval = 5 * 60            # collect peer state once in a while
+        window   = 3 * 24 * 60 * 60  # 72hr availability window
+        count    = 0
+
+        # delay initial execution, so peers have a chance to connect on startup
+        time.sleep(interval)
```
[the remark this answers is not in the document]
m-schmoock: Yes, wont but do now :D

PR #131, opened 2020-08-12
add trustedcoin, sparko and webhook to the main list.

fiatjaf:
Because I wasn't working well with a big plugin monorepo.


darosior (commented):
ACK on sparko and webhook, but trustedcoin is redundant with Sauron ?

fiatjaf:
It is similar, but has a different approach, a different deployment method, different sources. Why not allow both?

darosior:
No strong opinion :)

PR #137, opened 2020-08-28
noise: Fixup documentation about waiting for next message

### LINKED ISSUE #136: Noise plugin -1 as last_id not working
As per the readme, to get the last received message use:
`recvmsg -1` however it fails with:
`cli/lightning-cli: -1: unrecognized option`

cdecker:
As pointed out by @Fonta1n3 in #136, the docs were incorrect when it comes to
waiting for the next message. The docs have been updated accordingly.

Fixes #136

### DIFF
--- a/noise/README.org
+++ b/noise/README.org
@@ -25,13 +25,15 @@ lightning-cli recvmsg last_id
 #+END_SRC
 
 The ~last_id~ indicates the last message we read, so we can retrieve each message
-individually. If you'd just like to wait for the next message you can use a
-~last_id~ or ~-1~.
+individually. If you'd just like to wait for the next message then do not
+specify any ~last_id~.
+
 * Todo
 
 - [ ] Persist messages across restarts
-- [ ] Use ~rpc_command~ to intercept any payment listing and add the keysend
-  payments to it.
+- [X] Use ~rpc_command~ to intercept any payment listing and add the keysend
+  payments to it. (No longer needed, since keysends are handled correctly by
+  ~listpays~.)
 
 * Protocol
 The protocol was heavily inspired by the [[https://github.com/joostjager/whatsat#protocol][WhatSat protocol]]:
--- a/noise/noise.py
+++ b/noise/noise.py
@@ -156,7 +156,9 @@ def sendmsg(node_id, msg, plugin, request, pay=None, **kwargs):
 def recvmsg(plugin, request, last_id=None, **kwargs):
     next_id = int(last_id) + 1 if last_id is not None else len(plugin.messages)
     if next_id < len(plugin.messages):
-        request.set_result(plugin.messages[int(last_id)].to_dict())
+        res = plugin.messages[int(last_id)].to_dict()
+        res['total_messages'] = len(plugin.messages)
+        request.set_result(res)
     else:
         plugin.receive_waiters.append(request)
 
@@ -203,8 +205,13 @@ def on_htlc_accepted(onion, htlc, plugin, **kwargs):
         res = {'result': 'continue'}
 
     plugin.messages.append(msg)
+    print("Delivering message to {c} waiters".format(
+        c=len(plugin.receive_waiters)
+    ))
     for r in plugin.receive_waiters:
-        r.set_result(msg.to_dict())
+        m = msg.to_dict()
+        m['total_messages'] = len(plugin.messages)
+        r.set_result(m)
     plugin.receive_waiters = []
 
     return res
@@ -215,5 +222,6 @@ def init(configuration, options, plugin, **kwargs):
     print("Starting noise chat plugin")
     plugin.messages = []
     plugin.receive_waiters = []
+
 
 plugin.run()


noise/README.org
```diff
@@ -25,13 +25,15 @@ lightning-cli recvmsg last_id
...
+specify any ~last_id~.
+
 * Todo
 
 - [ ] Persist messages across restarts
-- [ ] Use ~rpc_command~ to intercept any payment listing and add the keysend
-  payments to it.
+- [X] Use ~rpc_command~ to intercept any payment listing and add the keysend
+  payments to it. (No longer needed, since keysends are handled correctly by
+  ~listpays~.)
```
darosior, line 36: :+1:

darosior (approved):
ut ACK

Fonta1n3 (commented):
Just testing this, looks like line 208 causes a crash, removing the `f` fixes it. 
`print(f"Delivering message to {len(plugin.receive_waiters)} waiters")`

Also I can confirm the following strange behavior:

Seems when ever the receiver specifies the most recent `id` index the call just hangs forever as an example:

```
lightning-cli recvmsg

{
   "payment": null,
   "verified": true,
   "signature": "20682ff309791b510281391482922fa572733aefe210c4423dcb39de877e92f59013b9319629d37574fcd229c8d038e57ff824953d6fefc28db01b1468cd1124d6",
   "id": 1,
   "body": "oi oi this is wa",
   "sender": "0339a47348194229cd66d9b805be7a1c64afac5418b037ba2c50c148e3141ab39c",
   "total_messages": 2
}

lightning-cli recvmsg 0
# the following response returns immediately, just like you would expect
{
   "payment": null,
   "verified": true,
   "signature": "1fa369a02388b5693867233d0d0d852e577c64c10c41649755ac775a002722715f5dbee6a14016646da1c9befe2773961864fbf15a633ee34fa872b52dc421d131",
   "id": 0,
   "body": "oi oi this is way better",
   "sender": "0339a47348194229cd66d9b805be7a1c64afac5418b037ba2c50c148e3141ab39c",
   "total_messages": 2
}

# however this hangs forever and never returns a response
lightning-cli recvmsg 1
```

darosior:
You need python3.6 for fstring. I hesitated while reviewing since py3.5 is still the default for Debian oldstable, but note that 3.5 is EOL in something like a month.

So i think it makes sense to have people upgrade to py3.6 to use this plugin. Do you see a big inconvenience to this for your project @Fontaine ?

-------- Original Message --------

Fonta1n3:
No that’s fine, 3.6 it is. That explains why one node did not crash. Thanks for the explanation.

cdecker:
Oops, that f-string should not be there (I added it just for testing). I'll remove it before merging.

As for hanging when specifying `id=total_messages`: that is intended behavior. Notice that one is a count, whereas the other one is an index into an array. In most programming languages the indices are 0-based, meaning that an array with 10 elements has a maximum index of 9. So asking for index 10 is one past the last element, which causes the plugin to assume you just want to wait for the next message (which can be interpreted as hanging).

I think this is what should be a very natural API: ask for the next message and the plugin will return immediately if it has one, or block and wait for the next if it hasn't yet. No need to disambiguate in the caller. We could add a timeout to the call if that's desired though.

Fonta1n3:
But the count is 2. I sent two messages, to get the first I ask for id 0 which works, to get the last message i ask for 1 yet it hangs.

cdecker:
> But the count is 2. I sent two messages, to get the first I ask for id 0 which works, to get the last message i ask for 1 yet it hangs.

My bad for misremembering the API I created myself: the parameter is supposed to be the `last_id` that you processed, i.e., the predecessor to the message you are asking for, not the `id` of the message that you are trying to retrieve. This is because, while we currently guarantee that the `message_id`s are sequential, they might not be in future (if we identify messages by their hash or timestamp or a UUID). That was also the reason to use `-1` to retrieve the very first message.

If you prefer I can bump the message IDs by `+1` when recording them, so `recvmsg 0` gets the first ever message with `id=1` and so forth (to sidestep the `-1` issue on the CLI), but I don't see how we can make a robust tailing API with non-consecutive IDs, without providing the high-watermark from the last message to get the next.

Fonta1n3:
> > But the count is 2. I sent two messages, to get the first I ask for id 0 which works, to get the last message i ask for 1 yet it hangs.
> 
> My bad for misremembering the API I created myself: the parameter is supposed to be the `last_id` that you processed, i.e., the predecessor to the message you are asking for, not the `id` of the message that you are trying to retrieve. This is because, while we currently guarantee that the `message_id`s are sequential, they might not be in future (if we identify messages by their hash or timestamp or a UUID). That was also the reason to use `-1` to retrieve the very first message.
> 
> If you prefer I can bump the message IDs by `+1` when recording them, so `recvmsg 0` gets the first ever message with `id=1` and so forth (to sidestep the `-1` issue on the CLI), but I don't see how we can make a robust tailing API with non-consecutive IDs, without providing the high-watermark from the last message to get the next.

Ok I think I may be getting confused now.. I thought I found the issue and submitted #145 which fixes it for me :) To be clear calling `rcvmsg 0` returns the zeroth message as is, it is only the max index I am having an issue with.
TBH I think by far the best solution here would be to add a unix time stamp to each message, then allow the user to retrieve the last `x` messages sorted by timestamp. That approach would seem to solve all issues and would allow adding a `uuid`. I would be happy to do that and submit a PR but my experience with python is lacking...

cdecker:
> Ok I think I may be getting confused now.. I thought I found the issue and submitted #145 which fixes it for me :) To be clear calling `rcvmsg 0` returns the zeroth message as is, it is only the max index I am having an issue with.
> 
> TBH I think by far the best solution here would be to add a unix time stamp to each message, then allow the user to retrieve the last `x` messages sorted by timestamp. That approach would seem to solve all issues and would allow adding a `uuid`. I would be happy to do that and submit a PR but my experience with python is lacking...

Good point, I think we can move away from the `recvmsg` and go to a `getmsgs` that retrieves a page of messages that have a higher timestamp than the one specified (meaning "I received up to this time, give me all later ones"). Does that match your expectations? Adding a new RPC method doesn't risk breaking any existing users, unlike changing the `recvmsg` semantics.

Fonta1n3:
@cdecker That would be fantastic, really appreciate it!

PR #138, opened 2020-08-28
prometheus: Change default port to 9750 as per port-allocation

### LINKED ISSUE #122: Advertise prometheus plugin
Please consider adding the prometheus plugin to the [list of existing exporters](https://github.com/prometheus/docs/blob/master/content/docs/instrumenting/exporters.md) so that people with weak googling skills won't try to [write their own](https://github.com/mmilata/prometheus-clightning-exporter):laughing:. Prometheus also maintains [list of known port allocations](https://github.com/prometheus/prometheus/wiki/Default-port-allocations).

cdecker:
We registered port 9750 the [port allocation] list for the prometheus project,
so we should use that port :-)

The addition to the list of known exporters is still pending https://github.com/prometheus/docs/pull/1737

Suggested-by: Martin Milata <@mmilata>
Fixes #122
[port allocation]: https://github.com/prometheus/prometheus/wiki/Default-port-allocations

### DIFF
--- a/prometheus/README.md
+++ b/prometheus/README.md
@@ -5,7 +5,7 @@ so it can be scraped, plotted and alerts can be created on it. The plugin adds
 the following command line arguments:
 
  - `prometheus-listen`: the IP address and port to bind the HTTP server to
-   (default: `0.0.0.0:9900`)
+   (default: `0.0.0.0:9750`)
    
 Exposed variables include:
 
--- a/prometheus/prometheus.py
+++ b/prometheus/prometheus.py
@@ -213,7 +213,7 @@ def init(options, configuration, plugin):
 
 plugin.add_option(
     'prometheus-listen',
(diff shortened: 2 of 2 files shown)


darosior:
Nice, brian-brazil left some good recommendations on your PR:
> Some suggestions to improve your exporter:
> _total is a suffix for counters, it shouldn't be used with gauges. Some metrics like lightning_channel_in_msatoshi_offered also sound like they might be counters rather than gauges.
> Units are inconsistent, sometimes missing and conflicting with the help text. You should consistently use satoshis, and also include satoshi as the unit in all relevant metric names for clarity.
> id is a bit of a generic label, peer or peer_id would be clearer.

I'll open an issue as they would be a great fit for a follow-up

darosior (approved):
utACK

PR #141, opened 2020-08-28
helpme: Verify that the caller has specified an invoice

### LINKED ISSUE #107: helpme: 'helpme channels' leads to error
I managed to complete stages 1 (funds) and 2 (peers), but calling stage 3 (channels) fails:
```
$ ./cli/lightning-cli helpme channels
"Error while processing helpme: UnboundLocalError(\"local variable 'inv' referenced before assignment\",)"
```
Apparently the function `give_channel_advice(plugin, *args)` accesses the dictionary `inv` near the end:
https://github.com/lightningd/plugins/blob/41e440522e7fd67b106fe79636990477dd8cd5c0/helpme/helpme.py#L767-L772
but this variable was never defined. It would only get initialized if the length of `args` is 1, but it is always called without arguments...
https://github.com/lightningd/plugins/blob/41e440522e7fd67b106fe79636990477dd8cd5c0/helpme/helpme.py#L921-L922

cdecker:
Reported-by: Sebastian Falbesoner <@theStack>
Fixes #107

### DIFF
--- a/helpme/helpme.py
+++ b/helpme/helpme.py
@@ -715,6 +715,8 @@ Note that the default expiry is 7 days, but you can change it: see 'man lightnin
 def give_channel_advice(plugin, *args):
     if len(args) > 1:
         raise ValueError("Sorry, I can only give channel advice for one invoice at a time")
+    elif len(args) < 1:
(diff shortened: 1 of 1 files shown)


darosior (approved):
utACK

PR #142, opened 2020-08-28
sauron: handle testnet API unreliability

darosior:
Cf the discussion on IRC. My guess is that Esplora's estimation depends on the buckets to be filled and that there are not enough transactions on testnet, so better to special-case it.


cdecker:
Rebased on `master` now that Travis should be firing on all cylinders again :-)

PR #145, opened 2020-08-29
Fixes bug for retrieving last message

Fonta1n3:
There was an off by one bug which resulted where `lightning-cli recvmsg <messages.count - 1>` would wait for the next message instead of returning the final message.

### DIFF
--- a/noise/noise.py
+++ b/noise/noise.py
@@ -155,7 +155,7 @@ def sendmsg(node_id, msg, plugin, request, pay=None, **kwargs):
 @plugin.async_method('recvmsg')
 def recvmsg(plugin, request, last_id=None, **kwargs):
     next_id = int(last_id) + 1 if last_id is not None else len(plugin.messages)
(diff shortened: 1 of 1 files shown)


noise/noise.py
```diff
@@ -155,7 +155,7 @@ def sendmsg(node_id, msg, plugin, request, pay=None, **kwargs):
 @plugin.async_method('recvmsg')
 def recvmsg(plugin, request, last_id=None, **kwargs):
     next_id = int(last_id) + 1 if last_id is not None else len(plugin.messages)
-    if next_id < len(plugin.messages):
+    if next_id <= len(plugin.messages):
```
cdecker, line 158: This implicitly relies on the fact that the message IDs are strictly incremented by one, which is ok in this case, but may not be in the future. This breaks the API that you specify the last ID that you processed, and we return the next one.

cdecker (changes requested):

PR #147, opened 2020-08-31
feeadjuster: a plugin to adjust fees depending on channel balancing

darosior:
This is really just a sketch, but i wanted to share the algorithm with you in case you have inputs.

This is structured such as it should be easy to set the `get_ratio` function depending on configuration flags. For example we can imagine a `--feeadjust-algo` option which would default to `basic` and just use the current one, and a `greedy` value would biase the ratio depending on:
- the funds on our side (how much we are exposed)
- if we are the channel funder (even worse, in practice this is my main source of losses)
- maybe the number of HTLCs in flight (this would actually take YA rpc call)

While this is an experiment i expect to run this on a node forwarding a substantial number of payments per day, so care has been taken to minimize the impact (mainly reduce the number of RPC calls).

I need to set up a threshold to avoid flooding, maybe "don't update if it didn't move more than 1% of the funds" so it needs more caching.

This is currently exponential, and i don't think it's the right thing as the adjustment strength will depend on the number of payments and not the proportion of funds.

I gathered some feedback from @renepickhardt and @whitslack by mail too, so if any of you want to weigh in please do :-)

### DIFF
--- a/feeadjuster/feeadjuster.py
+++ b/feeadjuster/feeadjuster.py
@@ -20,34 +20,20 @@ def get_ratio(our_percentage):
     return 50**(0.5 - our_percentage)
 
 
-def get_fees(plugin: Plugin, scid: str):
-    for half in plugin.rpc.listchannels(scid)["channels"]:
-        if plugin.our_node_id in half["source"]:
-            return (half["base_fee_millisatoshi"], half["fee_per_millionth"])
-
-    # Note: the half may not be present, so this may actually return None!
-
-
 def maybe_adjust_fees(plugin: Plugin, scids: list):
     for scid in scids:
         # FIXME: set a threshold to avoid flooding!
         if True:
-            # FIXME: it's exponential! We need to cache the startup
-            # fees instead of always factorising the updated ones.
-            fees = get_fees(plugin, scid)
-            if fees is None:
-                return
-
             our = plugin.adj_balances[scid]["our"]
             total = plugin.adj_balances[scid]["total"]
             ratio = get_ratio(our / total)
             try:
-                plugin.rpc.setchannelfee(scid, int(fees[0] * ratio),
-                                         int(fees[1] * ratio))
+                plugin.rpc.setchannelfee(scid, int(plugin.adj_basefee * ratio),
+                                         int(plugin.adj_ppmfee * ratio))
                 plugin.log("Adjusted fees of {} with a ratio of {}"
                            .format(scid, ratio))
             except RpcError as e:
-                plugin.log(str(e), level="warn")
+                plugin.log("setchannelfee error: " + str(e), level="warn")
 
 
 def get_chan(plugin: Plugin, scid: str):
@@ -85,7 +71,10 @@ def threaded_forward_event(plugin: Plugin, forward_event: dict):
 
     plugin.adj_balances[in_scid]["our"] += forward_event["in_msatoshi"]
     plugin.adj_balances[out_scid]["our"] -= forward_event["out_msatoshi"]
-    maybe_adjust_fees(plugin, [in_scid, out_scid])
+    try:
+        maybe_adjust_fees(plugin, [in_scid, out_scid])
+    except Exception as e:
+        plugin.log("Adjusting fees: " + str(e), level="error")
 
 
 @plugin.subscribe("forward_event")
@@ -101,6 +90,9 @@ def init(options: dict, configuration: dict, plugin: Plugin, **kwargs):
     plugin.adj_thread_pool = ThreadPoolExecutor(
         max_workers=int(options["feeadjuster-par"])
     )
+    config = plugin.rpc.listconfigs()
+    plugin.adj_basefee = config["fee-base"]
+    plugin.adj_ppmfee = config["fee-per-satoshi"]
 
     for peer in plugin.rpc.listpeers()["peers"]:
         if len(peer["channels"]) == 0:
@@ -116,5 +108,6 @@ def init(options: dict, configuration: dict, plugin: Plugin, **kwargs):
 
 
 plugin.add_option("feeadjuster-par", 8, "Maximum number of threads launched to"
-                  "adjust fees", opt_type="int")
+                  " adjust fees", opt_type="int")
+
 plugin.run()
--- a/feeadjuster/feeadjuster.py
+++ b/feeadjuster/feeadjuster.py
@@ -22,16 +22,20 @@ def get_ratio(our_percentage):
 
 def maybe_adjust_fees(plugin: Plugin, scids: list):
     for scid in scids:
-        # FIXME: set a threshold to avoid flooding!
-        if True:
-            our = plugin.adj_balances[scid]["our"]
-            total = plugin.adj_balances[scid]["total"]
-            ratio = get_ratio(our / total)
+        our = plugin.adj_balances[scid]["our"]
+        total = plugin.adj_balances[scid]["total"]
+        percentage = our / total
+        last_percentage = plugin.adj_balances[scid].get("last_percentage")
+
+        # Only update on substantial balance moves to avoid flooding
+        if last_percentage is None or abs(last_percentage - percentage) > 0.05:
+            ratio = get_ratio(percentage)
             try:
                 plugin.rpc.setchannelfee(scid, int(plugin.adj_basefee * ratio),
                                          int(plugin.adj_ppmfee * ratio))
                 plugin.log("Adjusted fees of {} with a ratio of {}"
                            .format(scid, ratio))
+                plugin.adj_balances[scid]["last_percentage"] = percentage
             except RpcError as e:
                 plugin.log("setchannelfee error: " + str(e), level="warn")
 
--- a/feeadjuster/feeadjuster.py
+++ b/feeadjuster/feeadjuster.py
@@ -9,6 +9,8 @@ plugin = Plugin()
 plugin.adj_balances = {}
 # Cache to avoid loads of calls to getinfo
 plugin.our_node_id = None
+# Users can configure this
+plugin.update_treshold = 0.05
 
 
 def get_ratio(our_percentage):
@@ -28,7 +30,7 @@ def maybe_adjust_fees(plugin: Plugin, scids: list):
 
         # Only update on substantial balance moves to avoid flooding, and add
         # some pseudo-randomness to avoid too easy channel balance probing
-        update_treshold = 0.05
+        update_treshold = plugin.update_treshold
         if not plugin.deactivate_fuzz:
             update_treshold += random.uniform(-0.015, 0.015)
 
@@ -98,6 +100,7 @@ def forward_event(plugin: Plugin, forward_event: dict, **kwargs):
 def init(options: dict, configuration: dict, plugin: Plugin, **kwargs):
     plugin.our_node_id = plugin.rpc.getinfo()["id"]
     plugin.deactivate_fuzz = options.get("feeadjuster-deactivate-fuzz", False)
+    plugin.update_treshold = float(options.get("feeadjuster-threshold", "0.05"))
     config = plugin.rpc.listconfigs()
     plugin.adj_basefee = config["fee-base"]
     plugin.adj_ppmfee = config["fee-per-satoshi"]
@@ -111,8 +114,10 @@ def init(options: dict, configuration: dict, plugin: Plugin, **kwargs):
         if chan["state"] != "CHANNELD_NORMAL":
             continue
 
-    plugin.log("Plugin feeadjuster initialized ({} base / {} ppm)"
-               .format(plugin.adj_basefee, plugin.adj_ppmfee))
+    plugin.log("Plugin feeadjuster initialized ({} base / {} ppm) with a "
+               "threshold of {}"
+               .format(plugin.adj_basefee, plugin.adj_ppmfee,
+                       plugin.update_treshold))
 
 
 plugin.add_option(
@@ -120,5 +125,12 @@ plugin.add_option(
     False,
     "Deactivate update threshold randomization and hysterisis.",
     "flag"
+)
+plugin.add_option(
+    "feeadjuster-threshold",
+    "0.05",
+    "Channel balance update threshold at which to trigger an update. Note "
+    "it's fuzzed.",
+    "string"
 )
 plugin.run()
--- a/feeadjuster/feeadjuster.py
+++ b/feeadjuster/feeadjuster.py
@@ -50,11 +50,13 @@ def get_chan(plugin: Plugin, scid: str):
     for peer in plugin.rpc.listpeers()["peers"]:
         if len(peer["channels"]) == 0:
             continue
-        chan = peer["channels"][0]
-        if "short_channel_id" not in chan:
-            continue
-        if chan["short_channel_id"] == scid:
-            return chan
+        # We might have multiple channel entries ! Eg if one was just closed
+        # and reopened.
+        for chan in peer["channels"]:
+            if "short_channel_id" not in chan:
+                continue
+            if chan["short_channel_id"] == scid:
+                return chan
 
 
 def maybe_add_new_balances(plugin: Plugin, scids: list):
--- a/feeadjuster/feeadjuster.py
+++ b/feeadjuster/feeadjuster.py
@@ -6,8 +6,6 @@ from pyln.client import Plugin, Millisatoshi, RpcError
 plugin = Plugin()
 # Our amount and the total amount in each of our channel, indexed by scid
 plugin.adj_balances = {}
-# Make requests to lightningd in parallel
-plugin.adj_thread_pool = None
 # Cache to avoid loads of calls to getinfo
 plugin.our_node_id = None
 
@@ -37,7 +35,8 @@ def maybe_adjust_fees(plugin: Plugin, scids: list):
                            .format(scid, ratio))
                 plugin.adj_balances[scid]["last_percentage"] = percentage
             except RpcError as e:
-                plugin.log("setchannelfee error: " + str(e), level="warn")
+                plugin.log("Could not adjust fees for channel {}: '{}'"
+                           .format(scid, e), level="warn")
 
 
 def get_chan(plugin: Plugin, scid: str):
@@ -68,32 +67,24 @@ def maybe_add_new_balances(plugin: Plugin, scids: list):
             }
 
 
-def threaded_forward_event(plugin: Plugin, forward_event: dict):
-    in_scid = forward_event["in_channel"]
-    out_scid = forward_event["out_channel"]
-    maybe_add_new_balances(plugin, [in_scid, out_scid])
-
-    plugin.adj_balances[in_scid]["our"] += forward_event["in_msatoshi"]
-    plugin.adj_balances[out_scid]["our"] -= forward_event["out_msatoshi"]
-    try:
-        maybe_adjust_fees(plugin, [in_scid, out_scid])
-    except Exception as e:
-        plugin.log("Adjusting fees: " + str(e), level="error")
-
-
 @plugin.subscribe("forward_event")
 def forward_event(plugin: Plugin, forward_event: dict, **kwargs):
     if forward_event["status"] == "settled":
-        plugin.adj_thread_pool.submit(threaded_forward_event,
-                                      plugin, forward_event)
+        in_scid = forward_event["in_channel"]
+        out_scid = forward_event["out_channel"]
+        maybe_add_new_balances(plugin, [in_scid, out_scid])
+
+        plugin.adj_balances[in_scid]["our"] += forward_event["in_msatoshi"]
+        plugin.adj_balances[out_scid]["our"] -= forward_event["out_msatoshi"]
+        try:
+            maybe_adjust_fees(plugin, [in_scid, out_scid])
+        except Exception as e:
+            plugin.log("Adjusting fees: " + str(e), level="error")
 
 
 @plugin.init()
 def init(options: dict, configuration: dict, plugin: Plugin, **kwargs):
     plugin.our_node_id = plugin.rpc.getinfo()["id"]
-    plugin.adj_thread_pool = ThreadPoolExecutor(
-        max_workers=int(options["feeadjuster-par"])
-    )
     config = plugin.rpc.listconfigs()
     plugin.adj_basefee = config["fee-base"]
     plugin.adj_ppmfee = config["fee-per-satoshi"]
@@ -107,11 +98,8 @@ def init(options: dict, configuration: dict, plugin: Plugin, **kwargs):
         if chan["state"] != "CHANNELD_NORMAL":
             continue
 
-    plugin.log("Plugin feeadjuster initialized with {} threads"
-               .format(options["feeadjuster-par"]))
+    plugin.log("Plugin feeadjuster initialized ({} base / {} ppm)"
+               .format(plugin.adj_basefee, plugin.adj_ppmfee))
 
-
-plugin.add_option("feeadjuster-par", 8, "Maximum number of threads launched to"
-                  " adjust fees", opt_type="int")
 
 plugin.run()
--- a/feeadjuster/feeadjuster.py
+++ b/feeadjuster/feeadjuster.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
-from concurrent.futures import ThreadPoolExecutor
+import random
+import time
 from pyln.client import Plugin, Millisatoshi, RpcError
 
 
@@ -25,8 +26,14 @@ def maybe_adjust_fees(plugin: Plugin, scids: list):
         percentage = our / total
         last_percentage = plugin.adj_balances[scid].get("last_percentage")
 
-        # Only update on substantial balance moves to avoid flooding
-        if last_percentage is None or abs(last_percentage - percentage) > 0.05:
+        # Only update on substantial balance moves to avoid flooding, and add
+        # some pseudo-randomness to avoid too easy channel balance probing
+        update_treshold = 0.05
+        if not plugin.deactivate_fuzz:
+            update_treshold += random.uniform(-0.015, 0.015)
+
+        if (last_percentage is None
+                or abs(last_percentage - percentage) > update_treshold):
             ratio = get_ratio(percentage)
             try:
                 plugin.rpc.setchannelfee(scid, int(plugin.adj_basefee * ratio),
@@ -77,6 +84,9 @@ def forward_event(plugin: Plugin, forward_event: dict, **kwargs):
         plugin.adj_balances[in_scid]["our"] += forward_event["in_msatoshi"]
         plugin.adj_balances[out_scid]["our"] -= forward_event["out_msatoshi"]
         try:
+            # Pseudo-randomly add some hysterisis to the update
+            if not plugin.deactivate_fuzz and random.randint(0, 9) == 9:
+                time.sleep(random.randint(0, 5))
             maybe_adjust_fees(plugin, [in_scid, out_scid])
         except Exception as e:
             plugin.log("Adjusting fees: " + str(e), level="error")
@@ -85,6 +95,7 @@ def forward_event(plugin: Plugin, forward_event: dict, **kwargs):
 @plugin.init()
 def init(options: dict, configuration: dict, plugin: Plugin, **kwargs):
     plugin.our_node_id = plugin.rpc.getinfo()["id"]
+    plugin.deactivate_fuzz = options.get("feeadjuster-deactivate-fuzz", False)
     config = plugin.rpc.listconfigs()
     plugin.adj_basefee = config["fee-base"]
     plugin.adj_ppmfee = config["fee-per-satoshi"]
@@ -102,4 +113,10 @@ def init(options: dict, configuration: dict, plugin: Plugin, **kwargs):
                .format(plugin.adj_basefee, plugin.adj_ppmfee))
 
 
+plugin.add_option(
+    "feeadjuster-deactivate-fuzz",
+    False,
+    "Deactivate update threshold randomization and hysterisis.",
+    "flag"
+)
 plugin.run()
--- a/feeadjuster/feeadjuster.py
+++ b/feeadjuster/feeadjuster.py
@@ -1,0 +1,120 @@
+#!/usr/bin/env python3
+from concurrent.futures import ThreadPoolExecutor
+from pyln.client import Plugin, Millisatoshi, RpcError
+
+
+plugin = Plugin()
+# Our amount and the total amount in each of our channel, indexed by scid
+plugin.adj_balances = {}
+# Make requests to lightningd in parallel
+plugin.adj_thread_pool = None
+# Cache to avoid loads of calls to getinfo
+plugin.our_node_id = None
+
+
+def get_ratio(our_percentage):
+    """
+    Basic algorithm: the farther we are from the optimal case, the more we
+    bump/lower.
+    """
+    return 50**(0.5 - our_percentage)
+
+
+def get_fees(plugin: Plugin, scid: str):
+    for half in plugin.rpc.listchannels(scid)["channels"]:
+        if plugin.our_node_id in half["source"]:
+            return (half["base_fee_millisatoshi"], half["fee_per_millionth"])
+
+    # Note: the half may not be present, so this may actually return None!
+
+
+def maybe_adjust_fees(plugin: Plugin, scids: list):
+    for scid in scids:
+        # FIXME: set a threshold to avoid flooding!
+        if True:
+            # FIXME: it's exponential! We need to cache the startup
+            # fees instead of always factorising the updated ones.
+            fees = get_fees(plugin, scid)
+            if fees is None:
+                return
+
+            our = plugin.adj_balances[scid]["our"]
+            total = plugin.adj_balances[scid]["total"]
+            ratio = get_ratio(our / total)
+            try:
+                plugin.rpc.setchannelfee(scid, int(fees[0] * ratio),
+                                         int(fees[1] * ratio))
+                plugin.log("Adjusted fees of {} with a ratio of {}"
+                           .format(scid, ratio))
+            except RpcError as e:
+                plugin.log(str(e), level="warn")
+
+
+def get_chan(plugin: Plugin, scid: str):
+    for peer in plugin.rpc.listpeers()["peers"]:
+        if len(peer["channels"]) == 0:
+            continue
+        chan = peer["channels"][0]
+        if "short_channel_id" not in chan:
+            continue
+        if chan["short_channel_id"] == scid:
+            return chan
+
+
+def maybe_add_new_balances(plugin: Plugin, scids: list):
+    for scid in scids:
+        if scid not in plugin.adj_balances:
+            # At this point we could call listchannels and pass the scid as
+            # argument, and deduce the peer id (!our_id). But -as unlikely as
+            # it is- we may not find it in our gossip (eg some corruption of
+            # the gossip_store occured just before the forwarding event).
+            # However, it MUST be present in a listpeers() entry.
+            chan = get_chan(plugin, scid)
+            assert chan is not None
+
+            plugin.adj_balances[scid] = {
+                "our": int(chan["to_us_msat"]),
+                "total": int(chan["total_msat"])
+            }
+
+
+def threaded_forward_event(plugin: Plugin, forward_event: dict):
+    in_scid = forward_event["in_channel"]
+    out_scid = forward_event["out_channel"]
+    maybe_add_new_balances(plugin, [in_scid, out_scid])
+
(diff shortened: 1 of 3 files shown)


darosior:
Added two commits:
- We now only update the configuration's static `fee-base` and `fee-ppm`.
- We only update if the channel was [un/re]balanced by more than 5% of its total value *since last update*.

feeadjuster/feeadjuster.py
```diff
@@ -0,0 +1,126 @@
...
+            our = plugin.adj_balances[scid]["our"]
+            total = plugin.adj_balances[scid]["total"]
+            ratio = get_ratio(our / total)
+            try:
+                plugin.rpc.setchannelfee(scid, int(fees[0] * ratio),
+                                         int(fees[1] * ratio))
+                plugin.log("Adjusted fees of {} with a ratio of {}"
+                           .format(scid, ratio))
+            except RpcError as e:
+                plugin.log(str(e), level="warn")
```
cdecker:
```suggestion
                plugin.log("Could not adjust fees for channel {scid}: {e}".format(scid=scid, e=e), level="warn")
```

cdecker (changes requested):
Nice plugin, but I wonder if all the parallelism is really required: the notification to which we react is dispatched after the forwarding has completed, and `lightningd` will not wait for a response (it doesn't even return a response since it's a JSON-RPC notification). So doing this sequentially should work well enough without any chances of delaying the main daemon. Furthermore if you are doing these in parallel you will have contention on the global caches (`adj_balances`) and risk trampling them and getting them in an inconsistent state.

Fee adjustments are broadcast using a staggered broadcast anyway, so a quick succession of adjustments will get to the peer, and be aggregated into 1 minute batches that forward only the last one anyway.

Might I suggest adding a timeout of 10 seconds or so on the first update in a series, after which you trigger a single adjustment and let that spread before immediately overriding it again?

feeadjuster/feeadjuster.py
```diff
@@ -0,0 +1,126 @@
...
+    """
+    return 50**(0.5 - our_percentage)
+
+
+def get_fees(plugin: Plugin, scid: str):
+    for half in plugin.rpc.listchannels(scid)["channels"]:
+        if plugin.our_node_id in half["source"]:
+            return (half["base_fee_millisatoshi"], half["fee_per_millionth"])
+
+    # Note: the half may not be present, so this may actually return None!
```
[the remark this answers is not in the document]
darosior: >  Or we could add typing information so that mypy can catch these

Oh yeah i missed it: still not used to type annotations

feeadjuster/feeadjuster.py
```diff
@@ -0,0 +1,126 @@
...
+        if True:
+            # FIXME: it's exponential! We need to cache the startup
+            # fees instead of always factorising the updated ones.
+            fees = get_fees(plugin, scid)
+            if fees is None:
+                return
+
+            our = plugin.adj_balances[scid]["our"]
+            total = plugin.adj_balances[scid]["total"]
+            ratio = get_ratio(our / total)
```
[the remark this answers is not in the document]
darosior: Nice catch (even though i don't think it's possible to have a forward event with our == total)

darosior:
Yes, thread workers are overkill :)

darosior:
Also, what do you think of this conceptually ? And did you give some thoughts to what heuristics may be good to take into account ?

feeadjuster/feeadjuster.py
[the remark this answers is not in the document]
darosior: Hmm actually i'm not even sure it's bad to input 0 to get_ratio ?

darosior:
Argh my test is flaky on `pay`..

cdecker:
It's definitely an interesting idea, and I think long term everybody will want to do something like this. My main concern is that having the feerate reflect the balance of channels too directly is likely a privacy leak, which is why I'd like to add a bit of hysteresis (randomly delay updates, suppress updates below a threshold, ...) to reintroduce uncertainty for an observer. Likely we'd want to suppress updates if the rate is low and broadcasting a change in the balance of a series of channels could easily be collated into reconstructing the path the payment took.

darosior:
Thanks, i definitely overlooked the privacy aspects.

Thinking more about this, the equilibrirum may aim to be the `sum(our balance in each channel) / sum(total balance of every channel)` instead of an isolated 50% for each channel (which is a pretty irrealistic overall ratio in practice from my experience).

cdecker:
> Thinking more about this, the equilibrirum may aim to be the `sum(our balance in each channel) / sum(total balance of every channel)` instead of an isolated 50% for each channel (which is a pretty irrealistic overall ratio in practice from my experience).

Do you think the fees will have an effect of that metric? Fees are mostly important for senders that have to decide whether to route through your node or not. Routing through your node doesn't really alter the `sum(my_channel_balances)`, ignoring the tiny fee, so increasing or decreasing the fees doesn't really affect that metric either. Things are different when you're the trying to guide senders to rebalance an individual channel, where chosing the entry and exit channel makes a difference to the balances of those channels.

darosior:
> Do you think the fees will have an effect of that metric? Fees are mostly important for senders that have to decide whether to route through your node or not. Routing through your node doesn't really alter the sum(my_channel_balances), ignoring the tiny fee, so increasing or decreasing the fees doesn't really affect that metric either.

I mean that this currently incentivize users of our capacity to balance channels at 50%, one may prefer to have its channels balanced at say 30% because it has a higher incoming capacity.
Forcing 50% in this case would result in an unreachable "perfect state".

darosior:
> Might I suggest adding a timeout of 10 seconds or so on the first update in a series, after which you trigger a single adjustment and let that spread before immediately overriding it again?

Not sure i get this as well:
- there is no series ? Or maybe it is [input_chan, output_chan] ?
- gossipd's timeout of 1 minute is anyway higher than those 10 seconds so both would be broadcast at the same time ?

darosior:
Fixed (and speed!) up the tests by using `sendpay` instead of the somewhat flaky `pay`.

Added some randomization and fuzzed hysteresis. Pinging @ZmnSCPxj as i discussed this with them on IRC.

feeadjuster/feeadjuster.py
```diff
@@ -0,0 +1,128 @@
...
+        if "short_channel_id" not in chan:
+            continue
+        if chan["state"] != "CHANNELD_NORMAL":
+            continue
+        # We don't store it as a percentage, to be able to update in on
+        # forward without calling listpeers()
+        plugin.adj_balances[chan["short_channel_id"]] = {
+                "our": int(chan["to_us_msat"]),
+                "total": int(chan["total_msat"])
+        }
```
m-schmoock, line 116: Am I right that this block is not required as the channels are build up on demand anyway within `forward_event` and `maybe_add_new_balances`?

m-schmoock (commented):
Like it :D 
I will test this the next days. Some comment upfront and some feedback next week or so...

feeadjuster/feeadjuster.py
m-schmoock: Just checked: pytests are still green if this block is commented out.

m-schmoock (changes requested):

feeadjuster/feeadjuster.py
```diff
@@ -0,0 +1,128 @@
...
+        total = plugin.adj_balances[scid]["total"]
+        percentage = our / total
+        last_percentage = plugin.adj_balances[scid].get("last_percentage")
+
+        # Only update on substantial balance moves to avoid flooding, and add
+        # some pseudo-randomness to avoid too easy channel balance probing
+        update_treshold = 0.05
+        if not plugin.deactivate_fuzz:
+            update_treshold += random.uniform(-0.015, 0.015)
+
```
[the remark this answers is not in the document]
darosior: Timing it by 4 is a huge efficiency decrease, and completely removing the the update between 0.2 and 0.8 kind of defeats the point of using this.. Ideally you would help you to never get to the `[0; 0.2]U[0.8; 1]` zone (which often means for a consequent router an error to forward, with all the consequences).

m-schmoock:
You can/should set the threshold (i.e. 80%) as a config value.
Also zmnscpxj noted in IRC that the plugin could do a base fee estimation on init and repeat that from time to time by going through all peers and their peers and averaging base and ppm.

feeadjuster/feeadjuster.py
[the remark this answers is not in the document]
m-schmoock: on the other hand its more useful to set custom fees only if the imbalance is high enough. by setting it when you are at 40/60 you gain close to nothing but reveal you balance...

feeadjuster/feeadjuster.py
darosior: You are right, it's not required. The rationale was to avoid updating the whole list on first forward but at startup instead. Now i'm not sure it makes much sense

darosior: Ok, removed it for the sake of me not able to justify it clearly.

feeadjuster/feeadjuster.py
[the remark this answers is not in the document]
ZmnSCPxj: I think we should consider hysteresis as well.  For example, have a low/med/high feerate.  Then:

    if feerate is "low":
        if balance < 0.25:
            feerate = "high"
        elif balance < 0.5:
            feerate = "med"
    elif feerate is "med":
        if balance < 0.25:
            feerate = "high"
        elif balance > 0.75:
            feerate = "low"
    else: # feerate is "high":
        if balance > 0.75:
            feerate = "low"
        elif balance > 0.5:
            feerate = "med"

This gives better stability in case of noise in the balance signal.

Though as @darosior points out, it would be nice to have finer divisions so you never even enter the extreme zones.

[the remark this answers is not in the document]
darosior: > I think we should consider hysteresis as well

Don't you think the update threshold is hysterisis ?

> Though as @darosior points out, it would be nice to have finer divisions so you never even enter the extreme zones.

Yeah, so i think that instead of having fine-grained thresholds for performing the update, it's better to have thresholds for not performing this update (which is similar, but the logic is cleaner).
Also, this threshold is fuzzed in this PR.

darosior:
> You can/should set the threshold (i.e. 80%) as a config value.

As discussed on IRC and the comment above, i think it's useless to run that with a 80% threshold.

I could make the per-update threshold configurable in a follow-up, or here when i have a minute.

darosior:
Another thing i just thought about: this does really play well with the jitrebalance algo / plugin.

You are likely going to need a JIT rebalance if your channel is outbalanced. However, if a request comes in for an unbalanced channel they are paying a high fee. Therefore it increases the probability for a JIT rebalance to be economical and likely increases your average forwarding success rate.

feeadjuster/feeadjuster.py
```diff
@@ -0,0 +1,122 @@
...
+def maybe_add_new_balances(plugin: Plugin, scids: list):
+    for scid in scids:
+        if scid not in plugin.adj_balances:
+            # At this point we could call listchannels and pass the scid as
+            # argument, and deduce the peer id (!our_id). But -as unlikely as
+            # it is- we may not find it in our gossip (eg some corruption of
+            # the gossip_store occured just before the forwarding event).
+            # However, it MUST be present in a listpeers() entry.
+            chan = get_chan(plugin, scid)
+            assert chan is not None
```
darosior, line 69: It just failed :sob:

m-schmoock: Wired. Did a channel recently close?

darosior: Even if it recently closed it'd be still present in listpeers, going to add debug lines and hope for the best..
It did not happen to you ?

m-schmoock: Nah

m-schmoock: It can happen if scid changes on reorg maybe

feeadjuster/feeadjuster.py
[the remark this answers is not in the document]
ZmnSCPxj: >Don't you think the update threshold is hysterisis ?

My python-fu is awful, so I have no real idea.

More honestly, I have not actually read any of the code LOL.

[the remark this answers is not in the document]
ZmnSCPxj: On the other hand, IMO *any* feeadjuster is better than none, feel free to ignore me here, I have no idea about the issues a consequential router has.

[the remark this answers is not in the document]
m-schmoock: I might add that run in current form, it reveals the channel balance with an accuracy of >90% when monitored over a certain period of time:
 - find out base fees by taking median of all channels... repeat over time
 - apply reverse calculation on each channel
 - ignore the 1.5% fuzz ;)

Don't get me wrong, I really like it, hence I use it (with a threshhold). Thus we need to improve.

feeadjuster/feeadjuster.py
m-schmoock: Maybe indexing list element `0` failed?:  `chan = peer["channels"][0]`
Do you have a stracktrace?

m-schmoock: Maybe it was a peer where you had a prior channel (`0`) but the scid was on a newer channel with the same peer..  :D

darosior: > It can happen if scid changes on reorg maybe

:sweat_smile: 

> Do you have a stracktrace?

```
2020-10-15T18:46:58.816Z INFO    plugin-feeadjuster.py: Traceback (most recent call last):
2020-10-15T18:46:58.816Z INFO    plugin-feeadjuster.py:   File \"/home/bitcoin/.local/lib/python3.8/site-packages/pyln/client/plugin.py\", line 497, in _dispatch_notification
2020-10-15T18:46:58.816Z INFO    plugin-feeadjuster.py:     self._exec_func(func, request)
2020-10-15T18:46:58.816Z INFO    plugin-feeadjuster.py:   File \"/home/bitcoin/.local/lib/python3.8/site-packages/pyln/client/plugin.py\", line 455, in _exec_func
2020-10-15T18:46:58.816Z INFO    plugin-feeadjuster.py:     return func(*ba.args, **ba.kwargs)
2020-10-15T18:46:58.816Z INFO    plugin-feeadjuster.py:   File \"/home/bitcoin/.lightning/plugins/feeadjuster/feeadjuster.py\", line 82, in forward_event
2020-10-15T18:46:58.816Z INFO    plugin-feeadjuster.py:     maybe_add_new_balances(plugin, [in_scid, out_scid])
2020-10-15T18:46:58.816Z INFO    plugin-feeadjuster.py:   File \"/home/bitcoin/.lightning/plugins/feeadjuster/feeadjuster.py\", line 69, in maybe_add_new_balances
2020-10-15T18:46:58.816Z INFO    plugin-feeadjuster.py:     assert chan is not None
2020-10-15T18:46:58.816Z INFO    plugin-feeadjuster.py: AssertionError
```

> Maybe it was a peer where you had a prior channel (0) but the scid was on a newer channel with the same peer.. :D

Ooooh now that's interesting :D

m-schmoock: Closing/Resolving due to IRC discussion...

feeadjuster/feeadjuster.py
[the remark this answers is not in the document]
darosior: > it reveals the channel balance with an accuracy of >90% when monitored over a certain period of time:
> find out base fees by taking median of all channels... repeat over time
> apply reverse calculation on each channel

Did you do that ? Didn't the (fuzzed) 5% threshold messed your computation ?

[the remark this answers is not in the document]
m-schmoock: Nope, didn't do it tbh. But wouldn't that simply reduce accuracy by another 5%?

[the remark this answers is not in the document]
darosior: >  On the other hand, the existence of the threshold in the first place is the hystersis, I think....... hmmm.

Yeah, it's just that i fuzz the hysteresis (if that makes any sense). 

> I would go for a randomized-hourly update myself rather than on a forward notification but shrug --- I have not run a "consequential" router, so my thoughts may be bunk here

That decision was not based on any needs but to:
- Avoid scheduled (thus predictable?) updates
- Flatten load (don't broadcast 2 * n_channels update each hour)
- Be more accurate

> Another is that I would use absolute bin sizes instead of relative percent-based ones

Hmm in this case i guess their size should increase exponentially (as bitcoind fee estimation buckets)

> I have no idea about the issues a consequential router has.

This is also for us small-router operators! :p
I look forward to run (a correct version of) this on my node too and i hear your concerns regarding balance de-obfuscation.

However, i fail to see how a bucket system would be more private than the current threshold (which is basically, have buckets of size 5% (fuzzed)).

In fact, i fail to see how any trick we could add would make this more private by not losing efficiency: this **is** about sharing our balance.
I think we should instead concentrate on making it not directly gamable by probing.

cdecker:
Seems like my hysteresis comment caused quite a few headaches, sorry about that. I think we have three options to induce uncertainty in an observer here:

 - Make the fee adjustments more coarse-grained: fixed thresholds, hysteresis to reduce changes between thresholds, etc
 - Randomize the adjustments themselves: insert random adjustments to pretend there was a balance change when there wasn't, randomize the fee that we are announcing around the targeted level by adjusting up or down by 1-2 levels (depending on granularity and number of levels)
 - Randomize the temporal relationship between balance change and fee adjustment: deferring the fee updates until a later time, and deferring until the channel is quiescent (no need to adjust if we expect something else to require an adjustment right away again)

We can pick any combination of these, but I think the last ones are likely relatively easy to implement, and don't have much of an effect on efficiency of the adjustments (gossip broadcasts are staggered with a 1 minute timer at each hop, so these updates are far from real-time as it is). By no means should we implement them all 😉

darosior:
Thanks for the feedback @cdecker !

> Make the fee adjustments more coarse-grained: fixed thresholds

Done

> hysteresis to reduce changes between thresholds

Not sure i get this: between thresholds there is no change.

> insert random adjustments to pretend there was a balance change when there wasn't

I did'nt implement this, however the fuzzing of the update threshold could:
- trigger an update when none should have expectedly been triggered
- no trigger an update when one shoudl have expectedly been triggered

> Randomize the temporal relationship between balance change and fee adjustment

I have a (maybe weak) implementation of that. Should it we wait more / more often ?

https://github.com/lightningd/plugins/pull/147/files#diff-d7c357d8d243ae7ce4dadb79543de7072ea6601b10ae12579e10e6215a6299b0R91-R94

m-schmoock:
@cdecker @darosior 
> (gossip broadcasts are staggered with a 1 minute timer at each hop, so these updates are far from real-time as it is)

Maybe we should also check that we do not do updates more often than i.e. once per hour or by some 'smart'  metric. Because, if we update frequently (i.e. within 30 minutes) chances are high someone else has outdated fee data on us and we must fail routing attempts. Example:

We have an expected average route length for example 8 hops, don't know exactly, some number. Then after 8 minutes the network is up-to-date for an 'average route' (still some sooner and some later). This means if we update more often than each ~60 minutes, we increase the chance of outdated information on the 'average outskirt' of the network (from our viewpoint) to over ~13%. And if we update each 30 minutes this goes up by a factor of two to around 25% already...

Not sure if you can follow me, this reads like a mess ^^

cdecker:
Got interrupted by a phone call, so I didn't have time to look into the code yet :-)

> > hysteresis to reduce changes between thresholds
> 
> Not sure i get this: between thresholds there is no change.

My bad, my description is pretty vague. The idea is to have a margin around the thresholds to avoid flip-flopping continuously. Let's say we have a threshold at 50%, then we go from 49% to 51%, resulting in a broadcast, then we go have the inverse movement, resulting in yet another broadcast.

If we add a 5% hysteresis we would have to go above 55% to trigger a broadcast for the up movement, and below 45% to trigger a broadcast on the down movement. So broadcasts can only be triggered by large swings, and not by minute changes around the threshold.

>> Randomize the temporal relationship between balance change and fee adjustment
>
> I have a (maybe weak) implementation of that. Should it we wait more / more often ?
> 
> https://github.com/lightningd/plugins/pull/147/files#diff-d7c357d8d243ae7ce4dadb79543de7072ea6601b10ae12579e10e6215a6299b0R91-R94

That adds a 0-5 second delay, in 10% of the cases. That comes out to an expected delay of 1/4 second which is << 30 second expected delay added by each hop through the staggered broadcast, so probably doesn't add much obfuscation here.

> Maybe we should also check that we do not do updates more often than i.e. once per hour or by some 'smart' metric. Because, if we update frequently (i.e. within 30 minutes) chances are high someone else has outdated fee data on us and we must fail routing attempts. Example:

Definitely an issue. Outdated fees, especially when we raise them, can be really bad for route-finding.

If the desired effect of adjusting the fees is to encourage users to route through a given channel, we need to make sure that the payment initiators have enough time to learn about these changes (staggered broadcast...) and make use of the information (having a payment lined up they'd like to perform and that can take advantage of the lower/higher fees). I think for this reason alone any change that is quicker than once every 5 - 10 minutes is likely counterproductive.

Also if the fee adjustment is to rebalance channels, then I think active rebalancings once in a while or jitrebalancing when we actually have a reason to rebalance is more effective. We don't sit there idly, announcing a desire that maybe someday someone will take us up on our offer. An offer that is long-lived and cannot be updated quickly due to the delays. </snark> :wink: 

I still think that adjusting fees automatically is a really good feature, and I'd definitely use this plugin to guard against the 80%+ and 20%- depletion scenarios, where someone taking me up on my lower fee offer doesn't risk unbalancing me in the opposite direction directly again.

It's not an easy tradeoff to characterize and formalize, that's for sure, but I'm impressed with the discussion so far, keep it up :+1:

darosior:
> The idea is to have a margin around the thresholds to avoid flip-flopping continuously. Let's say we have a threshold at 50%, then we go from 49% to 51%, resulting in a broadcast, then we go have the inverse movement, resulting in yet another broadcast.

My threshold implementation is "at least 5% (default) balance update since last update" to trigger a threshold, so this doesn't apply ?

> If we add a 5% hysteresis we would have to go above 55% to trigger a broadcast for the up movement, and below 45% to trigger a broadcast on the down movement. So broadcasts can only be triggered by large swings, and not by minute changes around the threshold.

Yeah that was my motivation for implementing this :sweat_smile: 

> That adds a 0-5 second delay, in 10% of the cases. That comes out to an expected delay of 1/4 second which is << 30 second expected delay added by each hop through the staggered broadcast, so probably doesn't add much obfuscation here.

Yeah, right. Will make it actually useful :)

> Maybe we should also check that we do not do updates more often than i.e. once per hour or by some 'smart' metric.

I too am worried about this especially for a node with a high number of forwarding requests. Couldn't come up with a better answer than "i'll try and see the number of FEE_INSUFFICIENT errors" though :/

On the other hand, for a node with less forwarding requests (thus less probability to lose a forward fee because of a request sent at the same time an update was broadcast) i think it's another prevention for balance probing (as already mentioned by Christian in his first post).

> Also if the fee adjustment is to rebalance channels, then I think active rebalancings once in a while or jitrebalancing when we actually have a reason to rebalance is more effective. We don't sit there idly, announcing a desire that maybe someday someone will take us up on our offer. An offer that is long-lived and cannot be updated quickly due to the delays. wink

I disagree ! :D

Active rebalancing is super costy if you start having many channels.. And really is necessary to earn a decent amount of fees. Then you could periodically rebalance and increase your fees to be profitable. I observed that you need to set very high fees to be somewhat almost profitable-ish (.. until there is a force close --that's something else :p).
But this approach comports several drawbacks:
- You directly lose money with bulk rebalancings (jit yeah but in my experience it cannot currently be the sole option)
- You indirectly lose money with "missed opportunities" by having a feerate that high

Here you clearly need a way to lead your liquidity users toward where it's most profitable for you. I think that's what i'm trying to solve with this plugin: given enough requests, a more fine-grained management both avoids (hopefully completely, but at least partially) bulk rebalancings and you can still opportunistically forward cheaper payment for which you'd not have claimed a fee at all with a "bulk fee policy".

Regarding JIT rebalancing, i think there is conceptually a great synergy with dynamic fee adjustment: you'll need JIT rebalance for channels unbalanced on your side, thus costy to use, thus which have a far higher probability to end up with an economical rebalance thus a successful forward. 

> It's not an easy tradeoff to characterize and formalize, that's for sure, but I'm impressed with the discussion so far, keep it up +1

Yeah, I initially coded this up really mechanically as it seemed obvious but now you guys really got me thinking about it for real, Thanks!

feeadjuster/feeadjuster.py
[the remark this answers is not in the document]
ZmnSCPxj: > However, i fail to see how a bucket system would be more private than the current threshold (which is basically, have buckets of size 5% (fuzzed)).

When you move out of the current bucket, you then set the fee accurately according to the exact balance percent.  Thus, by reversing your computation, a third party knows thee exact balance percent at the moment you change the fee.

With a strict bucket system, once you move from one bucket to another, you set the fee according to the *center* of the bucket you entered, not the actual current exact balance percent.  This leaks less accurate information when you move out of your current bucket.

Now, if your exact balance moves, but still within the current bucket, you will not change the fee.  This leads to an inefficiency.  Thus, you would suffer inefficiency either way, especially if there is a large amount of traffic through your channels, when the balance moves a small amount.  But with a strict bucket system, you do not leak your exact balance at the time you change fees, only within the bucket size, which are the error bars that a third party can use to monitor you.

This is probably just a minor effect however.

cdecker:
> > The idea is to have a margin around the thresholds to avoid flip-flopping continuously. Let's say we have a threshold at 50%, then we go from 49% to 51%, resulting in a broadcast, then we go have the inverse movement, resulting in yet another broadcast.
> 
> My threshold implementation is "at least 5% (default) balance update since last update" to trigger a threshold, so this doesn't apply ?

Sorry, I must have overlooked that while looking through the code. Consider this addressed :+1:

> > Also if the fee adjustment is to rebalance channels, then I think active rebalancings once in a while or jitrebalancing when we actually have a reason to rebalance is more effective. We don't sit there idly, announcing a desire that maybe someday someone will take us up on our offer. An offer that is long-lived and cannot be updated quickly due to the delays. wink
> 
> I disagree ! :D
> 
> Active rebalancing is super costy if you start having many channels.. And really is necessary to earn a decent amount of fees. Then you could periodically rebalance and increase your fees to be profitable. I observed that you need to set very high fees to be somewhat almost profitable-ish (.. until there is a force close --that's something else :p).

I'll try not to push this point too far, after all the existence of the plugin means that there is interest from users to have a passive strategy by signalling fees, and mine is more of a philosophical counterargument than anything else :-) Ideally you'd set your feerate to something that is just profitable for you (assuming you aren't forwarding just to offset your own payment fees, which I think most people will end up doing), then any downward adjustment to the feerate in order to attract payments to forward means you're out of pocket (otherwise you could have set the feerate lower in the first place).

My point is that we should accept that keeping channels balanced is going to cost us, either by forwarding payments for fees that don't cover the costs, or by taking a more active approach and rebalancing ourselves. My argument is that the more active approach is desirable, because it doesn't rely on others taking us up on the low-fee incentive, and instead does exactly what we want to do. If we signal via fees, the remotely driven rebalancing comes with a delay, might not come at all, and will almost surely overshoot or undershoot our rebalancing target, because we don't have any control.

It's like steering a car over a video console that adds several seconds of delay to what we see, and to our input :wink:

> But this approach comports several drawbacks:
> 
> * You directly lose money with bulk rebalancings (jit yeah but in my experience it cannot currently be the sole option)
> * You indirectly lose money with "missed opportunities" by having a feerate that high

 * You only lose money that you'd have otherwise lost due to lowered fees to attract payments (sure, we can rebalance too often and lose more, but optimal rebalancing strategies should outperform fee incentivization schemes)
 * You should set a feerate that you're ok forwarding, anything below that is likely a loss. By making the feerate variable we introduce additional complexity when evaluating what feerate we should be picking, because sometimes we'll dip below profitability and sometimes we'll go above what we felt was a good return of investment, denying some forwarded payments.

Anyway, let's conclude this philosophical aside, and get back to the actual logic in this plugin. I'd be more than happy to be proven wrong (or right) on my position, but we can't do that without finalizing this plugin and see it perform in nature.

darosior:
Separating my answer to not lose focus, while keeping up this valuable discussion.

## About the plugin

> I'd be more than happy to be proven wrong (or right) on my position, but we can't do that without finalizing this plugin and see it perform in nature.

But it's finalized, isn't it ? I'd still like to add some improvements but it really is usable.

Your last request about the plugin was (if i'm not mistaken) to implement either of the three following mitigations:

1. **Make the fee adjustments more coarse-grained**: fixed thresholds, hysteresis to reduce changes between thresholds, etc
2. **Randomize the adjustments themselves**: insert random adjustments to pretend there was a balance change when there wasn't, randomize the fee that we are announcing around the targeted level by adjusting up or down by 1-2 levels (depending on granularity and number of levels)
3. **Randomize the temporal relationship between balance change and fee adjustment**: deferring the fee updates until a later time, and deferring until the channel is quiescent (no need to adjust if we expect something else to require an adjustment right away again)

--

1. Is implemented: fixed (configurable and fuzzed) threshold.
2. Is weakly implemented: going to fix that soon, as it's almost free efficiency-wise but seems to offer a good deniability in return.
3. Is not implemented. I'd argue that staggered gossip broadcast are already a step down this road and a significant stab at efficiency.


## About the rebalancing thoughts

> Ideally you'd set your feerate to something that is just profitable for you

Hmm ideally i'd set it at what is maximizing my profit on a certain period ? (We've been ignoring operating risks since the begining but it's i think highly in favour of this behaviour)

> assuming you aren't forwarding just to offset your own payment fees, which I think most people will end up doing

I've already heard this point, and i'm not sure about it nor it actually what's happening today ? Participants can be (once again according to my observations, you certainly studied this more than i did) mainly separated into two classes:
- Liquidity users: they usually have wallets and want to receive and send payments, they should not have public channels
- Liquidity providers: they operate nodes with public channels and don't usually make payments

I don't see how could, nor do i expect, users ending up running routers to compensate their fees (once LN is not unfairly cheap anymore :p).

> then any downward adjustment to the feerate in order to attract payments to forward means you're out of pocket (otherwise you could have set the feerate lower in the first place).

I'd say the opposite actually: in one scenario you lose money, in the other one you gain (less than optimal) money.

> My point is that we should accept that keeping channels balanced is going to cost us, either by forwarding payments for fees that don't cover the costs, or by taking a more active approach and rebalancing ourselves

I agree, but my point is the former being *way less* costy than the latter.
Also, one could argue that rebalancing your channel by making a payment through the right channel is almost a provided that should not need fees to cover its cost :)

> My argument is that the more active approach is desirable, because it doesn't rely on others taking us up on the low-fee incentive, and instead does exactly what we want to do. If we signal via fees, the remotely driven rebalancing comes with a delay, might not come at all, and will almost surely overshoot or undershoot our rebalancing target, because we don't have any control.

*Given enough forwarding request, all unbalanced channels are swallow*

^ should be the description for the `lnmeta` repo :p 

> and sometimes we'll go above what we felt was a good return of investment, denying some forwarded payments.

Which we'll happily accept if they're providing us with a little more than a direct gift in sats (eg the service of using the right channel).

cdecker:
> 3. Is not implemented. I'd argue that staggered gossip broadcast are already a step down this road and a significant stab at efficiency.
>    EDIT: actually am i right that we can just get probed by an "attacker" making us send WIRE_INSUFFICIENT_FEE messages to discover our actual feerate, ie approximative balance ?

Upon receiving a probe with an insufficient feerate the node will reply with an error that contains the `channel_update` which includes the current feerate. This is one reason why relying solely on the staggered broadcast is not sufficient :-)

I'll review and merge the code today :-)

feeadjuster/feeadjuster.py
```diff
@@ -0,0 +1,136 @@
...
+# Users can configure this
+plugin.update_treshold = 0.05
+
+
+def get_ratio(our_percentage):
+    """
+    Basic algorithm: the farther we are from the optimal case, the more we
+    bump/lower.
+    """
+    return 50**(0.5 - our_percentage)
```
cdecker, line 21: Just computed the range of these functions to make sure I understand them:

 - `x=1` -> 0.14 ~= 1/7.14
 - `x=0.5` -> 1
 - `x=0` -> 7.0.7

So the ratio moves in the range `[1/7; 7]` approximately.

feeadjuster/feeadjuster.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+def maybe_adjust_fees(plugin: Plugin, scids: list):
+    for scid in scids:
+        our = plugin.adj_balances[scid]["our"]
```
cdecker, line 26: Must ensure that `scid in plugin.adj_balances`.

feeadjuster/feeadjuster.py
cdecker: I wonder if we should normalize this, to allow the user to specify the range.

feeadjuster/feeadjuster.py
```diff
@@ -0,0 +1,136 @@
...
+        update_treshold = plugin.update_treshold
+        if not plugin.deactivate_fuzz:
+            update_treshold += random.uniform(-0.015, 0.015)
+
+        if (last_percentage is None
+                or abs(last_percentage - percentage) > update_treshold):
+            ratio = get_ratio(percentage)
+            try:
+                plugin.rpc.setchannelfee(scid, int(plugin.adj_basefee * ratio),
+                                         int(plugin.adj_ppmfee * ratio))
```
cdecker, line 42: This is where the range of the ratio function is important. As is this allows only an adjustment to `[1/7;7]` times the normal fee.

feeadjuster/feeadjuster.py
```diff
@@ -0,0 +1,136 @@
...
+                plugin.adj_balances[scid]["last_percentage"] = percentage
+            except RpcError as e:
+                plugin.log("Could not adjust fees for channel {}: '{}'"
+                           .format(scid, e), level="warn")
+
+
+def get_chan(plugin: Plugin, scid: str):
+    for peer in plugin.rpc.listpeers()["peers"]:
+        if len(peer["channels"]) == 0:
+            continue
```
cdecker, line 54: Not really useful, the loop below will do the same.

feeadjuster/feeadjuster.py
```diff
@@ -0,0 +1,136 @@
...
+        out_scid = forward_event["out_channel"]
+        maybe_add_new_balances(plugin, [in_scid, out_scid])
+
+        plugin.adj_balances[in_scid]["our"] += forward_event["in_msatoshi"]
+        plugin.adj_balances[out_scid]["our"] -= forward_event["out_msatoshi"]
+        try:
+            # Pseudo-randomly add some hysterisis to the update
+            if not plugin.deactivate_fuzz and random.randint(0, 9) == 9:
+                time.sleep(random.randint(0, 5))
+            maybe_adjust_fees(plugin, [in_scid, out_scid])
```
cdecker, line 94: Doing this in the thread handling the subscription may create a backlog on very busy nodes. But should be good enough for nodes that handle less than 1 event every 2.5 seconds.

feeadjuster/feeadjuster.py
```diff
@@ -0,0 +1,136 @@
...
+        except Exception as e:
+            plugin.log("Adjusting fees: " + str(e), level="error")
```
cdecker, line 96:
```suggestion
            plugin.log("Adjusting fees: {e}".format(e=e), level="error")
```
`format` can be more efficient than string concatenation (might not be in this case with just 1 placeholder but it's a good habit to have).

cdecker (approved):
Just some minor comments that can be addressed in a separate PR.

ACK https://github.com/lightningd/plugins/pull/147/commits/11b9c5dfd68dd7e9abf8954ca26ecdf8065976b3

feeadjuster/feeadjuster.py
darosior: i should add a comment with this

Re normalization: my plan was to have many such "ratio getters" with diffferent formulae that the user could activate using an --adjustment-method option (eg simple, conservative, exhaustive, linear, etc..)

feeadjuster/feeadjuster.py
darosior: Could be an assert as it's called just after maybe_populate_balances

feeadjuster/feeadjuster.py
darosior: I wanted to be conservative with variations by default, but really happy to bikeshed my tinkered formula :)

feeadjuster/feeadjuster.py
darosior: Yeah i should really rework that... And would be back to the initial multi threading :)

feeadjuster/feeadjuster.py
darosior: Think i can even use fstrings now ? :D

cdecker: Oh that's true, way better :+1:

m-schmoock:
:) that was quick. Will check it anyway. We still can address anything important later

PR #150, opened 2020-09-29
Fixing bugs in autopilot

Moller40:
Avoiding zero division error and improve calculation of available funds.

### DIFF
--- a/autopilot/autopilot.py
+++ b/autopilot/autopilot.py
@@ -133,7 +133,8 @@ def init(configuration, options, plugin):
 def run_once(plugin, dryrun=False):
     # Let's start by inspecting the current state of the node
     funds = plugin.rpc.listfunds()
-    output_funds = sum([o['value'] for o in funds['outputs'] if o['status'] == 'confirmed'])
+    awaiting_lockin_funds = sum([o['channel_sat'] for o in funds['channels'] if o['state'] == 'CHANNELD_AWAITING_LOCKIN'])
+    output_funds = sum([o['value'] for o in funds['outputs'] if o['status'] == 'confirmed']) - awaiting_lockin_funds
     channels = funds['channels']
     available_funds = output_funds / 100.0 * plugin.percent
 
--- a/autopilot/autopilot.py
+++ b/autopilot/autopilot.py
@@ -140,6 +140,15 @@ def run_once(plugin, dryrun=False):
     # Now we can look whether and how we'd like to open new channels. This
     # depends on available funds and the number of channels we were configured
     # to open
+
+    if available_funds < plugin.min_capacity_sat:
+        print("Too low available funds: {} < {}".format(available_funds, plugin.min_capacity_sat))
+        return False
+
+    if len(channels) >= plugin.num_channels:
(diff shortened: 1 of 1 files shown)


autopilot/autopilot.py
```diff
@@ -133,13 +133,23 @@ def init(configuration, options, plugin):
 def run_once(plugin, dryrun=False):
     # Let's start by inspecting the current state of the node
     funds = plugin.rpc.listfunds()
-    output_funds = sum([o['value'] for o in funds['outputs'] if o['status'] == 'confirmed'])
+    awaiting_lockin_funds = sum([o['channel_sat'] for o in funds['channels'] if o['state'] == 'CHANNELD_AWAITING_LOCKIN'])
+    output_funds = sum([o['value'] for o in funds['outputs'] if o['status'] == 'confirmed']) - awaiting_lockin_funds
```
cdecker, line 137: I'm not sure subtracting the `awaiting_lockin_funds` is correct here: for transactions we send ourselves, such as funding transactions, we will mark the outputs as spent immediately, so the funds that are allocated to channels waiting for confirmation should already be subtracted in the `outputs` array.

Did you encounter an issue with the old computation?

Moller40: Sorry, my change is wrong. It only works in some cases. However, the old computation is also wrong. It only works if all is confirmed and nice.

Funds that are allocated to channels waiting for confirmation is NOT subtracted from the outputs array. They are subtracted first when the channel transaction is confirmed, I think...

However, checked a bit deeper and the "reserved" field should be possible to use. Then line 136 could change to

`output_funds = sum([o['value'] for o in funds['outputs'] if o['reserved'] == 'false')`

But there are an other problem. The "reserved" seems only valid for 12 hours. Found this code:
```
/* 12 hours is usually enough reservation time */
#define RESERVATION_INC (6 * 12)
```
So after 12 hours, if still unconfirmed, the calculation gets wrong anyway.
Another way may be to check the transactions and exclude any tx that is used as input. But it gets complicated...

I think I just remove this part of this PR.

Moller40: Checked some more. I think the following change will always calculate correct amount of available output funds, including from unconfirmed transactions.
```suggestion
    transactions = plugin.rpc.listtransactions()
    unconfirmed_transaction_inputs = [t['inputs'] for t in transactions['transactions'] if t['blockheight'] == 0]
    input_utxos = [input_utxo for sublist in unconfirmed_transaction_inputs for input_utxo in sublist]
    reserved_outputs = ["{}:{}".format(i['txid'], i['index']) for i in input_utxos]
    output_funds = sum([o['value'] for o in funds['outputs'] if not o['reserved'] and
        "{}:{}".format(o['txid'], o['output']) not in reserved_outputs])
```
However, if we are passed the 12 hour reservation time it may get us in trouble anyway (if not dryrun) since transaction can then be created with inputs that already is used in some older transaction that not yet is confirmed.  This seems to be a general problem in how c-lightning handles unconfirmed transactions. Only way around it right now seems to be to manually and carefully select what utxos to use when funding.

cdecker:
ACK https://github.com/lightningd/plugins/pull/150/commits/850190c6864fc7ee26dd533b212c49301a3db09d

PR #151, opened 2020-10-03
jitrebalance: channel lookup can fail

darosior:
This fixes the crashes i've been experiencing for a while :tada:


cdecker:
ACK https://github.com/lightningd/plugins/pull/151/commits/a6f4e5dcce90535277bec04c406fa88ae87bffca

PR #152, opened 2020-10-21
prometheus plugin: Added node alias to channel label for human readable legend

Zathrasss:
This contribution provides prometheus with an extra label option for the channels. You can now choose for peer id, channel id, and node alias. (this helped me to understand the graph's better without having to look up the id every time.) 

i notice now that the default listening port has changed, please edit that line away (i think that is just my config). 

Since it's my first commit ever, please provide feedback.

### DIFF
--- a/prometheus/prometheus.py
+++ b/prometheus/prometheus.py
@@ -97,72 +97,79 @@ class ChannelsCollector(BaseLnCollector):
         balance_gauge = GaugeMetricFamily(
             'lightning_channel_balance',
             'How many funds are at our disposal?',
-            labels=['id', 'scid'],
+            labels=['id', 'scid', 'alias'],
         )
         spendable_gauge = GaugeMetricFamily(
             'lightning_channel_spendable',
             'How much can we currently send over this channel?',
-            labels=['id', 'scid']
+            labels=['id', 'scid', 'alias'],
         )
         total_gauge = GaugeMetricFamily(
             'lightning_channel_capacity',
             'How many funds are in this channel in total?',
-            labels=['id', 'scid'],
+            labels=['id', 'scid', 'alias'],
         )
         htlc_gauge = GaugeMetricFamily(
             'lightning_channel_htlcs',
             'How many HTLCs are currently active on this channel?',
-            labels=['id', 'scid'],
+            labels=['id', 'scid', 'alias'],
         )
 
         # Incoming routing statistics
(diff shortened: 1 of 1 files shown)


prometheus/prometheus.py
```diff
@@ -97,72 +97,79 @@ def collect(self):
...
         peers = self.rpc.listpeers()['peers']
         for p in peers:
             for c in p['channels']:
-                labels = [p['id'], c.get('short_channel_id', c.get('channel_id'))]
+	        # append alias for human readable labels, if no label is found fill with shortid.
+                node = self.rpc.listnodes(p['id'])['nodes']
+                if len(node) != 0 and 'alias' in node[0]:
+                    alias = node[0]['alias']
+                else:
+                    alias = c.get('short_channel_id', c.get('channel_id'))
```
darosior, line 170: Hmm we already have the scid, and (personal flavour here) i don't find the channel_id to be an useful user interface.
Could we just explicit it's unknown in this case ?
```python
alias = node[0].get("alias", "unknown")
```

prometheus/prometheus.py
```diff
@@ -213,7 +220,7 @@ def init(options, configuration, plugin):
 
 plugin.add_option(
     'prometheus-listen',
-    '0.0.0.0:9750',
+    '0.0.0.0:9900',
```
darosior, line 223: You should not modify this, `9750` we modified it lately as it's now The Official C-lightning Prometheus Port :)

darosior (commented):
Nice! Looking forward to use this instead of scids on https://grafana.com/grafana/dashboards/12397 :D

prometheus/prometheus.py
cdecker: Problem is that switching from `short_channel_id` to the alias will cause a single timeseries to show up as two different series in the plot, which is quite confusing. I'd prefer just using a placeholder or leaving it out completely until we know the alias. For reference: we won't know the alias until our peer has a) opened and confirmed a channel with at least 6 confirmations and b) sent us a `node_announcement`.

cdecker (changes requested):
Please remove the change of the port, and I personally prefer leaving the `short_channel_id` unset instead of flipping it some time after being confirmed (~1/2 hour after funding is confirmed).

darosior:
@Zathrasss what's the state of this one ? I'm eager to use it on my dashboard :-)

cdecker (approved):

PR #156, opened 2020-11-11
backup: Do not buffer early writes, instead rely on backup.lock to find and write to the backend

### LINKED ISSUE #155: backup: Backup plugin gets out of sync if `init` doesn't complete
Occasionally I get an emergency shutdown from the backup plugin:

``` text
2020-11-11T10:20:18.023Z INFO    plugin-backup.py: Comparing backup version 1308057 versus first write version 1308060
2020-11-11T10:20:18.023Z INFO    plugin-backup.py: Backup is out of date, we cannot continue safely. Emergency shutdown.
2020-11-11T10:20:18.034Z INFO    plugin-summary.py: Plugin summary.py initialized
2020-11-11T10:20:19.026Z INFO    plugin-backup.py: Killing process lightningd (22862)
```

This is caused by an incomplete startup, in which the plugin started
buffering the pre-init `db_write`s but then couldn't flush them on
`init`. Since 97c702731 we are writing the `backup.lock` file which
contains all the necessary information for us to start writing the
changes to the backup right away and don't have to wait for the `init`
message (previously needed to get CLI arguments).

We should really just be reading this file and start writing the
changes to the file without buffering.

cdecker:
We address #155 by deprecating the `--backup-destination` CLI option, replacing it outright with the discovery based on the `backup.lock` file, which is generated by `backup-cli init [lightning-dir] [backend-uri]`. This means that we can immediately start appending changes to the backup without having to wait for the `init` message from `lightningd` (at which point we'd already have some writes buffered).

The `--backup-destination` flag is still present so we don't break any config files, but we warn about it being present, and ask the operator to fix their config.

Fixes #155

### DIFF
--- a/backup/backup.py
+++ b/backup/backup.py
@@ -299,6 +299,23 @@ def on_db_write(writes, data_version, plugin, **kwargs):
         kill("Could not append DB change to the backup. Need to shutdown!")
 
 
+@plugin.init()
+def on_init(options, **kwargs):
+    dest = options.get('backup-destination', 'null')
+    if dest != 'null':
+        plugin.log(
+            "The `--backup-destination` option is deprecated and will be "
+            "removed in future versions of the backup plugin. Please remove "
+            "it from your configuration. The destination is now determined by "
+            "the `backup.lock` file in the lightning directory",
+            level="warn"
+        )
+
+    configs = plugin.rpc.listconfigs()
+    if not configs['wallet'].startswith('sqlite3'):
+        kill("The backup plugin only works with the sqlite3 database.")
+
+
 def kill(message: str):
     plugin.log(message)
     time.sleep(1)
--- a/backup/backup.py
+++ b/backup/backup.py
@@ -335,10 +335,6 @@ def on_init(options: Mapping[str, str], plugin: Plugin, **kwargs):
         apply_write(plugin, c)
 
 
-plugin.add_option(
-    'backup-destination', None,
-    'Destination of the database backups (file:///filename/on/another/disk/).'
-)
 
 
 def kill(message: str):
@@ -365,6 +361,12 @@ def preflight(plugin: Plugin):
     """
     if not os.path.exists("backup.lock"):
         kill("Could not find backup.lock in the lightning-dir")
+
+plugin.add_option(
+    'backup-destination', None,
+    'UNUSED. Kept for backward compatibility only. Please update your configuration to remove this option.'
+)
+
 
 if __name__ == "__main__":
     preflight(plugin)
--- a/backup/backup.py
+++ b/backup/backup.py
@@ -289,52 +289,14 @@ def check_first_write(plugin, data_version):
 @plugin.hook('db_write')
 def on_db_write(writes, data_version, plugin, **kwargs):
     change = Change(data_version, None, writes)
-    if not hasattr(plugin, 'backend'):
-        plugin.early_writes.append(change)
-        return {"result": "continue"}
-    else:
-        return apply_write(plugin, change)
-
-
-def apply_write(plugin, change):
     if not plugin.initialized:
         assert(check_first_write(plugin, change.version))
         plugin.initialized = True
 
     if plugin.backend.add_change(change):
         return {"result": "continue"}
-
-
-@plugin.init()
-def on_init(options: Mapping[str, str], plugin: Plugin, **kwargs):
-    # Reach into the DB and
-    configs = plugin.rpc.listconfigs()
-    plugin.db_path = configs['wallet']
-    destination = options['backup-destination']
-
-    # Ensure that we don't inadventently switch the destination
-    if not os.path.exists("backup.lock"):
-        print("Files in the current directory {}".format(", ".join(os.listdir("."))))
-        kill("Could not find backup.lock in the lightning-dir, have you initialized using the backup-cli utility?")
-
-    d = json.load(open("backup.lock", 'r'))
-    if destination is None or destination == 'null':
-        destination = d['backend_url']
-    elif destination != d['backend_url']:
-        kill(
-            "The destination specified as option does not match the one "
-            "specified in backup.lock. Please check your settings"
-        )
-
-    if not plugin.db_path.startswith('sqlite3'):
-        kill("The backup plugin only works with the sqlite3 database.")
-
-    plugin.backend = get_backend(destination, require_init=True)
-
-    for c in plugin.early_writes:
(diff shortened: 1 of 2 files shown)


backup/backup.py
```diff
@@ -289,56 +289,28 @@ def check_first_write(plugin, data_version):
...
+        plugin.log(
+            "The `--backup-destination` option is deprecated and will be "
+            "removed in future versions of the backup plugin. Please remove "
+            "it from your configuration. The destination is now determined by "
+            "the `backup.lock` file in the lightning directory",
+            level="warn"
         )
 
-    if not plugin.db_path.startswith('sqlite3'):
-        kill("The backup plugin only works with the sqlite3 database.")
```
m-schmoock, line 330: The sqlite3 only check is removed by this change, as it is not added anywhere else. Was this intended?

m-schmoock (commented):
Tested and works.
Note1: The failing travis is jitrebalance related.
Note2: you maybe want to re-add the sqlite3 only check if it was removed unwittingly.

m-schmoock:
@cdecker sorry for bugging, but can you re-add the sqlite3 check and merge this?

cdecker:
Hm, I thought I had replied to this earlier, but it looks like I forgot.

The problem with the sqlite3 check is that at the time we perform the check (`init`) we've already performed some writes (the ones we previously buffered until we got the `init`). 

Might be alright though, better checking late than never 😉

m-schmoock:
> Might be alright though, better checking late than never 😉

In this case: your call ;)
I just try to pull all the loosing ends together at the plugin repo from time to time.

cdecker:
Re-added the `sqlite3` check :+1:

m-schmoock:
Strange that it worked before, but I get a stacktrace now:

```
2020-11-25T14:11:10.042Z INFO    plugin-backup.py: Traceback (most recent call last):
2020-11-25T14:11:10.042Z INFO    plugin-backup.py:   File \"/home/btc/lightning.git/contrib/pyln-client/pyln/client/plugin.py\", line 564, in _dispatch_request
2020-11-25T14:11:10.042Z INFO    plugin-backup.py:     result = self._exec_func(method.func, request)
2020-11-25T14:11:10.042Z INFO    plugin-backup.py:   File \"/home/btc/lightning.git/contrib/pyln-client/pyln/client/plugin.py\", line 549, in _exec_func
2020-11-25T14:11:10.042Z INFO    plugin-backup.py:     return func(*ba.args, **ba.kwargs)
2020-11-25T14:11:10.042Z INFO    plugin-backup.py:   File \"/home/btc/lightning.git/contrib/pyln-client/pyln/client/plugin.py\", line 796, in _init
2020-11-25T14:11:10.042Z INFO    plugin-backup.py:     return self._exec_func(self.child_init, request)
2020-11-25T14:11:10.042Z INFO    plugin-backup.py:   File \"/home/btc/lightning.git/contrib/pyln-client/pyln/client/plugin.py\", line 549, in _exec_func
2020-11-25T14:11:10.042Z INFO    plugin-backup.py:     return func(*ba.args, **ba.kwargs)
2020-11-25T14:11:10.042Z INFO    plugin-backup.py:   File \"/home/btc/.lightning/plugins/backup.py\", line 314, in on_init
2020-11-25T14:11:10.042Z INFO    plugin-backup.py:     if not plugin.db_path.startswith('sqlite3'):
2020-11-25T14:11:10.042Z INFO    plugin-backup.py: AttributeError: 'Plugin' object has no attribute 'db_path'
```

m-schmoock:
@cdecker  Update:  its missing the `plugin.db_path = configs['wallet']` at the top of the init ;)

backup/backup.py
```diff
@@ -289,57 +289,31 @@ def check_first_write(plugin, data_version):
...
-
-    d = json.load(open("backup.lock", 'r'))
-    if destination is None or destination == 'null':
-        destination = d['backend_url']
-    elif destination != d['backend_url']:
-        kill(
-            "The destination specified as option does not match the one "
-            "specified in backup.lock. Please check your settings"
+def on_init(options, **kwargs):
+    dest = options.get('backup-destination', 'null')
```
m-schmoock, line 304: We are missing this here, otherwise the sqlite3 check crashes :

```
    configs = plugin.rpc.listconfigs()
    plugin.db_path = configs['wallet']
```

m-schmoock (changes requested):
Small bug, otherwise fine.

backup/backup.py
cdecker: Ouch, right. Added the `listconfigs` call again, but not storing it in the `plugin` since we never use it again.

m-schmoock (commented):
Almost there. Now the plugin prints some strange output blob in INFO regarding the listconfigs call.
I really don't know why this happens, but it does.

```
2020-11-25T15:17:37.359Z INFO    plugin-backup.py: Received response for listconfigs call: {'jsonrpc': '2.0', 'id': 0, 'result': 
{'# version': 'v0.9.2', 'lightning-dir': '/home/btc/.lightning', 'network': 'bitcoin', 'allow-deprecated-apis': True, 'rpc-file': 'lightning-rpc', 
'plugins': [{'path': '/home/btc/.lightning/plugins/backup.py', 'name': 'backup.py', 'options': {'backup-destination': 'null'}}, {'path': 
'/home/btc/.lightning/plugins/drain/drain.py', 'name': 'drain.py', 'options': {'cltv-final': '10'}}, {'path': 
'/home/btc/.lightning/plugins/rebalance/rebalance.py', 'name': 'rebalance.py', 'options': {'cltv-final': '10'}}, {'path': 
'/home/btc/.lightning/plugins/helpme/helpme.py', 'name': 'helpme.py'}, {'path': 
'/home/btc/.lightning/plugins/feeadjuster/feeadjuster.py', 'name': 'feeadjuster.py', 'options': {'feeadjuster-deactivate-fuzz': False, 
'feeadjuster-threshold': '0.05', 'feeadjuster-adjustment-method': 'default', 'feeadjuster-imbalance': '0.3'}}, {'path': 
'/home/btc/.lightning/plugins/sendinvoiceless/sendinvoiceless.py', 'name': 'sendinvoiceless.py', 'options': {'cltv-final': '10', 'fee-
base': 'null', 'fee-per-satoshi': 'null'}}, {'path': '/home/btc/.lightning/plugins/summary/summary.py'... 
```

m-schmoock:
ACK 0a816b9

m-schmoock (approved):

PR #157, opened 2020-11-12
summary: use proxy if configuration says to.

rustyrussell:
All plugins should take care to do this if config
specifies always-use-proxy!

Signed-off-by: Rusty Russell <rusty@rustcorp.com.au>

### DIFF
--- a/summary/requirements.txt
+++ b/summary/requirements.txt
@@ -1,3 +1,4 @@
 pyln-client>=0.7.3
-requests>=2.0.0
+requests>=2.10.0
+requests[socks]>=2.10.0
 packaging>=14.1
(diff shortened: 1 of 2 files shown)


summary/requirements.txt
```diff
@@ -1,3 +1,4 @@
 pyln-client>=0.7.3
-requests>=2.0.0
+requests>=2.10.0
+requests[socks]>=2.10.0
```
cdecker, line 3: The latter supercedes the former, but it works anyway :-)

cdecker (approved):
ACK https://github.com/lightningd/plugins/pull/157/commits/e50fd5a19bccfaf4b5fc960a8a778a204722076d

PR #164, opened 2020-11-30
feeadjuster: option to handle big channels differently

gallizoltan:
- nit: lots of fstring

### DIFF
--- a/feeadjuster/feeadjuster.py
+++ b/feeadjuster/feeadjuster.py
@@ -20,6 +20,8 @@ def get_adjusted_percentage(plugin: Plugin, scid: str):
         then percentage is actually {our} / {total}, as it was before
     """
     channel = plugin.adj_balances[scid]
+    if plugin.big_enough_liquidity == Millisatoshi(0):
+        return channel["our"] / channel["total"]
     min_liquidity = min(channel["total"] / 2, int(plugin.big_enough_liquidity))
     theirs = channel["total"] - channel["our"]
     if channel["our"] >= min_liquidity and theirs >= min_liquidity:
@@ -200,13 +202,14 @@ def init(options: dict, configuration: dict, plugin: Plugin, **kwargs):
     plugin.deactivate_fuzz = options.get("feeadjuster-deactivate-fuzz", False)
     plugin.update_threshold = float(options.get("feeadjuster-threshold", "0.05"))
     plugin.update_threshold_abs = Millisatoshi(options.get("feeadjuster-threshold-abs", "0.001btc"))
-    plugin.big_enough_liquidity = Millisatoshi(options.get("feeadjuster-enough-liquidity", "0.02btc"))
+    plugin.big_enough_liquidity = Millisatoshi(options.get("feeadjuster-enough-liquidity", "0msat"))
     plugin.imbalance = float(options.get("feeadjuster-imbalance", 0.5))
-    plugin.get_ratio = get_ratio
-    if options.get("feeadjuster-adjustment-method", "default") == "soft":
-        plugin.get_ratio = get_ratio_soft
-    if options.get("feeadjuster-adjustment-method", "default") == "hard":
-        plugin.get_ratio = get_ratio_hard
+    adjustment_switch = {
+        "soft": get_ratio_soft,
+        "hard": get_ratio_hard,
+        "default": get_ratio
+        }
+    plugin.get_ratio = adjustment_switch.get(options.get("feeadjuster-adjustment-method", "default"), get_ratio)
     config = plugin.rpc.listconfigs()
     plugin.adj_basefee = config["fee-base"]
     plugin.adj_ppmfee = config["fee-per-satoshi"]
@@ -247,8 +250,8 @@ plugin.add_option(
 )
 plugin.add_option(
     "feeadjuster-enough-liquidity",
-    "0.02btc",
-    "Beyond this liquidity threshold there is no need to adjusting fees",
+    "0msat",
+    "Beyond this liquidity threshold there is no need to adjusting fees, 0 means the feature is turned off",
     "string"
 )
 plugin.add_option(
--- a/feeadjuster/feeadjuster.py
+++ b/feeadjuster/feeadjuster.py
@@ -17,7 +17,7 @@ def get_adjusted_percentage(plugin: Plugin, scid: str):
     """
     For big channels, there may be a wide range where the liquidity is just okay.
     Note: if big_enough_liquidity is greater than {total} * 2
-        then percentage is actually {our} / {total}, as it was before
+          then percentage is actually {our} / {total}, as it was before
     """
     channel = plugin.adj_balances[scid]
     if plugin.big_enough_liquidity == Millisatoshi(0):
@@ -72,7 +72,7 @@ def maybe_setchannelfee(plugin: Plugin, scid: str, base: int, ppm: int):
         plugin.rpc.setchannelfee(scid, base, ppm)
         return True
     except RpcError as e:
-        plugin.log(f"Could not adjust fees for channel {scid}: '{e}'", level="warn")
+        plugin.log(f"Could not adjust fees for channel {scid}: '{e}'", level="error")
     return False
 
 
@@ -199,17 +199,17 @@ def feeadjust(plugin: Plugin):
 @plugin.init()
 def init(options: dict, configuration: dict, plugin: Plugin, **kwargs):
     plugin.our_node_id = plugin.rpc.getinfo()["id"]
-    plugin.deactivate_fuzz = options.get("feeadjuster-deactivate-fuzz", False)
-    plugin.update_threshold = float(options.get("feeadjuster-threshold", "0.05"))
-    plugin.update_threshold_abs = Millisatoshi(options.get("feeadjuster-threshold-abs", "0.001btc"))
-    plugin.big_enough_liquidity = Millisatoshi(options.get("feeadjuster-enough-liquidity", "0msat"))
-    plugin.imbalance = float(options.get("feeadjuster-imbalance", 0.5))
+    plugin.deactivate_fuzz = options.get("feeadjuster-deactivate-fuzz")
+    plugin.update_threshold = float(options.get("feeadjuster-threshold"))
+    plugin.update_threshold_abs = Millisatoshi(options.get("feeadjuster-threshold-abs"))
+    plugin.big_enough_liquidity = Millisatoshi(options.get("feeadjuster-enough-liquidity"))
+    plugin.imbalance = float(options.get("feeadjuster-imbalance"))
     adjustment_switch = {
         "soft": get_ratio_soft,
         "hard": get_ratio_hard,
         "default": get_ratio
-        }
-    plugin.get_ratio = adjustment_switch.get(options.get("feeadjuster-adjustment-method", "default"), get_ratio)
+    }
+    plugin.get_ratio = adjustment_switch.get(options.get("feeadjuster-adjustment-method"), get_ratio)
     config = plugin.rpc.listconfigs()
     plugin.adj_basefee = config["fee-base"]
     plugin.adj_ppmfee = config["fee-per-satoshi"]
@@ -251,7 +251,9 @@ plugin.add_option(
 plugin.add_option(
     "feeadjuster-enough-liquidity",
     "0msat",
-    "Beyond this liquidity threshold there is no need to adjusting fees, 0 means the feature is turned off",
+    "Beyond this liquidity do not adjust fees. "
+    "This also modifies the fee curve to achieve having this amount of liquidity. "
+    "Default: '0msat' (turned off).",
     "string"
 )
 plugin.add_option(
--- a/feeadjuster/feeadjuster.py
+++ b/feeadjuster/feeadjuster.py
@@ -2,6 +2,7 @@
 import random
 import time
 from pyln.client import Plugin, Millisatoshi, RpcError
+from threading import Lock
 
 
 plugin = Plugin()
@@ -11,6 +12,9 @@ plugin.adj_balances = {}
 plugin.our_node_id = None
 # Users can configure this
 plugin.update_threshold = 0.05
+# forward_event must wait for init
+plugin.mutex = Lock()
+plugin.mutex.acquire()
 
 
 def get_adjusted_percentage(plugin: Plugin, scid: str):
@@ -157,6 +161,7 @@ def maybe_add_new_balances(plugin: Plugin, scids: list):
 
 @plugin.subscribe("forward_event")
 def forward_event(plugin: Plugin, forward_event: dict, **kwargs):
+    plugin.mutex.acquire(blocking=True)
     if forward_event["status"] == "settled":
         in_scid = forward_event["in_channel"]
         out_scid = forward_event["out_channel"]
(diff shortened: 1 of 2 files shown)


feeadjuster/feeadjuster.py
```diff
@@ -11,6 +11,25 @@
 plugin.our_node_id = None
 # Users can configure this
 plugin.update_threshold = 0.05
+# With too low liquidity, there is no point in using this plugin
+plugin.node_liquidity_threshold = 0.2
```
m-schmoock, line 15: Maybe we want this to be configured

feeadjuster/feeadjuster.py
```diff
@@ -204,10 +267,23 @@ def init(options: dict, configuration: dict, plugin: Plugin, **kwargs):
...
+    "Absolute channel balance update threshold (in msat) at which to trigger an update. "
+    "Note: it's also fuzzed by 1.5%",
+    "string"
+)
+plugin.add_option(
+    "feeadjuster-enough-liquidity",
+    "2000000000",
+    "Beyond this liquidity threshold (in msat) there is no need to adjusting fees",
+    "string"
+)
```
m-schmoock, line 286: 1. Can this also be an msat string. i.e.  `0.02btc` ? This would be better for most users.
2. 2000000000 is ~400USD which is already quite big. But its hard to get a meaningful default on magic numbers that are prone to external factors (network state, fiat price, ...)

m-schmoock (changes requested):
Thanks :D
Not quite done with a complete review yet, but this is my first round of thoughts.

feeadjuster/feeadjuster.py
```diff
@@ -51,7 +68,28 @@ def maybe_setchannelfee(plugin: Plugin, scid: str, base: int, ppm: int):
...
+    if last_liquidity is None:
+        return True
+    # Only update on substantial balance moves to avoid flooding, and add
+    # some pseudo-randomness to avoid too easy channel balance probing
+    update_threshold = plugin.update_threshold
+    update_threshold_abs = int(plugin.update_threshold_abs)
+    if not plugin.deactivate_fuzz:
+        r = random.uniform(-0.015, 0.015)
+        update_threshold += r
+        update_threshold_abs += update_threshold_abs * r
```
m-schmoock, line 87: Suggestion: `r` should be taken twice: once for fuzzing `update_threshold` and another one for `update_threshold_abs`. In this case I would skip the local asignment of `r` altogether and simply do: `update_threshold += random.uniform(-0.015, 0.015)` and `update_threshold_abs += random.uniform(-0.015, 0.015)`

feeadjuster/feeadjuster.py
```diff
@@ -204,10 +233,23 @@ def init(options: dict, configuration: dict, plugin: Plugin, **kwargs):
...
     "0.05",
-    "Channel balance update threshold at which to trigger an update. "
+    "Relative channel balance update threshold at which to trigger an update. "
+    "Note: it's also fuzzed by 1.5%",
+    "string"
+)
+plugin.add_option(
+    "feeadjuster-threshold-abs",
+    "0.001btc",
+    "Absolute channel balance update threshold at which to trigger an update. "
```
m-schmoock, line 243: Reading the code, this can be more precise: 
```
    "Absolute channel balance delta at which to always trigger an update. "
```

feeadjuster/feeadjuster.py
```diff
@@ -204,10 +233,23 @@ def init(options: dict, configuration: dict, plugin: Plugin, **kwargs):
 plugin.add_option(
     "feeadjuster-threshold",
     "0.05",
-    "Channel balance update threshold at which to trigger an update. "
+    "Relative channel balance update threshold at which to trigger an update. "
```
m-schmoock, line 236: Wording:
```
    "Relative channel balance delta at which to trigger an update. Default 0.05 means 5%."
```

feeadjuster/feeadjuster.py
```diff
@@ -13,6 +13,23 @@
...
+    """
+    channel = plugin.adj_balances[scid]
+    min_liquidity = min(channel["total"] / 2, int(plugin.big_enough_liquidity))
+    if channel["our"] < min_liquidity:
+        percentage = channel["our"] / min_liquidity / 2
+    elif channel["total"] - channel["our"] < min_liquidity:
+        percentage = (min_liquidity + channel["our"] - channel["total"]) / min_liquidity / 2 + 0.5
+    else:
+        percentage = 0.5
+    return percentage
```
m-schmoock, line 30: We still need testcases that covers the cornercases in here regarding `big_enough_liquidity`.

m-schmoock (commented):
My two Satoshis...

feeadjuster/feeadjuster.py
gallizoltan: Done.

feeadjuster/test_feeadjuster.py
```diff
@@ -216,3 +216,76 @@ def test_feeadjuster_imbalance(node_factory):
...
+    l2 will adjust its configuration-set base and proportional fees for
+    channels A and B as l1 and l3 exchange payments.
+    """
+    base_fee = 5000
+    ppm_fee = 300
+    l2_opts = {
+        "fee-base": base_fee,
+        "fee-per-satoshi": ppm_fee,
+        "plugin": plugin_path,
+        "feeadjuster-deactivate-fuzz": None,
```
m-schmoock, line 238: Hm, will this test work better with `"feeadjuster-deactivate-fuzz": True` ?

m-schmoock (commented):
Test is failing on CI but not locally, not sure why...

feeadjuster/test_feeadjuster.py
gallizoltan: In theory, it shouldn't matter. The fuzz is very little in this case, and the forwarded amount is far bigger.

In practice, I tried with `"feeadjuster-deactivate-fuzz": True`, but caused errors:
```
lightningd: --feeadjuster-deactivate-fuzz: doesn't allow an argument
```
Maybe it isn't how it works. Maybe `"feeadjuster-deactivate-fuzz": None` means the switch is turned on. (Kind of counterintuitive.) So I tried out what happens if I remove this line. But on my machine it does not matter, the tests are pass with or without this line.

I don't know why CI fails.

gallizoltan: I try to rebase to origin/master to see if it helps.

gallizoltan: It is not the rebase, but I've figured out a way to make CI work.
However, it's a bit frustrating that I cannot reliably test it on my machine because the CI test can differ.

m-schmoock: I think your right on `"feeadjuster-deactivate-fuzz": None` means turning it on :D

feeadjuster/feeadjuster.py
m-schmoock, line 30: Maybe its just me, but I find this function rather hard to grasp (and test).

 - Can the actual `ours > min_liquidity` check which is implied by `if/else return 0.5` be somehow pulled out to line 113 right after the `is not significant_update(): continue` check? Only do this if you also think it makes things better, not worse.
 - If not, can the `return 0.5` be at least an early return?
 - You can add a local variable `theirs` to be `channel["total"] - channel["our"]` and use it to make things more readable.

m-schmoock (commented):
Some more feedback... sorry for the delay

feeadjuster/feeadjuster.py
```diff
@@ -167,6 +197,8 @@ def init(options: dict, configuration: dict, plugin: Plugin, **kwargs):
     plugin.our_node_id = plugin.rpc.getinfo()["id"]
     plugin.deactivate_fuzz = options.get("feeadjuster-deactivate-fuzz", False)
     plugin.update_threshold = float(options.get("feeadjuster-threshold", "0.05"))
+    plugin.update_threshold_abs = Millisatoshi(options.get("feeadjuster-threshold-abs", "0.001btc"))
+    plugin.big_enough_liquidity = Millisatoshi(options.get("feeadjuster-enough-liquidity", "0.02btc"))
```
m-schmoock, line 201: @darosior  Do we want this feature turned on with `0.02btc` per default?

m-schmoock (commented):
A note/question to @darosior

feeadjuster/feeadjuster.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 import random
 import time
-from pyln.client import Plugin, RpcError
+from pyln.client import Plugin, Millisatoshi, RpcError
```
darosior, line 4: Err, i was reluctant to using Millisatoshi, it's always been a source of bugs in my experience... Hopefully m-schmoock is on the patch of trackin'em all down!

feeadjuster/feeadjuster.py
```diff
@@ -183,15 +217,11 @@ def init(options: dict, configuration: dict, plugin: Plugin, **kwargs):
...
-                       int(100 * plugin.imbalance),
-                       int(100 * (1 - plugin.imbalance)),
-                       plugin.update_threshold,
-                       plugin.deactivate_fuzz,
-                       plugin.get_ratio.__name__))
+    plugin.log(f"Plugin feeadjuster initialized ({plugin.adj_basefee} base / {plugin.adj_ppmfee} ppm) with an "
+               f"imbalance of {int(100 * plugin.imbalance)}%/{int(100 * ( 1 - plugin.imbalance))}%, "
+               f"update_threshold: {int(100 * plugin.update_threshold)}%, update_threshold_abs: {plugin.update_threshold_abs}, "
+               f"enough_liquidity: {plugin.big_enough_liquidity}, deactivate_fuzz: {plugin.deactivate_fuzz}, "
+               f"adjustment_method: {plugin.get_ratio.__name__}")
```
darosior, line 224: Much fstring much MSPV 3.6 much wow

feeadjuster/feeadjuster.py
darosior: Thanks for the ping. No: if we decide to include this, i think it should be optional.

darosior (commented):
Sorry for getting to this so late that you already had some review rounds.

I would like to understand the rationale for this change. I perfectly get the need for a complementary absolute threshold, but why we should stop adjusting fees above a certain liquidity threshold ?

feeadjuster/feeadjuster.py
gallizoltan: I've never heard about fstring before you told me about it. I love it! :)

feeadjuster/feeadjuster.py
```diff
@@ -204,10 +237,23 @@ def init(options: dict, configuration: dict, plugin: Plugin, **kwargs):
...
+    "feeadjuster-threshold-abs",
+    "0.001btc",
+    "Absolute channel balance delta at which to always trigger an update. "
     "Note: it's also fuzzed by 1.5%",
     "string"
 )
+plugin.add_option(
+    "feeadjuster-enough-liquidity",
+    "0msat",
+    "Beyond this liquidity threshold there is no need to adjusting fees, 0 means the feature is turned off",
```
m-schmoock, line 254: You could point out that this also that this changes the adjustment curve to achieve this.
```suggestion
    "Beyond this liquidity do not adjust fees. "
    "This also modifies the fee curve to achieve having this amount of liquidity. "
    "Default: '0msat' (turned off).",
```

m-schmoock (commented):
Its in good shape.
Now I think I also understand why you do the `get_adjusted_percentage` in this feature rather than not touch fees when there's enough liquidity ;). This wasn't initially clear to me. I thought of it as a `imbalance_abs` pendant to the imbalance (percentage) feature. Correct me if I'm wrong, but you adjust the fee curve so that this amount of liquidity is tried to achieve by setting the fees.

Also I think I found why the `test_feeadjuster_big_enough_liquidity` fails on CI, see review.
Other than that just remaining nits. One more forcepush and we are good to go ;)

darosior:
> I only would like to use feeadjuster, if my liquidity is too low/too high.

Then why not using the imbalance feature?

gallizoltan:
> > I only would like to use feeadjuster, if my liquidity is too low/too high.
> 
> Then why not using the imbalance feature?

@darosior I drew a graph for you to show what I was thinking about feeadjuster.
This is what feeadjuster does with fees (without imbalance and big liquidity):
https://zoltan.xyz/feeadjuster.png
And this happens, when imbalance gets activated:
https://zoltan.xyz/feeadjuster-imbalance.png
I don't like this graph because of the two sudden jumps.
So, instead of imbalance, I'd rather use big liquidity:
https://zoltan.xyz/feeadjuster-big-liquidity.png
No sudden jumps here, the feature nicely gets deactivated.

Beyond that, imbalance works the same way with small and big channels too. However, big liquidity works only with big channels. In the case of small channels (smaller then `big_enough_liquidity * 2`), it does nothing.

darosior:
Thanks for the graphs! Will take another look soon :)

feeadjuster/feeadjuster.py
```diff
@@ -200,13 +202,14 @@ def init(options: dict, configuration: dict, plugin: Plugin, **kwargs):
...
-    plugin.get_ratio = get_ratio
-    if options.get("feeadjuster-adjustment-method", "default") == "soft":
-        plugin.get_ratio = get_ratio_soft
-    if options.get("feeadjuster-adjustment-method", "default") == "hard":
-        plugin.get_ratio = get_ratio_hard
+    adjustment_switch = {
+        "soft": get_ratio_soft,
+        "hard": get_ratio_hard,
+        "default": get_ratio
+        }
```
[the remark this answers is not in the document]
gallizoltan: How do you set up your flake8? I check this too, but flake8 only says some "E501 line too long" for me.

m-schmoock:
@gallizoltan 

I found out why CI is unhappy: Loading the plugin produces a `**BROKEN**` message I found in the CI logs:

```
2020-12-15T18:15:30.985Z **BROKEN** plugin-feeadjuster.py: Adjusting fees: 'Plugin' object has no attribute 'imbalance'
```

Python does not like us accessing the on-the-fly `Plugin` instance members e.g. `imbalance` (but also others). For simplicity this style okay and also used in e.g. `summary` plugin. However, when the event hooks (e.g. `forward_event`) are called before `init` this will of course happen. The reason why this is not locally reproduce-able is that the CI is terribly slow when it comes to parallel processing (starting the plugin process...). So we need to make sure that the `forward_event` hook waits until `init` was called ;)

m-schmoock:
@darosior @gallizoltan 
When the above mentioned CI bug is fixed this PR is fine for me. But I would give @darosior the final word as the plugin was commited by him initially. Shaping the fee curves makes sense to me, and to be honest we should also do this for the relative feature (`imbalance`) as well. 

The downside of this is the slightly more complex ratio adjustment function...

Ideally we can make the relative feature `imbalance` and the absolute feature `enough_liquidity` streamline so that they look same / use same code blocks etc. In this case I would go for `imabalance` and `imbalance_abs`... but this can be done in a cleanup PR. Would cause rename of the property config ...

m-schmoock:
@gallizoltan
We can't go for blocking stuff in event hooks, those are time critical.
Just early `return` the hook if a boolean e.g. `plugin.initialized is False` and set it to `True` on init without any mutex.
We don't loose anything critical if we miss an HTLC, right?
If this leads to flaky tests, this must be addressed by modifying the test.

m-schmoock:
@gallizoltan 
Update: My last message is incorrect. It's an event notification, which is not time-critical.
So using the Mutex is simply overkill (imho) ;)

m-schmoock (approved):

darosior:
Sorry for being so late to come back to this.. If i have nits i can still PR them anyways :)

PR #169, opened 2020-11-30
Backup: restore without filename

m-schmoock:
- fix readme to restore to a filename not just a dir
- adds the option to restore to just a dir, in this case we assume `lightningd.sqlite3` as filename.
- fixes flake8 issues


cdecker:
ACK 256721f

PR #170, opened 2020-12-01
Deprecate/remove outdated sendinvoiceless plugin

cdecker:
Keysend is more secure, has a higher success-rate and doesn't require
both a a route to the destination and back.


gallizoltan (approved):
ACK

PR #175, opened 2020-12-07
automatic rebalance

gallizoltan:
- rebalances liquidity between channels through circular payments
- automatically selects appropriate channels and executes rebalance
- uses economic fee by default: the rebalance transaction fees are less than the node's default fee, so it is cheaper to fix the liquidity than it can be ruined by transaction forwards
- runs async in the background, sometimes for a very long time (maybe hours)
- tries to rebalance small channels to a 50/50 liquidity ratio and stores remaining liquidity in big channels
- does nothing if there is no cheap enough circular route for rebalance

### DIFF
--- a/rebalance/README.md
+++ b/rebalance/README.md
@@ -1,6 +1,6 @@
 # Rebalance plugin
 
-This plugin moves liquidity between your channels using circular payments:
+This plugin moves liquidity between your channels using circular payments
 
 
 ## Installation
@@ -11,8 +11,41 @@ For general plugin installation instructions see the repos main
 
 ## Usage
 
-Once the plugin is installed and active, you can use the `lightning-cli` to
-rebalance channels like this:
+Once the plugin is installed and active, there are two additional methods to rebalance channels:
+1) Either you can call `lightning-cli rebalanceall` to automatically fix all of your channels' liquidity.
+2) Or you can call `lightning-cli rebalance outgoing_scid incoming_scid` to rebalance individual channels.
+
+## Automatic rebalance
+
+A lightning node usually has multiple channels of different sizes. The node can perform best if all channels have `{enough_liquidity}` for both directions. So the rebalance has multiple purposes with different priority:
+1) **The primary goal** is to ensure all channels have `{enough_liquidity}` for both direction, or if a given channel is too small for that, then it has a 50/50 liquidity ratio.
+2) **The secondary goal** is to distribute the remaining liquidity evenly between the big channels.
+3) For the long run, it is very important **to do this economically**. So the fees of fixing liquidity have to be cheaper than the fees of transaction forwards, which can ruin the liquidity again. (This assumes your node has some rational fee setting.) This way the automatic rebalance can run regularly, and your node can earn more on transaction forwarding than spend for rebalancing.
+
+If the plugin cannot find a cheap enough circular route to rebalancing economically, then it does nothing by default. To not to cause a loss for users.
+
+#### Rebalancing strategy
+
+As a first step, depending on the actual situation, there is a need to get a value of `{enough_liquidity}`. The plugin searches for a maximum possible threshold. For which all channels theoretically can be balanced beyond this threshold. Or smaller than `threshold * 2` channels can be balanced to a 50/50 ratio. `{enough_liquidity}` will be half of this maximum threshold.
+
+The next step is to calculate `{ideal_ratio}` for big channels. Beyond the `{enough_liquidity}` threshold, big channels should share the remaining liquidity evenly, so every big channels' liquidity ratio should be close to the `{ideal_ratio}`.
+
+After we know the current `{enough_liquidity}` threshold and `{ideal_ratio}`, the plugin checks every possible channel pairs to seek a proper rebalance opportunity. If it finds a matching pair, it calls the individual rebalance method for them. If the rebalance fails, the plugin tries again with a lesser amount, until it reaches the minimum rebalancable amount, or the rebalance succeeds.
+
+This process may take a while. Automatic rebalance can run for hours in the background.
+
+#### Parameters for automatic rebalance
+
+- OPTIONAL: The `min_amount` parameter sets the minimum rebalancable amount in millisatoshis. The parameter also can be specified in other denominations by appending a valid suffix, i. e. '1000000sat', '0.01btc' or '10mbtc'. The default value is '50000sat'.
+- OPTIONAL: The `feeratio` sets how much the rebalance may cost as a ratio of your default fee. Its default value is `0.5`, which means it can use a maximum of half of your node's default fee.
+
+#### Tips and Tricks for automatic rebalance
+
+- It may work only with well-connected nodes. You should have several different channels to use it with a good chance for success.
+- Your node should have some rational default fee setting. If you use cheaper fees than your neighbors, it probably cannot find a cheap enough circular route to rebalance.
+
+## Individual channel rebalance
+You can use the `lightning-cli` to rebalance channels like this:
 
 ```
 lightning-cli rebalance outgoing_scid incoming_scid [msatoshi] [maxfeepercent] [retry_for] [exemptfee]
@@ -25,7 +58,7 @@ use always the `lightning-cli -k` (key=value) syntax like this:
 lightning-cli rebalance -k outgoing_scid=1514942x51x0 incoming_scid=1515133x10x0 maxfeepercent=1
 ```
 
-### Parameters
+#### Parameters for individual rebalance
 
 - The `outgoing_scid` is the short_channel_id of the sending channel,
 - The `incoming_scid` is the short_channel_id of the receiving channel.
@@ -43,7 +76,7 @@ lightning-cli rebalance -k outgoing_scid=1514942x51x0 incoming_scid=1515133x10x0
   exemptfee (default: 5000 millisatoshi).
 
 
-## Tips and Tricks
+#### Tips and Tricks for individual rebalance
 
 - To find the correct channel IDs, you can use the `summary` plugin which can
   be found [here](https://github.com/lightningd/plugins/tree/master/summary).
--- a/rebalance/README.md
+++ b/rebalance/README.md
@@ -32,7 +32,7 @@ The next step is to calculate `{ideal_ratio}` for big channels. Beyond the `{eno
 
 After we know the current `{enough_liquidity}` threshold and `{ideal_ratio}`, the plugin checks every possible channel pairs to seek a proper rebalance opportunity. If it finds a matching pair, it calls the individual rebalance method for them. If the rebalance fails, the plugin tries again with a lesser amount, until it reaches the minimum rebalancable amount, or the rebalance succeeds.
 
-This process may take a while. Automatic rebalance can run for hours in the background.
+This process may take a while. Automatic rebalance can run for hours in the background, but you can stop it anytime with `lightning-cli rebalancestop`.
 
 #### Parameters for automatic rebalance
 
--- a/rebalance/README.md
+++ b/rebalance/README.md
@@ -34,7 +34,7 @@ After we know the current `{enough_liquidity}` threshold and `{ideal_ratio}`, th
 
 This process may take a while. Automatic rebalance can run for hours in the background, but you can stop it anytime with `lightning-cli rebalancestop`.
 
-#### Parameters for automatic rebalance
+#### Parameters for rebalanceall
 
 - OPTIONAL: The `min_amount` parameter sets the minimum rebalancable amount in millisatoshis. The parameter also can be specified in other denominations by appending a valid suffix, i. e. '1000000sat', '0.01btc' or '10mbtc'. The default value is '50000sat'.
 - OPTIONAL: The `feeratio` sets how much the rebalance may cost as a ratio of your default fee. Its default value is `0.5`, which means it can use a maximum of half of your node's default fee.
@@ -58,7 +58,7 @@ use always the `lightning-cli -k` (key=value) syntax like this:
 lightning-cli rebalance -k outgoing_scid=1514942x51x0 incoming_scid=1515133x10x0 maxfeepercent=1
 ```
 
-#### Parameters for individual rebalance
+#### Parameters for rebalance
 
 - The `outgoing_scid` is the short_channel_id of the sending channel,
 - The `incoming_scid` is the short_channel_id of the receiving channel.
--- a/rebalance/rebalance.py
+++ b/rebalance/rebalance.py
@@ -220,50 +220,39 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi=None,
     return cleanup(plugin, label, payload, success_msg)
 
 
-def must_send(channel, enough_liquidity: int):
+def a_minus_b(a: Millisatoshi, b: Millisatoshi):
+    # a minus b, but Millisatoshi cannot be negative
+    return a - b if a > b else Millisatoshi(0)
+
+
+def must_send(liquidity):
     # liquidity is too high, must send some sats
-    min_liquidity = min(channel["msatoshi_total"] / 2, enough_liquidity)
-    their_liquidity = channel["msatoshi_total"] - channel["msatoshi_to_us"]
-    return max(0, min_liquidity - their_liquidity)
+    return a_minus_b(liquidity["min"], liquidity["their"])
 
 
-def should_send(channel, enough_liquidity: int, ideal_ratio: float):
+def should_send(liquidity):
     # liquidity is a bit high, would be good to send some sats
-    min_liquidity = min(channel["msatoshi_total"] / 2, enough_liquidity)
-    their_liquidity = channel["msatoshi_total"] - channel["msatoshi_to_us"]
-    their_ideal_liquidity = channel["msatoshi_total"] * (1 - ideal_ratio)
-    should_have = min(max(their_ideal_liquidity, min_liquidity), channel["msatoshi_total"] - min_liquidity)
-    return max(0, should_have - their_liquidity)
+    return a_minus_b(liquidity["ideal"]["their"], liquidity["their"])
 
 
-def could_send(channel, enough_liquidity: int):
+def could_send(liquidity):
     # liquidity maybe a bit low, but can send some more sats, if needed
-    min_liquidity = min(channel["msatoshi_total"] / 2, enough_liquidity)
-    our_liquidity = channel["msatoshi_to_us"]
-    return max(0, our_liquidity - min_liquidity)
+    return a_minus_b(liquidity["our"], liquidity["min"])
 
 
-def must_receive(channel, enough_liquidity: int):
+def must_receive(liquidity):
     # liquidity is too low, must receive some sats
-    min_liquidity = min(channel["msatoshi_total"] / 2, enough_liquidity)
-    our_liquidity = channel["msatoshi_to_us"]
-    return max(0, min_liquidity - our_liquidity)
+    return a_minus_b(liquidity["min"], liquidity["our"])
 
 
-def should_receive(channel, enough_liquidity: int, ideal_ratio: float):
+def should_receive(liquidity):
     # liquidity is a bit low, would be good to receive some sats
-    min_liquidity = min(channel["msatoshi_total"] / 2, enough_liquidity)
-    our_liquidity = channel["msatoshi_to_us"]
-    our_ideal_liquidity = channel["msatoshi_total"] * ideal_ratio
-    should_have = min(max(our_ideal_liquidity, min_liquidity), channel["msatoshi_total"] - min_liquidity)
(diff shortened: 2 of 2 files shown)


rebalance/rebalance.py
```diff
@@ -212,16 +214,260 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi=None,
...
+        plugin.options["feeratio"]["value"] = float(feeratio)
+        plugin.options["min-amount"]["value"] = Millisatoshi(min_amount)
+        t = Thread(target = rebalanceall_thread, args = (plugin, ))
+        t.start()
+    except Exception as e:
+        plugin.mutex.release()
+        raise e
+    return {"message" : "Rebalance started"}
+
+
```
m-schmoock, line 455: We should also add a `rebalancestop` method, that interrupts the `rebalanceall` thread.

rebalance/rebalance.py
```diff
@@ -212,16 +214,260 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi=None,
...
+        ch2 = get_chan(plugin, scid2)
+        assert ch2 is not None
+
+
+def maybe_rebalance_once(plugin: Plugin, failed_pairs: list):
+    channels = get_our_channels(plugin)
+    for ch1 in channels:
+        for ch2 in channels:
+            result = maybe_rebalance_pairs(plugin, ch1, ch2, failed_pairs)
+            if result["success"] == True:
```
m-schmoock, line 405: Nit: flake8 `E712 comparison to True should be 'if cond is True:' or 'if cond:'`

rebalance/rebalance.py
```diff
@@ -212,16 +214,260 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi=None,
...
+                   f"min rebalancable amount: {int(Millisatoshi(plugin.get_option('min-amount'))):,} msat, "
+                   f"feeratio: {plugin.get_option('feeratio')}")
+        plugin.options["enough-liquidity"]["value"] = enough_liquidity
+        plugin.options["ideal-ratio"]["value"] = ideal_ratio
+        failed_pairs = []
+        success = 0
+        fee_spent = 0
+        while True:
+            result = maybe_rebalance_once(plugin, failed_pairs)
+            if result["success"] == False:
```
m-schmoock, line 426: Nit: flake8 `E712 comparison to False should be 'if cond is False:' or 'if not cond:'`

rebalance/rebalance.py
```diff
@@ -212,16 +214,260 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi=None,
...
+
+
+@plugin.method("rebalanceall")
+def rebalanceall(plugin: Plugin, min_amount: Millisatoshi = Millisatoshi("50000sat"), feeratio: float = 0.5):
+    """Rebalance all unbalanced channels if possible for a very low fee.
+    Default minimum rebalancable amount is 50000sat. Default feeratio = 0.5, half of our node's default fee.
+    To be economical, it tries to fix the liquidity cheaper than it can be ruined by transaction forwards.
+    May run for a long time (hours) in the background.
+    """
+    if not plugin.mutex.acquire(blocking = False):
```
m-schmoock, line 443: Nit: flake8 `E251 unexpected spaces around keyword / parameter equals`

rebalance/rebalance.py
```diff
@@ -212,16 +214,260 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi=None,
...
+        return {"message" : "Rebalance is already running, this may take a while"}
```
m-schmoock, line 444: Nit: flake8 `E203 whitespace before ':'`

rebalance/rebalance.py
```diff
@@ -212,16 +214,260 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi=None,
...
+    try:
+        plugin.options["feeratio"]["value"] = float(feeratio)
+        plugin.options["min-amount"]["value"] = Millisatoshi(min_amount)
+        t = Thread(target = rebalanceall_thread, args = (plugin, ))
```
m-schmoock, line 448: Nit: flake8 `E251 unexpected spaces around keyword / parameter equals`

rebalance/rebalance.py
```diff
@@ -212,16 +214,260 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi=None,
...
+        t.start()
+    except Exception as e:
+        plugin.mutex.release()
+        raise e
+    return {"message" : "Rebalance started"}
```
m-schmoock, line 453: Nit: flake8 `E203 whitespace before ':'`

m-schmoock (commented):
Nice!
This is a good addition that does not change the basic function or code (besides reducing log, which is good too). Testcasing this will be a pain in the butt, but since we have no meaningful rebalance tests yet, we can start with the basic function test (later).

I will try this, when it seems good in action we can merge this

rebalance/rebalance.py
gallizoltan: Thanks for letting me know and sorry for the many nits in my code. I've just installed flake8 on my machine, and I'm planning to run it before every commit.

m-schmoock: No worries. This should (and will) be done by CI ;)

rebalance/rebalance.py
```diff
@@ -470,4 +483,5 @@ def init(options, configuration, plugin):
 plugin.add_option("enough-liquidity", 0, "Above this threshold (in msat), a channel's liquidity is big enough. Will be calculated.", "int")
 plugin.add_option("ideal-ratio", "0.5", "Ideal liquidity ratio for big channels. Will be calculated.", "string")
 plugin.add_option("min-amount", "50000sat", "The minimum rebalancable amount.", "string")
+plugin.add_option("rebalance-stop", False, "It stops the ongoing rebalanceall, set by rebalancestop", "flag")
```
m-schmoock, line 486: Whats the benefit of adding `rebalance-stop` as a documented plugin option? It shows up in `lightning-cli listconfigs` and is not intended for the user to see or access. Why not add it as a member to plugin in line 8?: `plugin.rebalance_stop = False`

rebalance/rebalance.py
```diff
@@ -438,19 +438,31 @@ def rebalanceall(plugin: Plugin, min_amount: Millisatoshi = Millisatoshi("50000s
     """Rebalance all unbalanced channels if possible for a very low fee.
     Default minimum rebalancable amount is 50000sat. Default feeratio = 0.5, half of our node's default fee.
     To be economical, it tries to fix the liquidity cheaper than it can be ruined by transaction forwards.
-    May run for a long time (hours) in the background.
+    It may run for a long time (hours) in the background, but can be stopped with the rebalancestop method.
     """
-    if not plugin.mutex.acquire(blocking = False):
-        return {"message" : "Rebalance is already running, this may take a while"}
+    if not plugin.mutex.acquire(blocking=False):
+        return {"message": "Rebalance is already running, this may take a while"}
```
m-schmoock, line 444:
```suggestion
return {"message": "Rebalance is already running, this may take a while. To stop it use the cli method 'rebalancestop'."}
```

m-schmoock (commented):
Neat! Some remarks...

rebalance/rebalance.py
gallizoltan: You're right. By the way, I think the same is true for `cltv-final`, `fee-base`, `fee-per-satoshi`, `enough-liquidity`, and `ideal-ratio`.

`feeratio` and `min-amount` aren't options also, but they are parameters for `rebalanceall`.

I think there should be no option here at all.

rebalance/rebalance.py
```diff
@@ -212,16 +215,260 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi=None,
...
+
+
+@plugin.method("rebalanceall")
+def rebalanceall(plugin: Plugin, min_amount: Millisatoshi = Millisatoshi("50000sat"), feeratio: float = 0.5):
+    """Rebalance all unbalanced channels if possible for a very low fee.
+    Default minimum rebalancable amount is 50000sat. Default feeratio = 0.5, half of our node's default fee.
+    To be economical, it tries to fix the liquidity cheaper than it can be ruined by transaction forwards.
+    It may run for a long time (hours) in the background, but can be stopped with the rebalancestop method.
+    """
+    if not plugin.mutex.acquire(blocking=False):
```
m-schmoock, line 440: This should be done by a `locked()` check, as the `acquire`/`release` should be done within the Thread run method itself and the thread start helper function.
```
    if plugin.mutex.locked():
```

rebalance/rebalance.py
```diff
@@ -212,16 +215,260 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi=None,
...
+        return {"message": "Rebalance is already running, this may take a while. To stop it use the cli method 'rebalancestop'."}
+    try:
+        plugin.feeratio = float(feeratio)
+        plugin.min_amount = Millisatoshi(min_amount)
+        t = Thread(target=rebalanceall_thread, args=(plugin, ))
+        t.start()
+    except Exception as e:
+        plugin.mutex.release()
```
m-schmoock, line 448: When the `mutex` is `acquire` within the thread run function, this `release` does not need to be here... but in the threads generic exception/finally handler (where it already is :)

rebalance/rebalance.py
```diff
@@ -212,16 +215,260 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi=None,
...
+        for ch2 in channels:
+            result = maybe_rebalance_pairs(plugin, ch1, ch2, failed_pairs)
+            if result["success"] or plugin.rebalance_stop:
+                return result
+    return {"success": False, "fee_spent": 0}
+
+
+def rebalanceall_thread(plugin: Plugin):
+    try:
+        channels = get_our_channels(plugin)
```
m-schmoock, line 411: First line here should actually do the `if not plugin.mutex.acquire(blocking=False): ...` acquire lock check, so we have the locks ownership on the correct thread and not the starter function.

rebalance/rebalance.py
```diff
@@ -212,16 +215,260 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi=None,
...
+        plugin.log("The feeadjuster plugin would be useful here")
+
+
+def get_max_amount(i: int, min_amount: int, enough_liquidity: int):
+    return max(min_amount, int(enough_liquidity / (4**(i + 1))))
+
+
+def get_max_fee(plugin: Plugin, msat: int):
+    # TODO: sanity check
+    return int((plugin.basefee + plugin.ppmfee * msat / 1000000) * plugin.feeratio)
```
m-schmoock, line 340: If possible, use `Millisatoshi` and get rid of the `int()` casts below
```suggestion
def get_max_fee(plugin: Plugin, msat: Millisatoshi):
    # TODO: sanity check
    return (plugin.basefee + msat * plugin.ppmfee / 1000000) * plugin.feeratio
```

rebalance/rebalance.py
```diff
@@ -174,7 +175,7 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi=None,
         for channel in mychannels:
             excludes += [channel['short_channel_id'] + '/0', channel['short_channel_id'] + '/1']
 
-        while int(time.time()) - start_ts < retry_for:
+        while int(time.time()) - start_ts < retry_for and not plugin.rebalance_stop:
```
m-schmoock, line 178: Great idea.
I tried to stop this also for a normal `rebalance` operation (1min default timeout). This does not work to stop it, not sure why, maybe because it does not detach and the `rebalancestop` operation is actually not starting before this finishes.
Can we make this work also for the normal 'rebalance` call?

m-schmoock (commented):
Great. Nearly there :D
Some issues with thread handling and msat/int casts...

rebalance/rebalance.py
```diff
@@ -361,29 +359,30 @@ def maybe_rebalance_pairs(plugin: Plugin, ch1, ch2, failed_pairs: list):
     result = {"success": False, "fee_spent": 0}
     if scid1 + ":" + scid2 in failed_pairs:
         return result
-    enough_liquidity = int(plugin.get_option("enough-liquidity"))
-    ideal_ratio = float(plugin.get_option("ideal-ratio"))
-    min_amount = int(Millisatoshi(plugin.get_option("min-amount")))
+    min_amount = int(plugin.min_amount)
```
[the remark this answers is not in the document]
gallizoltan: I've found a bug in `Millisatoshi` two years ago. I think it get fixed then, but I did not dare to use `Millisatoshi` again since.
I will try now.

rebalance/rebalance.py
```diff
@@ -212,16 +215,260 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi=None,
...
+    return {"message": "Rebalance stopped"}
+
+
 @plugin.init()
 def init(options, configuration, plugin):
-    plugin.options['cltv-final']['value'] = plugin.rpc.listconfigs().get('cltv-final')
+    config = plugin.rpc.listconfigs()
+    plugin.cltv_final = config.get("cltv-final")
+    plugin.basefee = config.get("fee-base")
+    plugin.ppmfee = config.get("fee-per-satoshi")
```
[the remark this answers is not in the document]
gallizoltan: `fee-base` defaults to 1000 and `fee-per-satoshi` defaults to 10:
https://github.com/ElementsProject/lightning/blob/master/doc/lightningd-config.5.md
I think `cltv-final` has a default value too, but not documented here.

They can be overridden in the config file or as a command-line option when `lightningd` starts.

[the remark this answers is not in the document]
m-schmoock: Ahh sorry, didn't realize those were daemon options not plugin custom options. sure. Ingore that stupid comment of mine. You could just rename the local var to `plugin.fee_ppm`, I find it reads better.

rebalance/rebalance.py
m-schmoock: Imho in theory we can have a CLI blocking `rebalance` call and still interrupt this with `rebalancestop`, but I'm not sure if we need changes to `pyln` or even clnd. JSON-RPC is capable of having messages detached from one another. But we maybe need a CLI RPC dispatcher or something so one thread reads the RPC sock ans `fork`'s new requests or notifies them on result.

gallizoltan: Yep, synchronous plugin methods have a blocking mutex by default, I think. `rebalancestop` will not actually start until normal `rebalance` finishes.

It may a good idea to make `rebalance` non-blocking, but I think it would be too much for now: I'd rather do it later in a different PR. It raises some questions also: if `rebalance` is non-blocking, then we cannot easily see when it finishes and what is its return value.

m-schmoock: Yep, your right.

rebalance/rebalance.py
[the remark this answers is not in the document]
gallizoltan: Ok, I can use a little more `Millisatoshi`, but I cannot use it everywhere. For example, this raises an exception:
```
    our = Millisatoshi(5)
    total = Millisatoshi(10)
    ratio = our/total
```
This too: `min(10 / 2, Millisatoshi(10))`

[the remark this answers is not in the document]
m-schmoock: We need to fix that in the main repo. The Millisatoshi implementation got messed up in the last release. I already fixed two issues with it. Work around where you have to for now -.-
This is really annoying.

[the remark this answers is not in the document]
m-schmoock: I found this from 2019 where I already addressed those issues. Dont know why it got broken... maybe a missing test

https://github.com/ElementsProject/lightning/commit/f99c461fedaab78a74f96928188683987b3665c9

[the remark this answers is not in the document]
gallizoltan: I can show some other example: `x = min(10, Millisatoshi(10))` and `x = Millisatoshi(5 / 2)`.
But this works: `x = Millisatoshi(5) / 2`.

Ok, I can work around for now, but will be ugly.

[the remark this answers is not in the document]
m-schmoock: Nope, if code gets uglier then no please :D
I thought this would be better...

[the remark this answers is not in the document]
m-schmoock: @gallizoltan see this PR https://github.com/ElementsProject/lightning/pull/4273, not sure though if multiplication and division with two Millisatoshis does make sense. Especially `Millisatoshi(10)/Millisatoshi(5) := 2msat` feels a bit wrong :D

rebalance/rebalance.py
```diff
@@ -212,16 +215,260 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi=None,
...
-        plugin.log("Exception: " + str(e))
+        plugin.log("Exception: " + str(e), 'debug')
         return cleanup(plugin, label, payload, success_msg, e)
     return cleanup(plugin, label, payload, success_msg)
 
 
+def must_send(channel, enough_liquidity: int):
+    # liquidity is too high, must send some sats
+    min_liquidity = min(channel["msatoshi_total"] / 2, enough_liquidity)
+    their_liquidity = channel["msatoshi_total"] - channel["msatoshi_to_us"]
```
darosior, line 226: nit: this (or similar info) is duplicated a lot in these functions, and they are called by the same caller. Maybe worth passing them all a dict with all the liquidity info ?

```python
liquidity = {
    "our": "",
    "their": "",
    "min": "",
    "ideal": {
         "our": "",
         "their": "",
    },
}
```

rebalance/rebalance.py
```diff
@@ -212,16 +215,260 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi=None,
...
+        plugin.log("The feeadjuster plugin would be useful here")
+
+
+def get_max_amount(i: int, plugin: Plugin):
+    return max(plugin.min_amount, Millisatoshi(plugin.enough_liquidity) / (4**(i + 1)))
+
+
+def get_max_fee(plugin: Plugin, msat: Millisatoshi):
+    # TODO: sanity check
+    return (plugin.fee_base + msat * plugin.fee_ppm / 1000000) * plugin.feeratio
```
darosior, line 340: nit: `10**6` would be more explicit for "millionth"

rebalance/README.md
```diff
@@ -11,8 +11,41 @@ For general plugin installation instructions see the repos main
...
+
+After we know the current `{enough_liquidity}` threshold and `{ideal_ratio}`, the plugin checks every possible channel pairs to seek a proper rebalance opportunity. If it finds a matching pair, it calls the individual rebalance method for them. If the rebalance fails, the plugin tries again with a lesser amount, until it reaches the minimum rebalancable amount, or the rebalance succeeds.
+
+This process may take a while. Automatic rebalance can run for hours in the background, but you can stop it anytime with `lightning-cli rebalancestop`.
+
+#### Parameters for automatic rebalance
+
+- OPTIONAL: The `min_amount` parameter sets the minimum rebalancable amount in millisatoshis. The parameter also can be specified in other denominations by appending a valid suffix, i. e. '1000000sat', '0.01btc' or '10mbtc'. The default value is '50000sat'.
+- OPTIONAL: The `feeratio` sets how much the rebalance may cost as a ratio of your default fee. Its default value is `0.5`, which means it can use a maximum of half of your node's default fee.
```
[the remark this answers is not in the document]
darosior: Ok now i read the code it's actually parameters to the `rebalanceall` method. The section could be called `Parameters for `rebalanceall`" or something to convey it's not automatically launched ? I was confused and i actually thought by reading the doc the plugin would now automatically rebalance in the background: it's not, you must call `rebalanceall`.

darosior (commented):
Untested-LGTM. Feel free to ignore the nits. Strongly agree with m-schmoock that some testing would be nice

rebalance/README.md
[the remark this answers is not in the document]
m-schmoock: > you must call `rebalanceall`

Which is good imho ;)

rebalance/rebalance.py
m-schmoock: Also, as cdecker points you, we should use the `..._msat` fields as they are auto intialized to Millisatoshi

gallizoltan: Yep, I'm planning to do that. But it will be a big change, takes some time.

m-schmoock:
ACK 566dd58

Excellent! Works for me.
Lets do the testing seperately

rebalance/README.md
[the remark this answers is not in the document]
darosior: Absolutely :)

PR #177, opened 2020-12-10
Fix example test

gijswijs:
This is the tiniest of PRs, but I think it helps to provide an example test that actually passes.


cdecker:
ACK https://github.com/lightningd/plugins/pull/177/commits/6e01aba3435ede0bea2133832f350d1ba989a7b9

PR #180, opened 2020-12-13
prometheus: small fixups

darosior:
Correct comment indentation and don't mix up scid with alias as per https://github.com/lightningd/plugins/pull/152#discussion_r512623918


cdecker (approved):
ACK 1f205053d0fdbf3f364a3df534a19f74bb81c442

PR #181, opened 2020-12-15
currencyrate: new plugin.

rustyrussell:
I'm enabling offers in alternate currencies using this API, so here's an implementation.

### DIFF
--- a/currencyrate/currencyrate.py
+++ b/currencyrate/currencyrate.py
@@ -3,9 +3,10 @@ from pyln.client import Plugin
 from collections import namedtuple
 from pyln.client import Millisatoshi
 from cachetools import cached, TTLCache
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util.retry import Retry
 import requests
 import statistics
-import time
 
 plugin = Plugin()
 
@@ -41,24 +42,40 @@ sources = [
 ]
 
 
-def get_currencyrate(plugin, currency, req_template, response_members):
+# Stolen from https://www.peterbe.com/plog/best-practice-with-retries-with-requests
+def requests_retry_session(
+    retries=3,
+    backoff_factor=0.3,
+    status_forcelist=(500, 502, 504),
+    session=None,
+):
+    session = session or requests.Session()
+    retry = Retry(
+        total=retries,
+        read=retries,
+        connect=retries,
+        backoff_factor=backoff_factor,
+        status_forcelist=status_forcelist,
+    )
+    adapter = HTTPAdapter(max_retries=retry)
+    session.mount('http://', adapter)
+    session.mount('https://', adapter)
+    return session
+
+
+def get_currencyrate(plugin, currency, urlformat, replymembers):
     # NOTE: Bitstamp has a DNS/Proxy issues that can return 404
     # Workaround: retry up to 5 times with a delay
     currency_lc = currency.lower()
-    url = req_template.format(currency_lc=currency_lc, currency=currency)
-    for _ in range(5):
-        r = requests.get(url, proxies=plugin.proxies)
-        if r.status_code != 200:
-            time.sleep(1)
-            continue
-        break
+    url = urlformat.format(currency_lc=currency_lc, currency=currency)
+    r = requests_retry_session(retries=5, status_forcelist=(404)).get(url, proxies=plugin.proxies)
 
     if r.status_code != 200:
         plugin.log(level='info', message='{}: bad response {}'.format(url, r.status_code))
         return None
 
     json = r.json()
-    for m in response_members:
+    for m in replymembers:
         expanded = m.format(currency_lc=currency_lc, currency=currency)
         if expanded not in json:
             plugin.log(level='debug', message='{}: {} not in {}'.format(url, expanded, json))
@@ -85,7 +102,7 @@ def set_proxies(plugin):
         plugin.proxies = None
 
 
-# Don't grab these more than once per hour.                        
+# Don't grab these more than once per hour.
 @cached(cache=TTLCache(maxsize=1024, ttl=3600))
 def get_rates(plugin, currency):
     rates = {}
@@ -97,8 +114,8 @@ def get_rates(plugin, currency):
     return rates
 
 
-@plugin.method("currencyrate")
-def currencyrate(plugin, currency):
+@plugin.method("currencyrates")
+def currencyrates(plugin, currency):
     """Gets currency from given APIs."""
 
     return get_rates(plugin, currency.upper())
--- a/currencyrate/currencyrate.py
+++ b/currencyrate/currencyrate.py
@@ -104,16 +104,30 @@ def init(options, configuration, plugin):
     set_proxies(plugin)
 
     if options['add-source'] != '':
-        parts = options['add-source'].split(',')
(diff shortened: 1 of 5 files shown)


currencyrate/currencyrate.py
```diff
@@ -0,0 +1,119 @@
...
+    # NOTE: Bitstamp has a DNS/Proxy issues that can return 404
+    # Workaround: retry up to 5 times with a delay
+    currency_lc = currency.lower()
+    url = req_template.format(currency_lc=currency_lc, currency=currency)
+    for _ in range(5):
+        r = requests.get(url, proxies=plugin.proxies)
+        if r.status_code != 200:
+            time.sleep(1)
+            continue
+        break
```
cdecker, line 54: This can be done cleaner with `urllib3.util.Retry`:

```python
    retry = Retry(
        total=retries,
        read=retries,
        connect=retries,
        backoff_factor=backoff_factor,
        status_forcelist=status_forcelist,
    )
    adapter = HTTPAdapter(max_retries=retry)
    session.mount('http://', adapter)
    session.mount('https://', adapter)
```

Taken from https://www.peterbe.com/plog/best-practice-with-retries-with-requests

currencyrate/currencyrate.py
```diff
@@ -0,0 +1,119 @@
...
+           'https://www.bitstamp.net/api/v2/ticker/btc{currency_lc}/',
+           ['last']),
+    # e.g. {"bitcoin":{"usd":17885.84}}
+    Source('coingecko',
+           'https://api.coingecko.com/api/v3/simple/price?ids=bitcoin&vs_currencies={currency_lc}',
+           ['bitcoin', '{currency_lc}']),
+]
+
+
+def get_currencyrate(plugin, currency, req_template, response_members):
```
cdecker, line 44: `response_members` is an alias of `replymembers` in the `Source` class. Could be unified.

currencyrate/currencyrate.py
```diff
@@ -0,0 +1,119 @@
...
+    json = r.json()
+    for m in response_members:
+        expanded = m.format(currency_lc=currency_lc, currency=currency)
+        if expanded not in json:
+            plugin.log(level='debug', message='{}: {} not in {}'.format(url, expanded, json))
+            return None
+        json = json[expanded]
+
+    try:
+        return Millisatoshi(int(10**11 / float(json)))
```
cdecker, line 69: Technically this is not `Millisatoshi` it's `Millisatoshi / currency` which we don't have a type for. We could keep it unitless or introduce a `Rate` class.

currencyrate/requirements.txt
```diff
@@ -0,0 +1,5 @@
+pyln-client>=0.7.3
+requests>=2.10.0
+requests[socks]>=2.10.0
+packaging>=14.1
```
cdecker, line 4: Is `packaging` required?

currencyrate/currencyrate.py
m-schmoock: hm thx :D ... we should go for this in the `summary` plugin price thread as well...

currencyrate/currencyrate.py
```diff
@@ -0,0 +1,148 @@
...
+    for s in sources:
+        r = get_currencyrate(plugin, currency, s.urlformat, s.replymembers)
+        if r is not None:
+            rates[s.name] = r
+
+    return rates
+
+
+@plugin.method("currencyrate")
+def currencyrate(plugin, currency):
```
m-schmoock, line 101: This method's name should be plural `currencyrates` as it returns a set of rates.

currencyrate/currencyrate.py
```diff
@@ -0,0 +1,148 @@
...
+@cached(cache=TTLCache(maxsize=1024, ttl=3600))
+def get_rates(plugin, currency):
+    rates = {}
+    for s in sources:
+        r = get_currencyrate(plugin, currency, s.urlformat, s.replymembers)
+        if r is not None:
+            rates[s.name] = r
+
+    return rates
+
```
m-schmoock, line 98: We could have a manual cache update method/trigger incase someone wants/needs to have fresh rates urgently.

```suggestion
@plugin.method("currencyupdate")
def currencyupdate(plugin)
    ...
```

currencyrate/currencyrate.py
```diff
@@ -0,0 +1,148 @@
...
+
+@plugin.method("currencyconvert")
+def currencyconvert(plugin, amount, currency):
+    """Converts currency using given APIs."""
+    rates = get_rates(plugin, currency.upper())
+    if len(rates) == 0:
+        raise Exception("No values available for currency {}".format(currency.upper()))
+    val = statistics.median([m.millisatoshis for m in rates.values()]) * float(amount)
+    return {"msat": Millisatoshi(round(val))}
```
m-schmoock, line 114: Do we want to notify or warn about if `len(rates)` is less than certain thresholds?
 - `< len(sources)` # less than defined sources
 - `< 3`  # a too small number
 - ... ?

m-schmoock (commented):
Concept Ack

currencyrate/currencyrate.py
m-schmoock: Or maybe have an optional parameter on `currencyconvert` and `currencyrates` that forces a (blocking) cache update.

m-schmoock: Or maybe have an optional parameter on `currencyrates` and `currencyconvert` that sets the caches to e.g. `ttl=300` temporarily just for that call... if something like that is possible.

m-schmoock:
I can imagine we want multi currency support, or at least a currency selection via config.
But we can do that in a follow up...

currencyrate/currencyrate.py
rustyrussell: Will rename to match the Source members, to make it clearer.

currencyrate/currencyrate.py
rustyrussell: No, it's the number of millisatoshi equivalent to 1 currency unit?

currencyrate/requirements.txt
rustyrussell: copy pasta...

currencyrate/currencyrate.py
rustyrussell: <shrug> Probably, but can we add it later when we figure it out?

currencyrate/currencyrate.py
rustyrussell: > I can imagine we want multi currency support, or at least a currency selection via config.
> But we can do that in a follow up...

And we want to allow specification of filters which only apply to certain currencies, as well, rather than this generic format.

currencyrate/currencyrate.py
rustyrussell: I'd rather have some warning when it's markedly different from previous values, TBH (implying some persistence).  Though with Bitcoin's price, it's hard to know what is a bug, and what is a feature!

currencyrate/currencyrate.py
cdecker: Hehe, that's just rephrasing it, I could just as well say it's number of millisatoshi units / currency, which would come out as the inverse of the currency. Currencies aren't unitless amounts.

Anyway, splitting hairs here, we can move on to more important things :-)

PR #184, opened 2020-12-17
backup: Increase version_count after each version

### LINKED ISSUE #183: backup: version_count is never increased
The header of the FileBackend backup file contains a `version_count` metadata field but this is never increased from `0` (It's also never used, so this issue doesn't affect anything in practice).
I suppose it is meant to be increased in `add_change`?

laanwj:
Fix #183. At least I think this was the intent.


cdecker:
Excellent catch, I had totally forgotten about that field. You're correct that it was meant to be incremented on each added change, so we can display the progress when restoring.

ACK https://github.com/lightningd/plugins/pull/184/commits/808e984983292f4843a50f84dd7a38e6d52cb90f

PR #189, opened 2021-01-02
backup: Implement network backup

laanwj:
This PR implements incremental backup over a TCP stream. This adds a `socket:127.0.0.1:1234` backup URI scheme, as well as the `backup-cli server` command.

To make it possible to initialize an empty backend, the lightning directory argument to `backup-cli init` is now optional.  E.g.

```bash
./backup-cli init ~/.lightning/bitcoin file:///mnt/external/location/file.bkp
```
becomes
```bash
./backup-cli init --lightning-dir ~/.lightning/bitcoin file:///mnt/external/location/file.bkp
```

It also does some preparational refactoring, primarily splitting up the monolithic `backup.py` into a file per class—this makes that `bitcoin-cli` no longer needs to depend on `pyln`.

See [remote.md](https://github.com/laanwj/plugins/blob/2021_01_streambackup/backup/remote.md) for design goals, detailed documentation, and description of the network protocol.

I've tested basic functionality like:

- Using the backup plugin with c-lightning, over a ssh tunnel
- Restoring a backup (over the network) and checking it against the original by comparing the `sqlite3 -readonly /tmp/test.sqlite .dump`
- Breaking the connection in various ways to see if `lightningd` shuts down immediately without doing further updates. In the future "wait-and-reconnect" would be a better approach, but this is robust at least.

And have been dogfeeding this myself on one of my nodes.


laanwj:
I'll have a look at why the CI tests are failing.

laanwj:
Backup also survived a crash in `connectd` (possibly ElementsProject/lightning#4222) taking down the process suddenly. Could restart `lightningd` and it carried on where it left off.
```
2021-01-09T23:46:06.458Z DEBUG   hsmd: Client: Received message 1 from client                                                                                                                                                       
2021-01-09T23:46:06.498Z DEBUG   …-connectd: Connect IN                                                                                                            
2021-01-09T23:46:06.498Z DEBUG   …connectd: peer_out WIRE_INIT                                                                                                    
lightningd: connectd failed (signal 6), exiting.                                                                           
```

cdecker:
Thanks @laanwj, sorry for the flaky tests, I'm working on remediating these asap.

Is this ready for merge? I'm more than happy with its current state :+1:

laanwj:
> Is this ready for merge? I'm more than happy with its current state and eager to get it into the hands of more users +1

I'm using it, it works for me. Agree that merging it will likely help getting more testing, and means that other work can be done on top. So if you're happy with the code change otherwise, let's do that 😄

laanwj:
Did a small update here: I removed the partially implemented `backup-cli dump <uri>` command. I think this command is useful (to see what records are in a backup, and for troubleshooting/testing). However it needs some more thinking with regard to presentation. It's also out of scope here.

laanwj:
Rebased, also added support for using the COMPACT command through the socket backend:
```
Received COMPACT       
Starting compaction: stats={'before': {'backupsize': 487140440, 'version_count': 49345}}
Adding intial snapshot with 181899264 bytes for version 1264659
Compacted 49343 changes, saving 305240217 bytes, swapping backups
```
```
user@medea:~ % lcli backup-compact
{
   "before": {
      "backupsize": 487140440,
      "version_count": 49345
   },
   "after": {
      "version_count": 2,
      "backupsize": 181900223
   }
}
```

laanwj:
Did something change in plugin handling?

After upgrading my c-lightning from  ElementsProject/lightning@990623676938adeff66488b43fbd03c1888624ad to ElementsProject/lightning@015ac37d924e31bb87b8597da9f863e82270657b., this seems to hang the daemon at startup.

cdecker:
> After upgrading my c-lightning from [ElementsProject/lightning@9906236](https://github.com/ElementsProject/lightning/commit/990623676938adeff66488b43fbd03c1888624ad) to [ElementsProject/lightning@015ac37](https://github.com/ElementsProject/lightning/commit/015ac37d924e31bb87b8597da9f863e82270657b)., this (passing `--important-plugin $HOME/src/plugins/backup/backup.py `, with socket backend) seems to hang the daemon at startup (and not at some obvious point where it's immediately clear to me what happens, it submits some changes but just never starts accepting commands on RPC).

I'm not aware of any changes in the plugin or RPC interface that might have caused this. Need to look into it myself.

backup/backup-cli
```diff
@@ -1,54 +1,61 @@
...
+handler = logging.StreamHandler(sys.stdout)
+handler.setLevel(logging.DEBUG)
+formatter = logging.Formatter('%(message)s')
+handler.setFormatter(formatter)
+root.addHandler(handler)
 
 @click.command()
-@click.argument("lightning-dir", type=click.Path(exists=True))
 @click.argument("backend-url")
+@click.option('--lightning-dir', type=click.Path(exists=True), default=None, help='Use an existing lightning directory (default: initialize an empty backup).')
```
cdecker, line 24: Making this an option, but then requiring it to be set seems like a bad idea. We could make it default to the default lightning-dir, in which case we could really make this an option.

backup/protocol.py
```diff
@@ -0,0 +1,70 @@
+'''
+Socket-based remote backup protocol. This is used to create a connection to a backup backend, and send5 it incremental database updates.
```
cdecker, line 2: `send5` -> `send`

backup/protocol.py
```diff
@@ -0,0 +1,70 @@
...
+def recv_packet(sock: socket.socket) -> Tuple[int, bytes]:
+    (typ, length) = struct.unpack('!BI', recvall(sock, 5))
+    payload = recvall(sock, length)
+    return (typ, payload)
+
+def change_from_packet(typ, payload):
+    '''Convert a network packet to a Change object.'''
+    if typ == PacketType.CHANGE:
+        (version, ) = struct.unpack('!I', payload[0:4])
+        payload = zlib.decompress(payload[4:])
```
cdecker, line 51: Are all changes zlib compressed? I was considering adding compression as well, and it might make sense to push that down to the file-backend as well, i.e., have a common type structure that allows for compression to be orthogonal to the message type.

Something like using the top nibble to describe the encoding and compression, while using the bottom nibble to be the type Snapshot or Change.

wdyt?

cdecker: I assume that the protocol is fully trusted at this point, so decompression bombs are out of scope :-)

backup/server.py
```diff
@@ -0,0 +1,87 @@
...
+                stats = self.backend.compact()
+                self._send_packet(PacketType.COMPACT_RES, json.dumps(stats).encode())
+            elif typ == PacketType.ACK:
+                logging.info('Received ACK')
+            elif typ == PacketType.NACK:
+                logging.info('Received NACK')
+            elif typ == PacketType.METADATA:
+                logging.info('Received METADATA')
+            elif typ == PacketType.COMPACT_RES:
+                logging.info('Received COMPACT_RES')
```
cdecker, line 72: Aren't these types unexpected to the server as well?

cdecker (commented):
Aside from the non-optional option this looks very good indeed :+1:

backup/protocol.py
laanwj: Yes, all changes and snapshots are compressed. And yes, the protocol is trusted, both from the client and server side, I don't think you can do much else in a backup protocol.
(Authentication is explicitly out of scope)

backup/server.py
laanwj: Yes, true.

backup/backup-cli
laanwj: No, it's not required. The whole idea is that you can leave out the option to make an empty backup (without an initial snapshot). This is useful for initializing the file backend on the server side.

cdecker: Ah I see now. That's clever indeed :+1: I must've missed that. So the first change that the backup plugin must stream is a snapshot then, correct? Is that signalled implicitly via the data version being 0 or do we have to remember this?

backup/protocol.py
cdecker: Sounds good to me :+1: Let's bikeshed in a new issue/PR, and get this merged asap :-)

backup/backup-cli
laanwj: It doesn't make an automatic snapshot. I suppose this could be done, but it's not :slightly_smiling_face:  It's a manual step right now as documented here, while initializing the `socket:` backend: https://github.com/lightningd/plugins/blob/7c8e6dcf0dde42a836dd49dd4f49d4c9b1d9e416/backup/remote.md#usage

cdecker: Ah I see, so this effectively breaks if we chose the `file://` backend, create an empty backup, and the plugin will kill `lightningd` at startup. Not really ergonomic, but I think we can solve it by special-casing `data_version = 0` and taking a snapshot followed by any eventual incremental change we might have.

laanwj: I'm not sure it's that bad. I mean, to get a `backup.lock` installed in the lightning directory you'd normally want to initialize your `socket:` backend (on the client side)? It's no different from initializing a `file://` backend for use with lightingd.

laanwj:
Pushed new version (only change is the typo fix).

backup/server.py
laanwj: Also, the client is [allowed to ACK](https://github.com/lightningd/plugins/blob/62745573b21a2cb165895f507a7236397afd9dff/backup/remote.md#restore) changes from the server when retrieving the backup. It currently doesn't.

I guess a NACK would be fatal though. It should be handled in some way that is not "unexpected message".

I would prefer to leave mailing this down for later.

laanwj:
Anything left to do? I explained the perceived "non-optional option" here: https://github.com/lightningd/plugins/pull/189#discussion_r565366508 .

cdecker:
Rebased on top of the deadlock fix, if it passes CI I'm more than happy to merge :+1:

laanwj:
Thanks!
It seems to pass now :smile:

cdecker:
ACK https://github.com/lightningd/plugins/pull/189/commits/a490c4a2adda6537164ec6ceb9d7794469a7c62a

PR #193, opened 2021-01-04
paytest: A plugin to test multi-part payments and optimize the `pay` plugin

cdecker:
With the introduction of the `pay` plugin, its modifiers, and
especially the MPP support, we have introduced quite a few new
parameters that need optimization. Since these optimizations depend on
the topology of the network and the behavior of the nodes in the
network, we should be testing on the real network. However, to do so
safely, and without putting the tester's funds at risk requires us to
get a bit tricky.

`paytest` is a plugin that is to be deployed both on the sender and
the recipient of the route to be tested. It'll pretend to accumulate
parts of a payment, and time them out correctly on the recipient side,
while it ensures that the sender's funds are safe, by paying an
invoice that the recipient cannot really settle.

### DIFF
--- a/paytest/README.org
+++ b/paytest/README.org
@@ -1,0 +1,60 @@
+#+TITLE: Paytest Plugin
+
+A plugin to benchmark the performance of the ~pay~ plugin. It can
+generate mock invoices for remote nodes, and it can hold on to
+incoming test multi-part payments as if they were real payments.
+
+🚧 This plugin is intended for LN developers. Use it at your own risk 👷
+
+The plugin consists of three parts:
+
+ - ~testinvoice~: An RPC method to generate a fake invoice which
+   forces any payer to go through a specific node as penultimate hop
+   in the route. That node can then decode the onion and pretend it is
+   the real destination. The invoice is valid, however the
+   ~payment_hash~ has no known preimage and can therefore not be
+   settled.
+
+ - An ~htlc_accepted~ hook that intercepts any HTLC resulting from a
+   remote ~testpay~ call, decodes the payload and pretends to be the
+   recipient, exercising the MPP aggregation logic.
+
+ - ~testpay~: A wrapper RPC method that takes generates a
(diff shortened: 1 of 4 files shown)


darosior:
CI failure is
```
Traceback (most recent call last):
  File "/home/runner/work/plugins/plugins/paytest/paytest.py", line 15, in <module>
    from pyln.proto.onion import RoutingOnion, chacha20_stream, ecdh
ImportError: cannot import name 'RoutingOnion'
Traceback (most recent call last):
  File "/home/runner/work/plugins/plugins/paytest/paytest.py", line 15, in <module>
    from pyln.proto.onion import RoutingOnion, chacha20_stream, ecdh
ImportError: cannot import name 'RoutingOnion'
```

cdecker:
> CI failure is
> 
> ```
> Traceback (most recent call last):
>   File "/home/runner/work/plugins/plugins/paytest/paytest.py", line 15, in <module>
>     from pyln.proto.onion import RoutingOnion, chacha20_stream, ecdh
> ImportError: cannot import name 'RoutingOnion'
> Traceback (most recent call last):
>   File "/home/runner/work/plugins/plugins/paytest/paytest.py", line 15, in <module>
>     from pyln.proto.onion import RoutingOnion, chacha20_stream, ecdh
> ImportError: cannot import name 'RoutingOnion'
> ```

Yes, looks like I'm not using pyln 0.9.2 which includes that, need to dig a bit deeper.

jsarenik:
Just noting here these INFO messages about missing docstring:

```
@40000000603506c6215687d8 2021-02-23T13:44:28.559Z DEBUG   plugin-manager: start
ed(1147891) /home/usher/.lightning/plugins/paytest.py
@40000000603506c63871ef98 2021-02-23T13:44:28.946Z INFO    plugin-paytest.py: RP
C method 'testinvoice' does not have a docstring.
@40000000603506c638747fd8 2021-02-23T13:44:28.947Z INFO    plugin-paytest.py: RPC method 'paytest' does not have a docstring.
```

PR #195, opened 2021-01-07
feeadjuster: forward event switch

gallizoltan:
possibility to turn off the forward event with config or plugin method

### DIFF
--- a/feeadjuster/feeadjuster.py
+++ b/feeadjuster/feeadjuster.py
@@ -161,6 +161,8 @@ def maybe_add_new_balances(plugin: Plugin, scids: list):
 
 @plugin.subscribe("forward_event")
 def forward_event(plugin: Plugin, forward_event: dict, **kwargs):
+    if not plugin.forward_event_subscription:
+        return
     plugin.mutex.acquire(blocking=True)
     if forward_event["status"] == "settled":
         in_scid = forward_event["in_channel"]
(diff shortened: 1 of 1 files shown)


feeadjuster/feeadjuster.py
```diff
@@ -204,10 +206,27 @@ def feeadjust(plugin: Plugin):
     return msg
 
 
+@plugin.method("feeadjuster_forward_event_switch")
```
m-schmoock, line 209: I would call the RPC method simply `feeadjuster_toggle` and make the method in a way that when `None` is given it toggles the current state.

m-schmoock (commented):
lgtm other than naming letting the function toggle if no arg is given.

@darosior what do you think?

feeadjuster/feeadjuster.py
```diff
@@ -204,10 +206,27 @@ def feeadjust(plugin: Plugin):
...
+def feeadjuster_forward_event_switch(plugin: Plugin, new_value: bool = None):
+    """Activates/Deactivates forward event subscription.
```
darosior, line 211: Technically this does not deactivate the subscription, but the fee updates

feeadjuster/feeadjuster.py
```diff
@@ -243,6 +262,12 @@ def init(options: dict, configuration: dict, plugin: Plugin, **kwargs):
     "Deactivate update threshold randomization and hysterisis.",
     "flag"
 )
+plugin.add_option(
+    "feeadjuster-deactivate-event-subscription",
+    False,
+    "Deactivate forward_event subscription.",
```
darosior, line 268: What about

> Deactivate new fee updates

This seems more user-friendly to describe the behaviour rather than the technical solution

darosior (commented):
It's trivial so concept ACK, just not sure about the naming

May i ask what was your specific usecase for this ? As it's a notification i believe we could just `plugin stop` the plugin.

m-schmoock:
> As it's a notification i believe we could just `plugin stop` the plugin.

You cant use the `feeadjust` manual trigger RPC method after stopping the plugin.

gallizoltan:
> May i ask what was your specific usecase for this ?

I'd like to turn off the fee updates. At least while I'm running a `rebalanceall`. Or maybe permanently, because after a forward event occurs at my node, I'm going to run a rebalance. And after the rebalance, I call the `feeadjust` method anyway.

darosior (approved):
ACK d982f158c8a9bc23688bd849ad632285b2252af7 -- would be neat if you squashed your fixup commit :)

m-schmoock:
tACK 0d183e8

PR #198, opened 2021-01-15
rebalance: better feeadjuster integration

gallizoltan:
- turns off feeadjuster's automatic fee updates while rebalance running
- runs feeadjust only once, at the end
- measures time for rebalance

### DIFF
--- a/rebalance/rebalance.py
+++ b/rebalance/rebalance.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 from pyln.client import Plugin, Millisatoshi, RpcError
 from threading import Thread, Lock
+from datetime import timedelta
 import time
 import uuid
 
@@ -314,10 +315,11 @@ def get_ideal_ratio(channels: list, enough_liquidity: Millisatoshi):
 
 
 def feeadjust_would_be_nice(plugin: Plugin):
-    try:
+    commands = [c for c in plugin.rpc.help().get("help") if c["command"].split()[0] == "feeadjust"]
+    if len(commands) == 1:
         msg = plugin.rpc.feeadjust()
         plugin.log(f"Feeadjust succeeded: {msg}")
-    except Exception:
+    else:
         plugin.log("The feeadjuster plugin would be useful here")
 
 
@@ -375,6 +377,7 @@ def maybe_rebalance_pairs(plugin: Plugin, ch1, ch2, failed_pairs: list):
         amount = min(amount, get_max_amount(i, plugin))
         maxfee = get_max_fee(plugin, amount)
         plugin.log(f"Try to rebalance: {scid1} -> {scid2}; amount={amount}; maxfee={maxfee}")
+        start_ts = time.time()
         try:
             res = rebalance(plugin, outgoing_scid=scid1, incoming_scid=scid2, msatoshi=amount, maxfeepercent=0, retry_for=1200, exemptfee=maxfee)
         except Exception:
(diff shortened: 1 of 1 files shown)


rebalance/rebalance.py
```diff
@@ -428,8 +442,10 @@ def rebalanceall_thread(plugin: Plugin):
                 break
             success += 1
             fee_spent += result["fee_spent"]
-            feeadjust_would_be_nice(plugin)
-        plugin.log(f"Automatic rebalance finished: {success} successful rebalance, {fee_spent} fee spent")
+        feeadjust_would_be_nice(plugin)
+        feeadjuster_toggle(plugin, feeadjuster_state)
+        elapsed_time = timedelta(seconds=time.time()-start_ts)
```
m-schmoock, line 447: nit: missing spaces
```suggestion
        elapsed_time = timedelta(seconds=time.time() - start_ts)
```

rebalance/rebalance.py
```diff
@@ -408,10 +412,20 @@ def maybe_rebalance_once(plugin: Plugin, failed_pairs: list):
     return {"success": False, "fee_spent": Millisatoshi(0)}
 
 
+def feeadjuster_toggle(plugin: Plugin, new_value: bool):
+    try:
+        msg = plugin.rpc.feeadjuster_toggle(new_value)
+        return msg["forward_event_subscription"]["previous"]
+    except Exception:
```
m-schmoock, line 419: Okay, but is there a better or more precise way to check for a specific RPC method?
`lightning-cli plugin list` is not very good, as it does not know the version of the plugin and we don't have plugin names but just full path names. So maybe we could parse `lightning-cli help` and look for the method name. If found use it. Then we can get rid of the `try/except Exception` and let it fail if the other plugin has a problem.

What do you think? If you change it please also do it for the call to the `feeadjust_would_be_nice` method ;)

m-schmoock (approved):
ACK 5ffa7c0

just some nits

gallizoltan:
I put this to draft until #200 got merged.

rebalance/rebalance.py
```diff
@@ -314,10 +315,11 @@ def get_ideal_ratio(channels: list, enough_liquidity: Millisatoshi):
 
 
 def feeadjust_would_be_nice(plugin: Plugin):
-    try:
+    commands = [c for c in plugin.rpc.help().get("help") if c["command"].split()[0] == "feeadjust"]
```
m-schmoock, line 318: Thx, nice!

We should add this a generic method for this to pyln rpc code, that calls 'help' just the first time and caches it. Then we could have something like this:

```python
if plugin.has_method("somefunction"):
    # foo
```

m-schmoock:
ACK d36e938

m-schmoock:
@gallizoltan 
Next time, but could you try to keep the commit message to a 50/72 format?
Meaning 50chars for the header and 72 chars for the text body part?

m-schmoock:
Just looks a bit clunky ;)
```
* b80c558 (origin/master, origin/HEAD, master) rebalance: better feeadjuster integration - turns off feeadjuster's automatic fee updates while rebalance running - runs feeadjust only once, at the end - measures time for rebalance
* 7c88402 feeadjuster: no underscore in method name
```

gallizoltan:
> Just looks a bit clunky ;)
> 
> ```
> * b80c558 (origin/master, origin/HEAD, master) rebalance: better feeadjuster integration - turns off feeadjuster's automatic fee updates while rebalance running - runs feeadjust only once, at the end - measures time for rebalance
> * 7c88402 feeadjuster: no underscore in method name
> ```

Ok, I think I just hit one Enter in the commit message instead of two. Will do better next time.

rebalance/rebalance.py
m-schmoock: See https://github.com/lightningd/plugins/pull/201

m-schmoock: Update: doesn't seem to work, as rpc code generates method on the fly when getattr is called.

PR #200, opened 2021-01-17
feeadjuster: no underscore in method name

gallizoltan:
cannot call a method from python framework with underscore in its name

https://github.com/ElementsProject/lightning/blob/6f924e63c2e7280b19419d589cca207cfbe6216d/contrib/pyln-client/pyln/client/lightning.py#L316

### DIFF
--- a/feeadjuster/feeadjuster.py
+++ b/feeadjuster/feeadjuster.py
@@ -206,17 +206,17 @@ def feeadjust(plugin: Plugin):
     return msg
 
 
-@plugin.method("feeadjuster_toggle")
-def feeadjuster_toggle(plugin: Plugin, new_value: bool = None):
+@plugin.method("feeadjustertoggle")
+def feeadjustertoggle(plugin: Plugin, value: bool = None):
     """Activates/Deactivates automatic fee updates for forward events.
 
-    The status will be set to new_value.
+    The status will be set to value.
     """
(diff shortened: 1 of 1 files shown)


m-schmoock:
ACK a460bf8

m-schmoock:
@gallizoltan just a follow up:

As you point out
https://github.com/ElementsProject/lightning/blob/6f924e63c2e7280b19419d589cca207cfbe6216d/contrib/pyln-client/pyln/client/lightning.py#L316
pyln RPC method names go with dashes:
```
        name = name.replace('_', '-')
```

Questions:
 - Calling it with `plugin.rpc.feeadjuster-toggle()` does not work for obvious reasons, right?
 - Would it be possible to call it wired like:   `plugin.rpc["feeadjuster-toggle"]()`  ?
 - Do you believe the above pyln code `name = name.replace('_', '-')` is correct or should it be the other way around?

m-schmoock:
@gallizoltan

A follow up: After a small chat with cdecker we should have renamed `feeadjuster_toggle` to `feeadjuster-toggle` then the RPC code would be able to find and call it by using an underscore. Because the quoted code assumes that when we call a method with underscore, the RPC interface assumes there is a dash version of it.

gallizoltan:
@m-schmoock Ok, understood. I will submit a PR later today.

PR #202, opened 2021-01-19
chore: renames RPC feeadjusttoggle to feeadjust-toggle

m-schmoock:
After small chat with cdecker, using dash/minus in RPC name is good and will be translated to underscore in pyln on the fly.

### DIFF
--- a/feeadjuster/feeadjuster.py
+++ b/feeadjuster/feeadjuster.py
@@ -206,8 +206,8 @@ def feeadjust(plugin: Plugin):
     return msg
 
 
-@plugin.method("feeadjustertoggle")
-def feeadjustertoggle(plugin: Plugin, value: bool = None):
+@plugin.method("feeadjuster-toggle")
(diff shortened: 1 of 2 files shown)


m-schmoock:
@gallizoltan as mentioned earlier ...

m-schmoock:
Also: do we want `feeadjuster-toggle` or `feeadjust-toggle` ? I tend to the first. Second one is only better if we rename the whole plugin, which I dont like.

gallizoltan:
tACK fb05a96
Now I can call `feeadjuster-toggle` (with dash) from command line and `feeadjuster_toggle` (with underscore) from python RPC.

gallizoltan:
> Also: do we want `feeadjuster-toggle` or `feeadjust-toggle` ? I tend to the first. Second one is only better if we rename the whole plugin, which I dont like.

I think the current `feeadjuster-toggle` is good in the code. Maybe the commit message could be more precise, but does not matter.

PR #206, opened 2021-01-27
Avoid out of IndexError from `historian-stats`

GuutBoy:
Fixes a bug where `historian-stats` would give IndexError when no announcements or updates where recorded yet.


cdecker:
ACK https://github.com/lightningd/plugins/pull/206/commits/950ab0c97aeb1f919f3e5677eaf8934686b1f9ec

PR #207, opened 2021-01-27
Fix reference to gossipd in historian.py

GuutBoy:
`gossipd` was referred to as `parser` resulting in errors.


cdecker:
ACK https://github.com/lightningd/plugins/pull/207/commits/5dce3e40ad105247358a539d6272faa9d7625e95

PR #208, opened 2021-01-28
add path for backyup solution info

MrManPew:


cdecker:
ACK https://github.com/lightningd/plugins/pull/208/commits/c721f58a5dbb099307307b9c0e774e79832e01ed

Issue #205, opened 2021-01-23
rebalance plugin crashes


thestick613:
```
2021-01-23T23:44:33.003Z INFO    plugin-rebalance.py: Plugin rebalance initialized with 1000msat base / 10 ppm fee, cltv_final: 18
2021-01-23T23:45:35.890Z UNUSUAL plugin-rebalance.py: Exception in thread Thread-1:
2021-01-23T23:45:35.890Z UNUSUAL plugin-rebalance.py: Traceback (most recent call last):
2021-01-23T23:45:35.891Z UNUSUAL plugin-rebalance.py:   File \"/usr/lib/python3.8/threading.py\", line 932, in _bootstrap_inner
2021-01-23T23:45:35.897Z UNUSUAL plugin-rebalance.py:     self.run()
2021-01-23T23:45:35.898Z UNUSUAL plugin-rebalance.py:   File \"/usr/lib/python3.8/threading.py\", line 870, in run
2021-01-23T23:45:35.898Z UNUSUAL plugin-rebalance.py:     self._target(*self._args, **self._kwargs)
2021-01-23T23:45:35.898Z UNUSUAL plugin-rebalance.py:   File \"/home/user/plugins/rebalance/rebalance.py\", line 433, in rebalanceall_thread
2021-01-23T23:45:35.899Z UNUSUAL plugin-rebalance.py:     plugin.ideal_ratio = get_ideal_ratio(channels, plugin.enough_liquidity)
2021-01-23T23:45:35.899Z UNUSUAL plugin-rebalance.py:   File \"/home/user/plugins/rebalance/rebalance.py\", line 304, in get_ideal_ratio
2021-01-23T23:45:35.900Z UNUSUAL plugin-rebalance.py:     if smallest_channel[\"total_msat\"] * min(ratio, 1 - ratio) > enough_liquidity:
2021-01-23T23:45:35.900Z UNUSUAL plugin-rebalance.py:   File \"/usr/local/lib/python3.8/dist-packages/pyln/client/lightning.py\", line 180, in __mul__
2021-01-23T23:45:35.901Z UNUSUAL plugin-rebalance.py:     return Millisatoshi(self.millisatoshis * other)
2021-01-23T23:45:35.901Z UNUSUAL plugin-rebalance.py:   File \"/usr/local/lib/python3.8/dist-packages/pyln/client/lightning.py\", line 72, in __init__
[...]

PR #211, opened 2021-02-01
rix: rebalance divsion by zero

m-schmoock:
As @aido found commented in #205 we had a division by zero issue in rebalance: https://github.com/lightningd/plugins/issues/205#issuecomment-770545323

### DIFF
--- a/rebalance/rebalance.py
+++ b/rebalance/rebalance.py
@@ -359,6 +359,34 @@ def liquidity_info(channel, enough_liquidity: Millisatoshi, ideal_ratio: float):
     return liquidity
 
 
+def wait_for(success, timeout: int = 60):
+    # cyclical lambda helper
+    # taken and modified from pyln-testing/pyln/testing/utils.py
+    start_time = time.time()
+    interval = 0.25
+    while not success():
+        time_left = start_time + timeout - time.time()
+        if time_left <= 0:
+            raise ValueError("Timeout error while waiting for {}", success)
+        time.sleep(min(interval, time_left))
+        interval *= 2
+        if interval > 5:
+            interval = 5
+
+
+def wait_for_htlcs(plugin, scids: list = None):
+    # HTLC settlement helper
+    # taken and modified from pyln-testing/pyln/testing/utils.py
+    peers = plugin.rpc.listpeers()['peers']
+    for p, peer in enumerate(peers):
+        if 'channels' in peer:
+            for c, channel in enumerate(peer['channels']):
+                if scids is not None and channel['short_channel_id'] not in scids:
+                    continue
+                if 'htlcs' in channel:
+                    wait_for(lambda: len(plugin.rpc.listpeers()['peers'][p]['channels'][c]['htlcs']) == 0)
+
+
 def maybe_rebalance_pairs(plugin: Plugin, ch1, ch2, failed_pairs: list):
     scid1 = ch1["short_channel_id"]
     scid2 = ch2["short_channel_id"]
@@ -395,8 +423,8 @@ def maybe_rebalance_pairs(plugin: Plugin, ch1, ch2, failed_pairs: list):
         plugin.log(f"Rebalance succeeded: {res}")
         result["success"] = True
         result["fee_spent"] += res["fee"]
-        # refresh channels
-        time.sleep(10)
+        # wait for settlement
+        wait_for_htlcs(plugin, [scid1, scid2])
         ch1 = get_chan(plugin, scid1)
         assert ch1 is not None
         ch2 = get_chan(plugin, scid2)
--- a/rebalance/rebalance.py
+++ b/rebalance/rebalance.py
@@ -256,7 +256,7 @@ def could_receive(liquidity):
     return a_minus_b(liquidity["their"], liquidity["min"])
 
 
-def get_our_channels(plugin: Plugin):
+def get_open_channels(plugin: Plugin):
     channels = []
     for peer in plugin.rpc.listpeers()["peers"]:
         for ch in peer["channels"]:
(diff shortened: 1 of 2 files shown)


m-schmoock:
@gallizoltan Do you agree?

gallizoltan (approved):
LGTM! Thank you, I've actually never tried to rebalance 0 channels! :D
Thanks for the nit fixes also!

rebalance/rebalance.py
```diff
@@ -352,12 +353,39 @@ def liquidity_info(channel, enough_liquidity: Millisatoshi, ideal_ratio: float):
...
+    # taken and modified from pyln-testing/pyln/testing/utils.py
+    start_time = time.time()
+    interval = 0.25
+    while not success() and time.time() < start_time + timeout:
+        time.sleep(interval)
+        interval *= 2
+        if interval > 5:
+            interval = 5
+    if time.time() > start_time + timeout:
+        raise ValueError("Error waiting for {}", success)
```
gallizoltan, line 373: Nit: `if not success():`
I think the `success` is more important here than the `timeout`.

rebalance/rebalance.py
```diff
@@ -394,8 +422,8 @@ def maybe_rebalance_pairs(plugin: Plugin, ch1, ch2, failed_pairs: list):
         plugin.log(f"Rebalance succeeded: {res}")
         result["success"] = True
         result["fee_spent"] += res["fee"]
-        # refresh channels
-        time.sleep(10)
+        # wait for settlement
+        wait_for_htlcs(plugin, [scid1, scid2])
```
gallizoltan, line 426: Yes, this is what I meant here back then. Thank you!

rebalance/rebalance.py
```diff
@@ -404,9 +432,11 @@ def maybe_rebalance_pairs(plugin: Plugin, ch1, ch2, failed_pairs: list):
 
 
 def maybe_rebalance_once(plugin: Plugin, failed_pairs: list):
-    channels = get_our_channels(plugin)
+    channels = get_open_channels(plugin)
     for ch1 in channels:
         for ch2 in channels:
+            if ch1 == ch2:
+                continue
```
gallizoltan, line 439: No need for this, I think. If `maybe_rebalance_pairs` calculates the sending amount correctly, in this case the amount will be zero, and there will be no try to rebalance.
(Of course, this can remain here, to make the thing safer/clearer.)

gallizoltan (approved):
LGTM! Thanks for the good work!
utACK

rebalance/rebalance.py
m-schmoock: One question about this `wait_for_htlcs` thingy. How should the `timeout` value be? Is 60s good or do you think its more sane to use a higher value out in the wild mainnet?

gallizoltan: I don't know. In theory, you could ask someone else.
In practice, if you put in a log about htlc time measure, I will run this code every day, and check the logs regularly. So we may figure out the correct value soon.

m-schmoock: Can you re-check your remark "I think the `success` is more important here than the `timeout`."?
I force pushed the code to only have two final states (not three):
 - success()
 - timeout() and success() := False
 
 Prior to this it was:
 - success()
 - timeout() and success() := False
 - timeout() and success() := True

Does this address your concern? Also see https://github.com/ElementsProject/lightning/pull/4369 for this...

m-schmoock:
Also, its nice that `rebalnceall` is now a lot quicker by not sleeping 10s each try but waiting for HTLC settlement ;)

m-schmoock:
@gallizoltan I figured the `rabalanceall` algo tries to repeat a certain chunk size, if they don't work it decreases the chunk to be more successful. However, this adds up on the static `fee_base` routing fees. Can you imagine/improve the algo to try a higher (highes?) chunk size first and not some small number? What are the potential drawbacks?

rebalance/rebalance.py
gallizoltan: Yes, it's a lot cleaner this way! Thank you!

m-schmoock: I think about changing it again (for the main repo) to make the final sleep cycle just so that it will terminate just at the time a timout would occur. But lets see what the others say. I'm not too optimistic about that change (in the main repo) as it does it can return AFTER timeout occured.

m-schmoock:
@gallizoltan what do you think about https://github.com/lightningd/plugins/issues/205#issuecomment-771642659 and https://github.com/lightningd/plugins/issues/205#issuecomment-771829114 ?

rebalance/rebalance.py
m-schmoock: Did a final change on that. maybe https://github.com/ElementsProject/lightning/pull/4370 is better where we do the final `time.sleep()` just with the right amount of time ;)

gallizoltan:
> @gallizoltan I figured the `rabalanceall` algo tries to repeat a certain chunk size, if they don't work it decreases the chunk to be more successful. However, this adds up on the static `fee_base` routing fees. Can you imagine/improve the algo to try a higher (highes?) chunk size first and not some small number? What are the potential drawbacks?

You can modify the `get_max_amount`, if you want to try higher chunk sizes. The drawback I see: with too high chunk size a transaction rarely works. So the `rebalanceall` may run longer for the futile tries. But this is not a big problem.

gallizoltan:
> @gallizoltan what do you think about [#205 (comment)](https://github.com/lightningd/plugins/issues/205#issuecomment-771642659) and [#205 (comment)](https://github.com/lightningd/plugins/issues/205#issuecomment-771829114) ?

I'm not sure. Maybe it would be good to check if `total - our < plugin.min_amount` or `our < plugin.min_amount` then no reason to run `rebalanceall`.

m-schmoock:
Okay, lets bring this PR in when tests ran green.
Now that we have a testable `rebalance_all` base we can try to construct cases that need to be fixed.
I don't understand (yet) why @aido is having issues I can't reproduce right now.

The issues he describes are where our == total, but our testcase totally covers this exact scenario. :confused:

Issue #209, opened 2021-01-29
backup: sqlite3 check in on_init can cause a hang at startup


laanwj:
The call to `listconfigs` inside the `db_sync` hook can hang indefinitely in some circumstances. See https://github.com/ElementsProject/lightning/issues/4348#issuecomment-769449638

I think it would be good to move this check to a different hook. Or remove it.

cdecker:
That is indeed correct, `db_write` is a synchronous hook, i.e., we don't process anything else until the hook returns. This also includes RPC calls such as `listconfigs`. I'm surprised however that we get stuck on the `init` call, which is used to tell us that the RPC interface is up and where to find it.

That is unless the `listconfigs` call creates a new DB transaction, which happens to trigger some changes, which then get reported back to `db_write`, however the plugin is busy waiting for the `listconfigs` reply and the `db_write` hook hangs indefinitely. Would be interesting to see if just moving the `listconfigs` call to a new thread is sufficient.

laanwj:
>  Would be interesting to see if just moving the listconfigs call to a new thread is sufficient.

It would be possible to defer the check, but as I understand the nature of the check seems to be to want to fail quickly in the case of a wrong database kind.

laanwj:
I agree it's really strange. I thought for a minute that `on_init` was called from first invocation of the `db_sync` hook, but it seems it is already a separate hook.

Not entirely sure what is happening here, to be honest.

m-schmoock:
Bet that's why the CI often failed/timeout a lot on the backup plugin lately.  ( @gallizoltan )
Someone already some clues? I will try to reproduce / isolate this.

m-schmoock:
When I run `test_compact` frequently by i.e. `py.test --count 100 -x -n 16 --timeout 60 -v backup/test_backup.py -k test_compact` there are deadlocking instances, see:

https://gist.github.com/m-schmoock/8ff49188399547c2d6eb5a188a68d6f9

The testcase itself unsurprisingly locks up when trying to get the nodes from the factory:
```python
l1 = node_factory.get_node(options=opts, cleandir=False)
```

If I see it correctly the threads are just waiting for basic stuff such as pytest threadpool or the daemon startup (read process stdout) which never happens. I think we need to modify the way `pyln-testing` is starting `lightningd` to see where it blocks...

m-schmoock:
[...]

PR #212, opened 2021-02-01
backup: fix deadlock by removing the sqlite3 check

m-schmoock:
'fixes' https://github.com/lightningd/plugins/issues/209

I commented out that check and placed a NOTE that future developers
to see upfront putting in RPC stuff in init is potentially bad.

### DIFF
--- a/backup/backup.py
+++ b/backup/backup.py
@@ -443,9 +443,12 @@ def on_init(options, **kwargs):
             level="warn"
         )
 
-    configs = plugin.rpc.listconfigs()
-    if not configs['wallet'].startswith('sqlite3'):
-        kill("The backup plugin only works with the sqlite3 database.")
+    # IMPORTANT NOTE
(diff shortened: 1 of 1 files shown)


cdecker:
Thanks @m-schmoock for reproducing and fixing this. I think quite a bit has to go wrong for this to happen: you need to a) perform an RPC call from a plugin that b) subscribes to a synchronous hook (only `db_write` atm iirc), and c) your RPC call needs to be interleaved with an action that triggers a `db_write` call (or a sync hook call if there are any others). On the downside this is not limited to the `init` call, but any place where we may end up in this situation. Luckily the backup plugin is pretty unique in this sense :-)

ACK https://github.com/lightningd/plugins/pull/212/commits/cdf8c7d39532bcfd370f1e2e04b5ab8a16bd3e10

laanwj:
Thanks!

PR #213, opened 2021-02-01
Add Dockerfile

3nprob:
This adds an extended Dockerfile based on the official image, with common dependencies (as well as postgresql to enable out-of-box compatability for that, which lacks in the base file).

The build arg `PLUGINS` can be modified to select specific plugins.

Notably plugins that require compilation (e.g. sparko) are not built here. That can be resolved either by layering either image on the other or copy over files from https://github.com/fiatjaf/sparko/blob/master/Dockerfile - or ideally this Dockerfile itself can be improved by adding build args to optionally build them using additional stages

Plenty of things that can be improved here and use-cases that aren't catered for explicitly, but I'm figuring this is a decently sane base for anyone to extend their custom things on while making it smoother to tie things together for a base setup.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+FROM elementsproject/lightningd
+
+ARG EXTRA_PLUGINS='--recurse-submodules=csvexportpays \
+--recurse-submodules=graphql \
+--recurse-submodules=jwt-factory \
+--recurse-submodules=python-teos \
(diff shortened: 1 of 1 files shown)


cdecker:
Very neat idea :+1:

ACK https://github.com/lightningd/plugins/pull/213/commits/4fc3a167829b0db3f74928a4bc43efabe38af5ef

PR #215, opened 2021-02-03
Fix missing parser error

GuutBoy:
Fixes an exception thrown for not having a parser for message type 4103. These messages are now just ignored (maybe should be handled somehow?)


cdecker:
ACK https://github.com/lightningd/plugins/pull/215/commits/93a217bb1fb53d66bb722efcc018ff7ac8f155b6

PR #216, opened 2021-02-03
rebalance: adds a check for no incoming liquidity

m-schmoock:
Currently, when we have 100% of the liquidity on our side run (expected) into the `rebalanceall` finishing without doing anything.

This is very rarely, but it can be confusing, so we add a pre-check that returns an error on `rebalanceall` in that scenario.
See:  https://github.com/lightningd/plugins/issues/205#issuecomment-772676555

### DIFF
--- a/rebalance/rebalance.py
+++ b/rebalance/rebalance.py
@@ -490,12 +490,23 @@ def rebalanceall(plugin: Plugin, min_amount: Millisatoshi = Millisatoshi("50000s
     To be economical, it tries to fix the liquidity cheaper than it can be ruined by transaction forwards.
     It may run for a long time (hours) in the background, but can be stopped with the rebalancestop method.
     """
+    # some early checks before we start the async thread
     if plugin.mutex.locked():
         return {"message": "Rebalance is already running, this may take a while. To stop it use the cli method 'rebalancestop'."}
-    if len(get_open_channels(plugin)) <= 1:
-        return {"message": "Error: Not enough open channels to balance anything"}
+    channels = get_open_channels(plugin)
+    if len(channels) <= 1:
+        return {"message": "Error: Not enough open channels to rebalance anything"}
+    our = sum(ch["to_us_msat"] for ch in channels)
+    total = sum(ch["total_msat"] for ch in channels)
+    min_amount = Millisatoshi(min_amount)
+    if total - our < min_amount or our < min_amount:
+        return {"message": "Error: Not enough liquidity to rebalance anything"}
+
+    # param parsing ensure correct type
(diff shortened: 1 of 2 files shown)


m-schmoock:
@gallizoltan
Do we want this early return?

rebalance/rebalance.py
```diff
@@ -492,8 +492,13 @@ def rebalanceall(plugin: Plugin, min_amount: Millisatoshi = Millisatoshi("50000s
...
         return {"message": "Rebalance is already running, this may take a while. To stop it use the cli method 'rebalancestop'."}
-    if len(get_open_channels(plugin)) <= 1:
-        return {"message": "Error: Not enough open channels to balance anything"}
+    channels = get_open_channels(plugin)
+    if len(channels) <= 1:
+        return {"message": "Error: Not enough open channels to rebalance anything"}
+    our = sum(ch["to_us_msat"] for ch in channels)
+    total = sum(ch["total_msat"] for ch in channels)
+    if our == total:
+        return {"message": "Error: No incoming liquidity to rebalance anything"}
```
gallizoltan, line 501: What about something like this?
```
if total - our < min_amount or our < min_amount:
    return {"message": "Error: Not enough liquidity to rebalance anything"}
```

rebalance/rebalance.py
```diff
@@ -490,10 +490,17 @@ def rebalanceall(plugin: Plugin, min_amount: Millisatoshi = Millisatoshi("50000s
...
         return {"message": "Rebalance is already running, this may take a while. To stop it use the cli method 'rebalancestop'."}
-    if len(get_open_channels(plugin)) <= 1:
-        return {"message": "Error: Not enough open channels to balance anything"}
+    channels = get_open_channels(plugin)
+    if len(channels) <= 1:
+        return {"message": "Error: Not enough open channels to rebalance anything"}
+    our = sum(ch["to_us_msat"] for ch in channels)
+    total = sum(ch["total_msat"] for ch in channels)
+    if total - our < Millisatoshi(min_amount) or our < Millisatoshi(min_amount):
+        return {"message": "Error: No incoming liquidity to rebalance anything"}
```
gallizoltan, line 502: Nit: "Error: Not enough liquidity to rebalance anything" is a better description, I think. I would not mention **incoming** liquidity, as we check both directions here.

gallizoltan (approved):
LGTM! Nice work, thank you!

m-schmoock:
Ok, lets merge then once we are green!

cdecker:
ACK https://github.com/lightningd/plugins/pull/216/commits/88c523cff7ff8a05ec51a08d2d2c165868063dd3

PR #217, opened 2021-02-04
Fix handling of ignored messages in historian.py

GuutBoy:
Fixes a bug where gossip messages that are ignored by the historian plugin will trigger an exception being printed.


cdecker:
ACK https://github.com/lightningd/plugins/pull/217/commits/99b563b7437074dd4655d6bb56755f09bf3cd0b3

PR #218, opened 2021-02-04
backup: Add support for IPv6 addresses in socket backend

laanwj:
Support the bracketed `socket:[::]:1234` syntax for IPv6 addresses.

Also, factor out the socket URI parsing to a function and add tests for it. 

As a forward compatibility measure, reject query strings (`?a=b`) because I intend to use these for parameters such as SOCKS5 proxy (for Tor) in a future patch.

### DIFF
--- a/backup/socketbackend.py
+++ b/backup/socketbackend.py
@@ -1,21 +1,71 @@
-import json
-import logging, socket, struct
+from collections import namedtuple
+import json, logging, socket, re, struct
 from typing import Tuple, Iterator
-from urllib.parse import urlparse
+from urllib.parse import urlparse, parse_qs
 
 from backend import Backend, Change
 from protocol import PacketType, recvall, PKT_CHANGE_TYPES, change_from_packet, packet_from_change, send_packet, recv_packet
+
+SocketURLInfo = namedtuple('SocketURLInfo', ['host', 'port', 'addrtype'])
+
+class AddrType:
+    IPv4 = 0
+    IPv6 = 1
+    NAME = 2
+
+def parse_socket_url(destination: str) -> SocketURLInfo:
+    '''Parse a socket: URL to extract the information contained in it.'''
+    url = urlparse(destination)
+    if url.scheme != 'socket':
+        raise ValueError('Scheme for socket backend must be socket:...')
+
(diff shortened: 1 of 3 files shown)


backup/socketbackend.py
```diff
@@ -1,21 +1,71 @@
...
+    except ValueError:
+        raise ValueError('Invalid port number')
+
+    # parse query parameters
+    # reject unknown parameters (currently all of them)
+    qs = parse_qs(url.query)
+    if len(qs):
+        raise ValueError('Invalid query string')
+
+    return SocketURLInfo(host=host, port=int(port), addrtype=addrtype)
```
cdecker, line 54: The `int()` cast is not required since you convert it on line 44. Doesn't harm though.

backup/socketbackend.py
```diff
@@ -1,21 +1,71 @@
...
+        eidx += 1
+        if eidx >= len(url.path) or url.path[eidx] != ':':
+            raise ValueError('Port number missing.')
+        eidx += 1
+    else:
+        eidx = url.path.find(':')
+        if eidx == -1:
+            raise ValueError('Port number missing.')
+        host = url.path[0:eidx]
+        if re.match('\d+\.\d+\.\d+\.\d+$', host): # matches IPv4 address format
```
cdecker, line 37: Since you're using `re` anyway here, why not make this entire function one big regular expression?

cdecker (approved):

backup/socketbackend.py
laanwj: Good catch, will remove it

backup/socketbackend.py
laanwj: I tried, but it became ugly as hell. Also this gives better error feedback.

laanwj:
Removed the unnecessary `int()`.

cdecker:
Great thanks ([range-diff](http://46.101.246.115:8888/lightningd/plugins/pull/218/range_diff/4ad16d8baaae7a507d1b5de8323608cc3dbbe5c2..3512292e7a60c63c4fe2034ddcb806d4269bf076)) 

ACK https://github.com/lightningd/plugins/pull/218/commits/3512292e7a60c63c4fe2034ddcb806d4269bf076

PR #220, opened 2021-02-06
Add missing submodule boltz channel-creation

3nprob:
I'm guessing this belongs here ith the rest?


cdecker:
Well we are already linking to the upstream repository. That has the advantage of not being pinned to a specific commit, which submodule are. Open to suggestions though, if a pinned version is desired.

3nprob:
> Well we are already linking to the upstream repository. That has the advantage of not being pinned to a specific commit, which submodule are. Open to suggestions though, if a pinned version is desired.

 Yeah, it caught me by surprise a bit (doesn't that go for all the other submodules as well? python-teos/watchtower is updated at leats as much? Maybe can be split into a separate list in the README and also include some other popular plugins like clboss?)

cdecker:
> Yeah, it caught me by surprise a bit (doesn't that go for all the other submodules as well? python-teos/watchtower is updated at leats as much? Maybe can be split into a separate list in the README and also include some other popular plugins like clboss?)

I wonder if it would make sense to just link to those plugins, instead of including them, either via in-tree copy or submodules. wdyt?

3nprob:
> > Yeah, it caught me by surprise a bit (doesn't that go for all the other submodules as well? python-teos/watchtower is updated at leats as much? Maybe can be split into a separate list in the README and also include some other popular plugins like clboss?)
> 
> I wonder if it would make sense to just link to those plugins, instead of including them, either via in-tree copy or submodules. wdyt?

 I think the submodule approach is a niceone making it very convenient yet easy to switch to your own branch.
 
 but just linking (separate list form the submodules list) is all good too I think , your call realy :)

3nprob:
Maybe the one thing arguing against the submodule approach is the increased time for you maintainer(s) to keep them up to date and review update-PRs - especially given the nature of this software it's especially important to not be laissez-fairez with those since sneaking in a hostile fork could be very bad.

Just a list removes all that completely.

3nprob:
@cdecker If you're still up for the submodules though, I could spend some more time on getting the dockerfile nicer and making a modular setup for all plugins with building and dependencies - should save a lot of redundant work for users

cdecker:
> @cdecker If you're still up for the submodules though, I could spend some more time on getting the dockerfile nicer and making a modular setup for all plugins with building and dependencies - should save a lot of redundant work for users.

That'd be awesome indeed. I'm still not decided whether it's best to keep the submodules, or have a machine readable metadata file pointing to repos and dependencies, but since we have submodules already we can start with those 👍

> So we don't *all* have to scratch our heads when python-cryptography starts requiring rust ;)

🤣

cdecker:
@mergifyio rebase

cdecker:
@Mergifyio rebase

cdecker (approved):

chrisguida:
Needs rebase

chrisguida:
Couple of things:

- we're moving to a new process where all plugins whose code is not hosted in the main repo tree must be explicitly linked to as submodules, because [the CI checks out the code and runs the tests for all submodules now](https://github.com/lightningd/plugins/pull/485)
- second, this plugin is no longer maintained and [has been archived](https://github.com/lightningd/plugins/pull/513), so this PR is no longer relevant. Let us know if you'd like to revive it! :)

PR #221, opened 2021-02-06
summary: Add fees_collected info to output

aido:
Hi,

It might be nice to see `fees_collected_msat` in summary output


cdecker:
ACK

PR #222, opened 2021-02-08
rebalance: measure and log HTLC settlement time

gallizoltan:
log messages show how much time need for an HTLC settlement


gallizoltan:
@m-schmoock what do you think? :)

m-schmoock:
ACK a709581

PR #226, opened 2021-02-15
backup: Improve logging configurability for socket server

laanwj:
Add two optional arguments to `backup-cli server`:

- `--log-mode`: either `plain` (simply print the log message) or `systemd` (prefix log message with systemd log level, useful for running as a  a service)

- `--log-level`: minimum level for messages to be logged, from `DEBUG` to `CRITICAL`

Also, move some noisy messages to the debug level.

Example systemd service file (might make sense to add this to the documentation):
```
[Unit]
Description=C-lightning backup server
After=network.target
StartLimitIntervalSec=0

[Service]
Type=simple
Restart=always
RestartSec=1
User=<user>
ExecStart=<path>/plugins/backup/backup-cli server --log-level debug --log-mode systemd file://<backup> 127.0.0.1:8700

[Install]
WantedBy=multi-user.target
```

### DIFF
--- a/backup/backup-cli
+++ b/backup/backup-cli
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 from backends import get_backend
 from backend import Change
-from server import SocketServer
+from server import SocketServer, setup_server_logging
 
 import os
 import click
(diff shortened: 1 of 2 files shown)


cdecker:
ACK https://github.com/lightningd/plugins/pull/226/commits/ef6be9f85f74e14749a501e4d7203f3737b7b915

PR #231, opened 2021-02-26
backup: Implement automatic reconnect in socket backend

laanwj:
Add automatic reconnect (with exponential back-off) to the socket backend.

If the connection is lost while pusing a change, try to reconnect. Handle edge-cases that might come up depending on when the disconnect happaned.

Currently the parameters are hardcoded: it will retry 5 times, on the first retry it will wait 5 seconds before reconnecting. There is an exponential backoff of 1.5, so on the fifth try it will wait about 25 seconds.

This is particularly useful when conencting over Tor, as transient interruptions are fairly common there.

Example output (reconnecting after one try):
```
2021-03-01T07:48:45.673Z INFO    plugin-backup.py: Connection was lost while sending change (retry 1 of 5, will try again after 5 seconds)
2021-03-01T07:48:50.802Z INFO    plugin-backup.py: Connecting to <>:8700 (addrtype 2, proxytype 1, proxytarget HostPortInfo(host='127.0.0.1', port=9050, addrtype=0))...
2021-03-01T07:48:56.709Z INFO    plugin-backup.py: Connected to <>:8700?proxy=socks5:127.0.0.1:9050
```

Example output (failing after 5 tries):
```
2021-03-01T08:07:54.125Z INFO    plugin-backup.py: Connection was lost while sending change (retry 1 of 5, will try again after 5 seconds)
2021-03-01T08:07:59.128Z INFO    plugin-backup.py: Connecting to <>:8700 (addrtype 2, proxytype 1, proxytarget HostPortInfo(host='127.0.0.1', port=9050, addrtype=0))...
2021-03-01T08:08:04.028Z INFO    plugin-backup.py: Connection was lost while sending change (retry 2 of 5, will try again after 7.5 seconds)
2021-03-01T08:08:11.530Z INFO    plugin-backup.py: Connecting to <>:8700 (addrtype 2, proxytype 1, proxytarget HostPortInfo(host='127.0.0.1', port=9050, addrtype=0))...
2021-03-01T08:08:12.406Z INFO    plugin-backup.py: Connection was lost while sending change (retry 3 of 5, will try again after 11.25 seconds)
2021-03-01T08:08:23.657Z INFO    plugin-backup.py: Connecting to <>:8700 (addrtype 2, proxytype 1, proxytarget HostPortInfo(host='127.0.0.1', port=9050, addrtype=0))...
2021-03-01T08:08:24.392Z INFO    plugin-backup.py: Connection was lost while sending change (retry 4 of 5, will try again after 16.875 seconds)
2021-03-01T08:08:41.268Z INFO    plugin-backup.py: Connecting to <>:8700 (addrtype 2, proxytype 1, proxytarget HostPortInfo(host='127.0.0.1', port=9050, addrtype=0))...
2021-03-01T08:08:43.216Z INFO    plugin-backup.py: Connection was lost while sending change (retry 5 of 5, will try again after 25.3125 seconds)
2021-03-01T08:09:08.546Z INFO    plugin-backup.py: Connecting to <>:8700 (addrtype 2, proxytype 1, proxytarget HostPortInfo(host='127.0.0.1', port=9050, addrtype=0))...
2021-03-01T08:09:09.339Z INFO    plugin-backup.py: Connection was lost while sending change (giving up after 5 retries)
2021-03-01T08:09:09.526Z **BROKEN** lightningd: Plugin …
```

### DIFF
--- a/backup/socketbackend.py
+++ b/backup/socketbackend.py
@@ -1,10 +1,19 @@
 from collections import namedtuple
-import json, logging, socket, re, struct
+import json, logging, socket, re, struct, time
 from typing import Tuple, Iterator
 from urllib.parse import urlparse, parse_qs
 
 from backend import Backend, Change
 from protocol import PacketType, recvall, PKT_CHANGE_TYPES, change_from_packet, packet_from_change, send_packet, recv_packet
+
+# Total number of reconnection tries
+RECONNECT_TRIES=5
+
+# Delay in seconds between reconnections (initial)
+RECONNECT_DELAY=5
+
+# Scale delay factor after each failure
+RECONNECT_DELAY_BACKOFF=1.5
 
 HostPortInfo = namedtuple('HostPortInfo', ['host', 'port', 'addrtype'])
 SocketURLInfo = namedtuple('SocketURLInfo', ['target', 'proxytype', 'proxytarget'])
@@ -85,7 +94,9 @@ class SocketBackend(Backend):
         self.prev_version = None
         self.destination = destination
         self.url = parse_socket_url(destination)
+        self.connect()
 
+    def connect(self):
         if self.url.proxytype == ProxyType.DIRECT:
             if self.url.target.addrtype == AddrType.IPv6:
                 self.sock = socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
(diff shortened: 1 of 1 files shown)


cdecker:
Looks good, but I wonder if it might be worth it to add a tiny dependency and externalize the retry mechanism (I'm looking at the `retry` decorator here https://pypi.org/project/retry/), instead of putting the retry logic inside the function we could then just annotate any function that needs to retry (in the current implementation retrieving the metadata for example is not retried).

laanwj:
Will look into it. I intentionally didn't choose to retry everything. E.g. I think user-initiated commands are okay to fail if the connection is not there, and reconnecting through `RESTORE` would need some additional logic besides simply retrying—the retry logic isn't entirely context independent.

> (in the current implementation retrieving the metadata for example is not retried).

Not when it fails after the initial connection, but the one inside `add_change` does—any error setting up the new connection will (or should at least, maybe im missing something) cause a retry.

rustyrussell:
Since this is a net improvement, I'm merging even though we may do better in future.

PR #237, opened 2021-03-25
paytest: Fix crash reported by m-schmoock on direct payment

cdecker:
We crash the plugin accidentally if we're not handling a paytest but
rather a payment destined for us, due to us falsely assuming the
short-channel-id to be set in the onion

Reported-by: Michael Schmoock <@m-schmoock>


m-schmoock:
tACK adb277d

m-schmoock:
ACK 086ee71

PR #246, opened 2021-05-28
Rebalance improvements

m-schmoock:
- this will fix and cleanup some things
- adds a feature where the plugin tries routes first that are more likely to be successful while brute forcing all possible routes later on. Meaning it will try shorter routes first, then a longer routes while simultaneously setting a higher requested capacity to `getroute` method and finally just trying all routes we can get.
- exclude nodes for future getroute requests that produce 4 or more `erring_node` errors (regardless the cause).

### DIFF
--- a/rebalance/rebalance.py
+++ b/rebalance/rebalance.py
@@ -179,7 +179,7 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
     excludes = [my_node_id]  # excude all own channels to prevent shortcuts
     nodes = {}               # here we store erring node counts
     maxhops = 1              # start with short routes and increase
-    msat_factor = plugin.msat_factor # start with high msatoshi factor to reduce
+    msatfactor = plugin.msatfactor # start with high msatoshi factor to reduce
                              # WIRE_TEMPORARY failures because of imbalances
 
     # 'disable' maxhops filter if set to <= 0
@@ -200,20 +200,20 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
                 r = plugin.rpc.getroute(incoming_node_id,
                                         fromid=outgoing_node_id,
                                         exclude=excludes,
-                                        msatoshi=msatoshi * msat_factor,
+                                        msatoshi=msatoshi * msatfactor,
                                         maxhops=maxhops,
                                         riskfactor=10, cltv=9)
                 time_getroute += time.time() - time_start
             except RpcError as e:
                 # could not find route -> change params and restart loop
                 if e.method == "getroute" and e.error.get('code') == 205:
-                    # reducce msatoshi factor to look for smaller channels now
-                    msat_factor -= 1
-                    if msat_factor == 0:
+                    # reduce msatfactor to look for smaller channels now
+                    msatfactor -= 1
+                    if msatfactor == 0:
                         # when we reached neutral msat factor:
-                        # increase maxhops and restart with msat_factor
+                        # increase maxhops and restart with msatfactor
                         maxhops = maxhops + 1
-                        msat_factor = plugin.msat_factor
+                        msatfactor = plugin.msatfactor
                         continue
                     # we observed that we only have a ~3% success rate on routes that
                     # are 8 hops or longer. Hence we cut at 6 (+2 in/out).
@@ -258,7 +258,7 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
 
             except RpcError as e:
                 time_sendpay += time.time() - time_start
-                plugin.log(f"maxhops:{maxhops}  msat_factor:{msat_factor}  running_for:{int(time.time()) - start_ts}  count_getroute:{count}  time_getroute:{time_getroute}  time_getroute_avg:{time_getroute / count}  count_sendpay:{count_sendpay}  time_sendpay:{time_sendpay}  time_sendpay_avg:{time_sendpay / count_sendpay}", 'debug')
+                plugin.log(f"maxhops:{maxhops}  msatfactor:{msatfactor}  running_for:{int(time.time()) - start_ts}  count_getroute:{count}  time_getroute:{time_getroute}  time_getroute_avg:{time_getroute / count}  count_sendpay:{count_sendpay}  time_sendpay:{time_sendpay}  time_sendpay_avg:{time_sendpay / count_sendpay}", 'debug')
                 #plugin.log(f"RpcError: {str(e)}", 'debug')
                 # check if we ran into the `rpc.waitsendpay` timeout
                 if e.method == "waitsendpay" and e.error.get('code') == 200:
@@ -616,11 +616,11 @@ def init(options, configuration, plugin):
     plugin.fee_ppm = config.get("fee-per-satoshi")
     plugin.mutex = Lock()
     plugin.maxhops = int(options.get("rebalance-maxhops"))
-    plugin.msat_factor = float(options.get("rebalance-msat_factor"))
+    plugin.msatfactor = float(options.get("rebalance-msatfactor"))
     plugin.log(f"Plugin rebalance initialized with {plugin.fee_base} base / {plugin.fee_ppm} ppm fee  "
                f"cltv_final:{plugin.cltv_final}  "
                f"maxhops:{plugin.maxhops}  "
-               f"msat_factor:{plugin.msat_factor}")
+               f"msatfactor:{plugin.msatfactor}")
 
 
 plugin.add_option(
@@ -633,7 +633,7 @@ plugin.add_option(
 )
 
 plugin.add_option(
-    "rebalance-msat_factor",
+    "rebalance-msatfactor",
     "4",
     "Will instruct `getroute` call to use higher requested capacity first. "
     "Note: This will decrease to 1 when no routes can be found.",
--- a/rebalance/rebalance.py
+++ b/rebalance/rebalance.py
@@ -197,8 +197,9 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
 
             try:
                 plugin.rpc.sendpay(route, payment_hash)
-                plugin.rpc.waitsendpay(payment_hash, retry_for + start_ts - int(time.time()))
-                return success_msg
+                running_for = int(time.time()) - start_ts
+                plugin.rpc.waitsendpay(payment_hash, max(retry_for - running_for, 0))
+                break
 
             except RpcError as e:
                 plugin.log("RpcError: " + str(e), 'debug')
--- a/rebalance/rebalance.py
+++ b/rebalance/rebalance.py
@@ -280,11 +280,11 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
                 if erring_channel is not None and erring_direction is not None:
                     excludes.append(erring_channel + '/' + str(erring_direction))
                 # count and exclude nodes that produce a lot of errors
-                if erring_node:
+                if erring_node and plugin.erringnodes > 0:
                     if nodes.get(erring_node) is None:
                         nodes[erring_node] = 0
                     nodes[erring_node] += 1
-                    if nodes[erring_node] > 3:
+                    if nodes[erring_node] >= plugin.erringnodes:
                         excludes.append(erring_node)
 
     except Exception as e:
@@ -624,10 +624,13 @@ def init(options, configuration, plugin):
     plugin.mutex = Lock()
     plugin.maxhops = int(options.get("rebalance-maxhops"))
     plugin.msatfactor = float(options.get("rebalance-msatfactor"))
+    plugin.erringnodes = int(options.get("rebalance-erringnodes"))
+
     plugin.log(f"Plugin rebalance initialized with {plugin.fee_base} base / {plugin.fee_ppm} ppm fee  "
                f"cltv_final:{plugin.cltv_final}  "
                f"maxhops:{plugin.maxhops}  "
-               f"msatfactor:{plugin.msatfactor}")
+               f"msatfactor:{plugin.msatfactor} "
+               f"erringnodes:{plugin.erringnodes}")
 
 
 plugin.add_option(
@@ -646,4 +649,13 @@ plugin.add_option(
     "Note: This will decrease to 1 when no routes can be found.",
     "string"
 )
+
+plugin.add_option(
+    "rebalance-erringnodes",
+    "5",
+    "Exclude nodes from routing that raised N or more errors. "
+    "Note: Use 0 to disable.",
+    "string"
+)
+
 plugin.run()
--- a/rebalance/rebalance.py
+++ b/rebalance/rebalance.py
@@ -173,7 +173,7 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
         excludes = [my_node_id]  # excude all own channels to prevent shortcuts
 
         while int(time.time()) - start_ts < retry_for and not plugin.rebalance_stop:
-            r = plugin.rpc.getroute(incoming_node_id, msatoshi, riskfactor=1, cltv=9, fromid=outgoing_node_id, exclude=excludes)
+            r = plugin.rpc.getroute(incoming_node_id, msatoshi, riskfactor=10, cltv=9, fromid=outgoing_node_id, exclude=excludes, maxhops=5)
             route_mid = r['route']
             route = [route_out] + route_mid + [route_in]
             setup_routing_fees(plugin, route, msatoshi)
--- a/rebalance/rebalance.py
+++ b/rebalance/rebalance.py
@@ -130,7 +130,9 @@ def calc_optimal_amount(out_ours, out_total, in_ours, in_total, payload):
 
 @plugin.method("rebalance")
 def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = None,
-              maxfeepercent: float = 0.5, retry_for: int = 60, exemptfee: Millisatoshi = Millisatoshi(5000)):
+              maxfeepercent: float = 0.5, retry_for: int = 60,
+              exemptfee: Millisatoshi = Millisatoshi(5000),
+              maxhops: int = None, msatfactor: float = None):
     """Rebalancing channel liquidity with circular payments.
 
     This tool helps to move some msatoshis between your channels.
@@ -139,6 +141,10 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
         msatoshi = Millisatoshi(msatoshi)
     maxfeepercent = float(maxfeepercent)
     retry_for = int(retry_for)
+    if maxhops is None:
+        maxhops = plugin.maxhops
+    if msatfactor is None:
+        msatfactor = plugin.msatfactor
     exemptfee = Millisatoshi(exemptfee)
     payload = {
         "outgoing_scid": outgoing_scid,
@@ -178,13 +184,14 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
     rpc_result = None
     excludes = [my_node_id]  # excude all own channels to prevent shortcuts
     nodes = {}               # here we store erring node counts
-    maxhops = 1              # start with short routes and increase
-    msatfactor = plugin.msatfactor # start with high msatoshi factor to reduce
+    _maxhops = 1             # start with short routes and increase
+    _msatfactor = msatfactor # start with high msatoshi factor to reduce
                              # WIRE_TEMPORARY failures because of imbalances
 
     # 'disable' maxhops filter if set to <= 0
+    # I know this is ugly, but we don't ruin the rest of the code this way
     if plugin.maxhops <= 0:
-        maxhops = 20
+        _maxhops = 20
 
     # trace stats
     count = 0
@@ -200,22 +207,22 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
                 r = plugin.rpc.getroute(incoming_node_id,
                                         fromid=outgoing_node_id,
                                         exclude=excludes,
-                                        msatoshi=msatoshi * msatfactor,
-                                        maxhops=maxhops,
+                                        msatoshi=msatoshi * _msatfactor,
+                                        maxhops=_maxhops,
                                         riskfactor=10, cltv=9)
                 time_getroute += time.time() - time_start
             except RpcError as e:
                 # could not find route -> change params and restart loop
                 if e.method == "getroute" and e.error.get('code') == 205:
-                    # reduce msatfactor to look for smaller channels now
-                    msatfactor -= 1
-                    if msatfactor < 1:
+                    # reduce _msatfactor to look for smaller channels now
+                    _msatfactor -= 1
+                    if _msatfactor < 1:
                         # when we reached neutral msat factor:
-                        # increase maxhops and restart with msatfactor
-                        maxhops = maxhops + 1
-                        msatfactor = plugin.msatfactor
+                        # increase _maxhops and restart with msatfactor
+                        _maxhops = _maxhops + 1
+                        _msatfactor = msatfactor
                     # abort if we reached maxhop limit
-                    if plugin.maxhops > 0 and maxhops > plugin.maxhops:
+                    if maxhops > 0 and _maxhops > maxhops:
                         rpc_result = {'status': 'error', 'message': 'No suitable routes found'}
                         return cleanup(plugin, label, payload, rpc_result)
                     continue
@@ -256,7 +263,7 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
 
             except RpcError as e:
                 time_sendpay += time.time() - time_start
-                plugin.log(f"maxhops:{maxhops}  msatfactor:{msatfactor}  running_for:{int(time.time()) - start_ts}  count_getroute:{count}  time_getroute:{time_getroute}  time_getroute_avg:{time_getroute / count}  count_sendpay:{count_sendpay}  time_sendpay:{time_sendpay}  time_sendpay_avg:{time_sendpay / count_sendpay}", 'debug')
+                plugin.log(f"maxhops:{_maxhops}  msatfactor:{_msatfactor}  running_for:{int(time.time()) - start_ts}  count_getroute:{count}  time_getroute:{time_getroute}  time_getroute_avg:{time_getroute / count}  count_sendpay:{count_sendpay}  time_sendpay:{time_sendpay}  time_sendpay_avg:{time_sendpay / count_sendpay}", 'debug')
                 #plugin.log(f"RpcError: {str(e)}", 'debug')
                 # check if we ran into the `rpc.waitsendpay` timeout
                 if e.method == "waitsendpay" and e.error.get('code') == 200:
@@ -483,7 +490,9 @@ def maybe_rebalance_pairs(plugin: Plugin, ch1, ch2, failed_channels: list):
         plugin.log(f"Try to rebalance: {scid1} -> {scid2}; amount={amount}; maxfee={maxfee}")
         start_ts = time.time()
         try:
-            res = rebalance(plugin, outgoing_scid=scid1, incoming_scid=scid2, msatoshi=amount, maxfeepercent=0, retry_for=1200, exemptfee=maxfee)
+            res = rebalance(plugin, outgoing_scid=scid1, incoming_scid=scid2,
+                    msatoshi=amount, maxfeepercent=0, retry_for=1200,
+                    exemptfee=maxfee, maxhops=6, msatfactor=1)
             if not res.get('status') == 'complete':
                 raise Exception  # fall into exception handler below
         except Exception:
--- a/rebalance/rebalance.py
+++ b/rebalance/rebalance.py
@@ -65,16 +65,22 @@ def find_worst_channel(route):
     return worst
 
 
-def cleanup(plugin, label, payload, success_msg, error=None):
+def cleanup(plugin, label, payload, rpc_result, error=None):
     try:
         plugin.rpc.delinvoice(label, 'unpaid')
     except RpcError as e:
         # race condition: waitsendpay timed out, but invoice get paid
         if 'status is paid' in e.error.get('message', ""):
-            return success_msg
-    if error is None:
-        error = RpcError("rebalance", payload, {'message': 'Rebalance failed'})
-    raise error
+            return rpc_result
+
+    if error is not None and isinstance(error, RpcError):
+        # unwrap rebalance errors as 'normal' RPC result
+        if error.method == "rebalance":
+            return {"status": "exception",
+                    "message": error.error.get('message', "error not given")}
+        raise error
+
+    return rpc_result
 
 
 # This function calculates the optimal rebalance amount
@@ -168,7 +174,8 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
     description = "%s to %s" % (outgoing_scid, incoming_scid)
     invoice = plugin.rpc.invoice(msatoshi, label, description, retry_for + 60)
     payment_hash = invoice['payment_hash']
-    success_msg = ""
+    rpc_result = None
+
     try:
         excludes = [my_node_id]  # excude all own channels to prevent shortcuts
 
@@ -188,7 +195,7 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
                 excludes.append(worst_channel['channel'] + '/' + str(worst_channel['direction']))
                 continue
 
-            success_msg = {"sent": msatoshi + fees, "received": msatoshi, "fee": fees, "hops": len(route),
+            rpc_result = {"sent": msatoshi + fees, "received": msatoshi, "fee": fees, "hops": len(route),
                            "outgoing_scid": outgoing_scid, "incoming_scid": incoming_scid, "status": "settled",
                            "message": f"{msatoshi + fees} sent over {len(route)} hops to rebalance {msatoshi}"}
             plugin.log("Sending %s over %d hops to rebalance %s" % (msatoshi + fees, len(route), msatoshi), 'debug')
@@ -198,24 +205,30 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
             try:
                 plugin.rpc.sendpay(route, payment_hash)
                 running_for = int(time.time()) - start_ts
-                plugin.rpc.waitsendpay(payment_hash, max(retry_for - running_for, 0))
-                break
+                result = plugin.rpc.waitsendpay(payment_hash, max(retry_for - running_for, 0))
+                if result.get('status') == "complete":
+                    return cleanup(plugin, label, payload, rpc_result)
 
             except RpcError as e:
-                plugin.log("RpcError: " + str(e), 'debug')
+                #plugin.log("RpcError: " + str(e), 'debug')
+                # check if we ran into the `rpc.waitsendpay` timeout
+                if e.method == "waitsendpay" and e.error.get('code') == 200:
+                    raise RpcError("rebalance", payload, {'message': 'Timeout reached'})
+                # check if we have problems with our own channels
                 erring_channel = e.error.get('data', {}).get('erring_channel')
                 if erring_channel == incoming_scid:
                     raise RpcError("rebalance", payload, {'message': 'Error with incoming channel'})
                 if erring_channel == outgoing_scid:
                     raise RpcError("rebalance", payload, {'message': 'Error with outgoing channel'})
+                # exclude other erroring channels
                 erring_direction = e.error.get('data', {}).get('erring_direction')
                 if erring_channel is not None and erring_direction is not None:
                     excludes.append(erring_channel + '/' + str(erring_direction))
 
     except Exception as e:
-        plugin.log("Exception: " + str(e), 'debug')
-        return cleanup(plugin, label, payload, success_msg, e)
-    return cleanup(plugin, label, payload, success_msg)
+        return cleanup(plugin, label, payload, rpc_result, e)
+    rpc_result = {'status': 'error', 'message': 'Timeout reached'}
+    return cleanup(plugin, label, payload, rpc_result)
 
 
 def a_minus_b(a: Millisatoshi, b: Millisatoshi):
@@ -416,6 +429,8 @@ def maybe_rebalance_pairs(plugin: Plugin, ch1, ch2, failed_channels: list):
         start_ts = time.time()
         try:
             res = rebalance(plugin, outgoing_scid=scid1, incoming_scid=scid2, msatoshi=amount, maxfeepercent=0, retry_for=1200, exemptfee=maxfee)
+            if not res.get('status') == 'complete':
+                raise Exception  # fall into exception handler below
         except Exception:
             failed_channels.append(scid1 + ":" + scid2)
             # rebalance failed, let's try with a smaller amount
--- a/rebalance/rebalance.py
+++ b/rebalance/rebalance.py
@@ -174,13 +174,45 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
     description = "%s to %s" % (outgoing_scid, incoming_scid)
     invoice = plugin.rpc.invoice(msatoshi, label, description, retry_for + 60)
     payment_hash = invoice['payment_hash']
+
     rpc_result = None
+    count = 0
+    excludes = [my_node_id]  # excude all own channels to prevent shortcuts
+    nodes = {}               # here we store erring node counts
+    maxhops = 1              # start with short routes and increase
+    msat_factor = 4          # start with high msatoshi factor to reuce
+                             # WIRE_TEMPORARY failures because of imbalances
 
     try:
-        excludes = [my_node_id]  # excude all own channels to prevent shortcuts
-
         while int(time.time()) - start_ts < retry_for and not plugin.rebalance_stop:
-            r = plugin.rpc.getroute(incoming_node_id, msatoshi, riskfactor=10, cltv=9, fromid=outgoing_node_id, exclude=excludes, maxhops=5)
+            count += 1
+            try:
+                r = plugin.rpc.getroute(incoming_node_id,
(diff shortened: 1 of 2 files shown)


m-schmoock:
@gallizoltan Do you want to have a look at this? When this round is done we should start with a multipart payment implementation.

I will also adapt this changes to the drain plugin, which really needs some caretaking...

rebalance/rebalance.py
```diff
@@ -168,16 +174,35 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
...
-    try:
-        excludes = []
-        # excude all own channels to prevent unwanted shortcuts [out,mid,in]
-        mychannels = plugin.rpc.listchannels(source=my_node_id)['channels']
-        for channel in mychannels:
-            excludes += [channel['short_channel_id'] + '/0', channel['short_channel_id'] + '/1']
 
+    rpc_result = None
+    count = 0
+    excludes = [my_node_id]  # excude all own channels to prevent shortcuts
```
gallizoltan, line 180: Wow! Much cleaner than the previous version:
```
for channel in mychannels:
    excludes += [channel['short_channel_id'] + '/0', channel['short_channel_id'] + '/1']
```
Nice work!

gallizoltan (approved):
LGTM! Much better this way! Thank you for the nice work!

rebalance/rebalance.py
```diff
@@ -168,16 +174,35 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
...
+            except RpcError as e:
+                # could not find route -> change params and restart loop
+                if e.method == "getroute" and e.error.get('code') == 205:
+                    maxhops = maxhops + 1
+                    if maxhops == 5:    # set msatoshi factor to neutral
+                        msat_factor = 1
+                        continue
+                    if maxhops == 10:
+                        rpc_result = {'status':'error', 'message': 'No suitable routes found'}
+                        return cleanup(plugin, label, payload, rpc_result)
```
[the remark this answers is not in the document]
m-schmoock: Question about this: Do your numbers include the `route_in` and `route_out` first and last hop we add manually? Because we do `getroute` only for the path in between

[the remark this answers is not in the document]
gallizoltan: Yes, my numbers originate from this message:
```
success_msg = {"sent": msatoshi + fees, "received": msatoshi, "fee": fees, "hops": len(route),
               "outgoing_scid": outgoing_scid, "incoming_scid": incoming_scid, "status": "settled",
               "message": f"{msatoshi + fees} sent over {len(route)} hops to rebalance {msatoshi}"}
```
The `route_in` and `route_out` is included.

rebalance/rebalance.py
```diff
@@ -168,16 +174,35 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
...
-        mychannels = plugin.rpc.listchannels(source=my_node_id)['channels']
-        for channel in mychannels:
-            excludes += [channel['short_channel_id'] + '/0', channel['short_channel_id'] + '/1']
 
+    rpc_result = None
+    count = 0
+    excludes = [my_node_id]  # excude all own channels to prevent shortcuts
+    nodes = {}               # here we store erring node counts
+    maxhops = 1              # retry higher when getroute fails
+    msat_factor = 4
```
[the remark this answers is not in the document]
m-schmoock: Not sure, Im still trying and fiddlig with numbers, wanted your feedback on the idea itself.

My rationale was to try very short routes first and then go up to 3,4,5 hops while simultaniously increasing this factor, because it increases the likelyhood a channel will not throw `WIRE_TEMPORARY_CHANNEL_FAILURE` because its to imbalanced.

After this has been tried the also should remove this tricks and try default auto routing of i.e. 8 or 10 max hops.

Also we could consider clearing the `nodes` list which is filled with bad performing nodes on the go.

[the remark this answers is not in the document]
m-schmoock: Note: the `msat_factor` is used in a `max(maxhops, msat_factor)` function so it increases as we try longer routes.
We can rename the variable to reflect that ;)

rebalance/rebalance.py
m-schmoock: yeah, just a little rtfm and we could have seen this earlier ;)

rebalance/rebalance.py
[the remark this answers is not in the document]
m-schmoock: @gallizoltan also, how do you think about `riskfactor`? I set it to `10` where it was `1` iirc, as 10 is also used for normal sendpay.

I tried around with it and recognized that if we use very low values (0 or 1) it tends to return longer routes with bigger `delay` values, which is to be expected, there seems to be some inconsistencies (maybe by design):

Low risk values tend to suppress certain possible short routes, see for example (retry on your own with some ID, you will figure out soon):
```
# try to find a route with low riskfactor and small maxhops
lightning-cli getroute id=02843833374b6c42daa4b780df2066a9985cb8cf97bc4566dd63cf693d32cbf7b5 msatoshi=0.001btc riskfactor=0 maxhops=3
{
   "code": 205,
   "message": "Could not find a route"
}

# same with higher riskfactor gives a route:
lightning-cli getroute id=02843833374b6c42daa4b780df2066a9985cb8cf97bc4566dd63cf693d32cbf7b5 msatoshi=0.001btc riskfactor=10 maxhops=3
{
   "route": [
      {
         "id": "0298f6074a454a1f5345cb2a7c6f9fce206cd0bf675d177cdbf0ca7508dd28852f",
         "channel": "685285x349x0",
         "direction": 1,
         "msatoshi": 100003700,
         "amount_msat": "100003700msat",
         "delay": 63,
         "style": "tlv"
      },
      {
         "id": "03c2abfa93eacec04721c019644584424aab2ba4dff3ac9bdab4e9c97007491dda",
         "channel": "579764x2740x0",
         "direction": 0,
         "msatoshi": 100003500,
         "amount_msat": "100003500msat",
         "delay": 29,
         "style": "tlv"
      },
      {
         "id": "02843833374b6c42daa4b780df2066a9985cb8cf97bc4566dd63cf693d32cbf7b5",
         "channel": "564229x2034x0",
         "direction": 1,
         "msatoshi": 100000000,
         "amount_msat": "100000000msat",
         "delay": 9,
         "style": "tlv"
      }
   ]
}
```

Normally I would assume that it is harder to find routes with a higher rather than lower `riskfactor` value...

m-schmoock:
Force push:
- renamed msat_factor to msat_factor_max
- set the msat_factor_max value to 2 when we are trying effective 6 hops or more  (brute force mode)
- cutoff routes with effective length of 8 or more. They make only 3% total success rate but take a lot of time to try.

m-schmoock:
Force push:

- renamed `msat_factor` to `msat_factor_max`.
- set the `msat_factor_max` value to `2` when we are trying effective `6` hops or more (brute force mode)
- cutoff routes with effective length of `8` or more. They make only 3% total success rate but take a lot of time to try.
- fix non-positive rpc_result being used in `rebalanceall`.

rebalance/rebalance.py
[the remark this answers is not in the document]
gallizoltan: Ok, I tell you the story of `riskfactor`.
I was writing these parts of the code 2 years ago, and honestly, I had no idea what is `riskfactor`. So I've read the manual, which says ['1 is a conservative value for a stable lightning network with very few failures.'](https://github.com/ElementsProject/lightning/blob/master/doc/lightning-getroute.7.md#recommended-riskfactor-values)
This sounded comforting to me. So I've decided to `riskfactor=1` can be a good choice. I've tried it, and it actually worked. And I was happy about it. 

I didn't try any other values beyond 1. Feel free to optimize it!

rebalance/rebalance.py
```diff
@@ -168,16 +174,40 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
...
+
+    try:
         while int(time.time()) - start_ts < retry_for and not plugin.rebalance_stop:
-            r = plugin.rpc.getroute(incoming_node_id, msatoshi, riskfactor=1, cltv=9, fromid=outgoing_node_id, exclude=excludes)
+            count += 1
+            try:
+                r = plugin.rpc.getroute(incoming_node_id,
+                                        fromid=outgoing_node_id,
+                                        exclude=excludes,
+                                        msatoshi=msatoshi * min(maxhops, msat_factor_max),
```
gallizoltan, line 192: So `maxhops` can be between 1 and 5? It can be a reasonable choice, reducing lots of unnecessary route searches.

And this `min(maxhops, msat_factor_max)` multiplier is:
```
maxhops=1 -> 1
maxhops=2 -> 2
maxhops=3 -> 3
maxhops=4 -> 2
maxhops=5 -> 2
```
Is this what we want? (I don't know it is good or bad, just asking. Maybe I need to sleep one to think about this. :) )

m-schmoock: yes, good night. Just one question, maybe you read this tomorrow: I just tried to rebalance our channel so that your node has less balance on you side, meaning I used our channel as incoming_scid. Funnily your node gave me `WIRE_TEMPORARY_CHANNEL_FAILURE` even thou you have enough liqudity to push to me:

```
2021-05-28T22:38:26.322Z DEBUG   plugin-rebalance.py: count:1 maxhops:1 RpcError: RPC call failed: method: 
waitsendpay, payload: {'payment_hash': 'd3b07b31919b580272942d8672f81fb55ffd3d5de92015d684f951eb514e8467', 
'timeout': 600}, error: {'code': 204, 'message': 'failed: WIRE_TEMPORARY_CHANNEL_FAILURE (reply from remote)', 
'data': {'id': 224, 'payment_hash': 'd3b07b31919b580272942d8672f81fb55ffd3d5de92015d684f951eb514e8467', 
'destination': '02a4aae2e021b0785f2f2a7c9dc9089361995085298598992bf69f11302a5dbd2c', 'msatoshi': 1000000000, 
'amount_msat': 1000000000msat, 'msatoshi_sent': 1000411495, 'amount_sent_msat': 1000411495msat, 'created_at': 
1622241503, 'status': 'pending', 'erring_index': 2, 'failcode': 4103, 'failcodename': 
'WIRE_TEMPORARY_CHANNEL_FAILURE', 'erring_node': 
'03d67f36c4f81789e2fe425028bacc96b199813eae426c517f589a45f1136c1fe5', 'erring_channel': '685132x1087x0', 
'erring_direction': 1, 'raw_message': '1...''}}
```

Are there other reasons (aside imbalanced channel) that a forward can fail with `WIRE_TEMPORARY_CHANNEL_FAILURE` ??

m-schmoock: And also ACINQ node seems to give me `WIRE_CHANNEL_DISABLED` , ever had this?

gallizoltan: I've checked my logs and I've found this:
```
2021-05-28T22:38:25.203Z INFO    02a4aae2e021b0785f2f2a7c9dc9089361995085298598992bf69f11302a5dbd2c-chan#13696: Attempt to send HTLC but unowned (CHANNELD_NORMAL)
```
It's your node id. But I don't know, what is "Attempt to send HTLC but unowned" means.

Also, I've checked for `WIRE_CHANNEL_DISABLED` and I have some:
```
2021-05-28T20:32:20.990Z INFO    02f3069a342ae2883a6f29e275f06f28a56a6ea2e2d96f5888a3266444dcf542b6-chan#13597: htlc 14017 failed from 2th node with code 0x1014 (WIRE_CHANNEL_DISABLED)
...
2021-05-29T06:38:59.841Z INFO    024a8228d764091fce2ed67e1a7404f83e38ea3c7cb42030a2789e73cf3b341365-chan#13580: htlc 14104 failed from 2th node with code 0x1014 (WIRE_CHANNEL_DISABLED)
```
The second one surely comes from a c-lightning node.

gallizoltan: About the `msat_factor_max`, could it be a config option? I'm still not convinced that it is a good idea. I would like the option to set it to 1 if it isn't working well. Or may try different values too.

m-schmoock: Maybe I remove the msatoshi scaling all together since I'm also unsure. We need to to A-B tests to know if its better.

What is your opinion about the incremental approach to maxhops in trying short routes first? Do we want this?

m-schmoock: Other way would be to keep the msatoshi scaling in a modified way, maybe less aggressive and turning it to neutral (1) when it failed to a certain maxhop, so the algorithm still tries all the routes...

rebalance/rebalance.py
```diff
@@ -208,6 +210,16 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
...
                     continue
+
+            # build route cache and filter duplicates
+            route_str = ""
+            for hop in r['route']:
+                route_str += hop['channel'] + ' '
+            if routes.get(route_str) is not None:
+                rpc_result = {'status':'error', 'message': 'ROUTE DUP FOUND'}
+                return cleanup(plugin, label, payload, rpc_result)
+            routes[route_str] = True
```
gallizoltan, line 221: This should be unnecessary. If a route failed previously, then one of its components is in the `excludes` list. So there should be no duplicates anyway.
Or what I miss?

rebalance/rebalance.py
gallizoltan: I'm not sure about the incremental approach of `maxhops`. If we call `getroute(maxhops=4)` and after that `getroute(maxhops=5)` it will do partly the same route search twice?

Could this incremental search be configurable? So I may choose just one search with for example `maxhops=5`?

rebalance/rebalance.py
m-schmoock: Sorry, needed to push work-in-progress nonsense ... my machine broke down and I'm reinstalling now :D

m-schmoock:
@gallizoltan can you re-review the code that modifies the getroute parameters on each iteration like I explained in the mailing list:

- Start trying shorter routes first via "maxhops" parameter.
- In the first iteration use a lot higher "msatoshi" value than needed, for example by a factor of 4 times bigger.
- When you get "route not found errors" decrease the msatoshi factor and retry getroute.
- When you reached a msatoshi factor of 1 (neutral) and still run into "route not found": increase the maxhops paramter by 1 and reset the msatoshi factor to 4 again for the next interation.
- Repeat the loop until maxhop reached i.e. 10

This way we still try all the routes but make sure we look for short and bigger routes first. I think something like this is required for the rebalance plugin (currently) as it does not yet support MPP and potential rebalance amounts are just too high. At least thats what I am struggling with...

rebalance/rebalance.py
```diff
@@ -168,16 +174,43 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
...
+                # could not find route -> change params and restart loop
+                if e.method == "getroute" and e.error.get('code') == 205:
+                    # reducce msatoshi factor to look for smaller channels now
+                    msat_factor -= 1
+                    if msat_factor == 1:
+                        # when we reached neutral msat factor:
+                        # increase maxhops and restart with high msatoshi factor
+                        maxhops = maxhops + 1
+                        msat_factor = 4
+                        continue
```
gallizoltan, line 207: The condition should be `if msat_factor == 0:`.  Otherwise we will never check the neutral msat factor 1.

rebalance/rebalance.py
```diff
@@ -168,16 +174,43 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
...
-        for channel in mychannels:
-            excludes += [channel['short_channel_id'] + '/0', channel['short_channel_id'] + '/1']
 
+    rpc_result = None
+    count = 0
+    excludes = [my_node_id]  # excude all own channels to prevent shortcuts
+    nodes = {}               # here we store erring node counts
+    routes = {}              # trace all routes to suppress duplicates
+    maxhops = 1              # start with short routes and increase
+    msat_factor = 4          # start with high msatoshi factor to reuce
```
gallizoltan, line 184: Could this be something like this?
```
    maxhops = plugin.starting_maxhops
    msat_factor = plugin.starting_msat_factor
```
And then `plugin.starting_maxhops` and `plugin.starting_msat_factor` could come from config, with a default value of 1 and 4. But it would be possible to set them 5 and 1, to call `getroute` with fixed `maxhops=5` and `msat_factor=1`.

gallizoltan (commented):
Actually can be good. If `msat_factor` goes from 4 to 1, and `maxhops` goes from 1 to 5, it is 4*5=20 different `getroute` search. And the default time limit is still 20 minutes, as I can see. I'm not sure, it's enough time for this many searches, while a lot of them are multiplied work: for example if we fail with `maxhops=3` and then try `maxhops=4`, it will search again all the 3-hops possibilities too.

Could this be configurable? To use fixed `maxhops=5` and `msat_factor=1`?

rebalance/rebalance.py
m-schmoock: > I've checked my logs and I've found this:
> 
> ```
> 2021-05-28T22:38:25.203Z INFO    02a4aae2e021b0785f2f2a7c9dc9089361995085298598992bf69f11302a5dbd2c-chan#13696: Attempt to send HTLC but unowned (CHANNELD_NORMAL)
> ```
> 
> It's your node id. But I don't know, what is "Attempt to send HTLC but unowned" means.
> ...

Talked to cdecker about this, my node happened to turn offline just at the moment your node tried to send me :D

rebalance/rebalance.py
m-schmoock: good catch!

About the default values: sure it can be configurable and turned of per default.
But: we should test something like this in practice before hiding it behind a config switch that nobody uses ;) 

In the mailing list there's some chat about it that the (only?) potential drawback is computational time used for getroute.
If that's true then its really a matter of CPU vs Net-Latency tradeoff which in a lot situations will be no problem for the CPU.

rebalance/rebalance.py
m-schmoock: Sure, lets test it to find best values. Or maybe other approaches to improve successful path finding

m-schmoock: In the mailing list there's some chat about it that the (only?) potential drawback is computational time used for getroute.
If that's true then its really a matter of CPU vs Net-Latency tradeoff which in a lot situations will be no problem for the CPU.

Lets optimize values... we need to trace how much CPU getroute overhead is caused vs how much net io is saved ...

m-schmoock: Maybe we can have a "reference value" about how long an average `sendpay` takes and then let the plugin runtime measure how fast `getroute` is on this machine to turn on or off this optimization ...

m-schmoock:
@gallizoltan 

Hey, I added config plugin switches `rebalance-maxhops` and `rebalance-msat_factor` and time tracing in the result and log for it. We can now test and fiddle with the values. But as far as I can tell from my slow CPU RaspberryPi using this getroute optimization actually makes more sense than iterating the routes one by one as we get them.

Try for yourself and compare the numbers please...

rebalance/rebalance.py
```diff
@@ -546,8 +612,25 @@ def init(options, configuration, plugin):
...
+    plugin.msat_factor = float(options.get("rebalance-msat_factor"))
+    plugin.log(f"Plugin rebalance initialized with {plugin.fee_base} base / {plugin.fee_ppm} ppm fee  "
+               f"cltv_final:{plugin.cltv_final}  "
+               f"maxhops:{plugin.maxhops}  "
+               f"msat_factor:{plugin.msat_factor}")
+
+plugin.add_option(
+    "rebalance-maxhops",
+    "6",
+    "Maximum number of hops for `getroute` call. "
```
gallizoltan, line 625: This is the starting value of the `maxhops` iteration. Maybe you intended to set it to "1"?

gallizoltan (approved):
LGTM! Time will test it for sure!

rebalance/rebalance.py
m-schmoock: lol, thx

m-schmoock: fixed in recent force push

rebalance/rebalance.py
```diff
@@ -186,39 +229,60 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
...
+                    if nodes[erring_node] > 3:
+                        excludes.append(erring_node)
 
     except Exception as e:
-        plugin.log("Exception: " + str(e), 'debug')
-        return cleanup(plugin, label, payload, success_msg, e)
-    return cleanup(plugin, label, payload, success_msg)
+        return cleanup(plugin, label, payload, rpc_result, e)
+    rpc_result = {'status':'error', 'message': 'Timeout reached'}
+    return cleanup(plugin, label, payload, rpc_result)
```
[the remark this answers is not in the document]
m-schmoock: will recheck, I thought so ;)

rebalance/rebalance.py
```diff
@@ -186,39 +229,60 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
...
             try:
                 plugin.rpc.sendpay(route, payment_hash)
-                plugin.rpc.waitsendpay(payment_hash, retry_for + start_ts - int(time.time()))
-                return success_msg
+                running_for = int(time.time()) - start_ts
+                result = plugin.rpc.waitsendpay(payment_hash, max(retry_for - running_for, 0))
+                time_sendpay += time.time() - time_start
+                if result.get('status') == "complete":
+                    rpc_result["stats"] = f"running_for:{int(time.time()) - start_ts}  count_getroute:{count}  time_getroute:{time_getroute}  time_getroute_avg:{time_getroute / count}  count_sendpay:{count_sendpay}  time_sendpay:{time_sendpay}  time_sendpay_avg:{time_sendpay / count_sendpay}"
+                    return cleanup(plugin, label, payload, rpc_result)
```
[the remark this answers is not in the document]
m-schmoock: will recheck, I thought so ;)

m-schmoock:
On my Raspberry PI I now get results like:

m-schmoock:
Using current default values on my Raspberry PI I get results like:

`"stats": "running_for:5  count_getroute:12  time_getroute:0.28259897232055664  time_getroute_avg:0.023549914360046387  count_sendpay:4  time_sendpay:2.7961127758026123  time_sendpay_avg:0.6990281939506531"`

Which indicate exessive calling of `getroute` isn't the big deal  but `sendpay` is.
How are your results?

m-schmoock:
@gallizoltan 
Hi I made both values configureable. Setting `rebalance-maxhops` to "0" will disable the maxhops filter.
I made a statistically not relevant test, here are the results

```
# Disable both filters (like without this PR, just the erring node filter=
#> lnc rebalance outgoing_scid=685247x957x0 incoming_scid=685132x1087x0 msatoshi=0.005btc
Plugin rebalance initialized with 1000msat base / 10 ppm fee  cltv_final:18  maxhops:0  msatfactor:1.0
maxhops:20  msatfactor:1.0  running_for:4  count_getroute:1  time_getroute:0.16330862045288086  time_getroute_avg:0.16330862045288086  count_sendpay:1  time_sendpay:4.338505029678345  time_sendpay_avg:4.338505029678345
maxhops:20  msatfactor:1.0  running_for:8  count_getroute:2  time_getroute:0.46318650245666504  time_getroute_avg:0.23159325122833252  count_sendpay:2  time_sendpay:7.869954586029053  time_sendpay_avg:3.9349772930145264
maxhops:20  msatfactor:1.0  running_for:12  count_getroute:3  time_getroute:0.6282370090484619  time_getroute_avg:0.2094123363494873  count_sendpay:3  time_sendpay:11.94294023513794  time_sendpay_avg:3.980980078379313
maxhops:20  msatfactor:1.0  running_for:19  count_getroute:4  time_getroute:0.845451831817627  time_getroute_avg:0.21136295795440674  count_sendpay:4  time_sendpay:18.372679471969604  time_sendpay_avg:4.593169867992401
maxhops:20  msatfactor:1.0  running_for:23  count_getroute:5  time_getroute:1.018324375152588  time_getroute_avg:0.20366487503051758  count_sendpay:5  time_sendpay:22.277318716049194  time_sendpay_avg:4.4554637432098385
maxhops:20  msatfactor:1.0  running_for:27  count_getroute:6  time_getroute:1.1925179958343506  time_getroute_avg:0.1987529993057251  count_sendpay:6  time_sendpay:25.69105839729309  time_sendpay_avg:4.281843066215515
maxhops:20  msatfactor:1.0  running_for:31  count_getroute:7  time_getroute:1.4166362285614014  time_getroute_avg:0.2023766040802002  count_sendpay:7  time_sendpay:29.830421209335327  time_sendpay_avg:4.261488744190761
maxhops:20  msatfactor:1.0  running_for:35  count_getroute:8  time_getroute:1.590040683746338  time_getroute_avg:0.19875508546829224  count_sendpay:8  time_sendpay:33.221359729766846  time_sendpay_avg:4.152669966220856
maxhops:20  msatfactor:1.0  running_for:38  count_getroute:9  time_getroute:1.763425588607788  time_getroute_avg:0.19593617651197645  count_sendpay:9  time_sendpay:36.21986770629883  time_sendpay_avg:4.024429745144314
maxhops:20  msatfactor:1.0  running_for:42  count_getroute:10  time_getroute:1.9394252300262451  time_getroute_avg:0.19394252300262452  count_sendpay:10  time_sendpay:39.911234855651855  time_sendpay_avg:3.9911234855651854
maxhops:20  msatfactor:1.0  running_for:44  count_getroute:11  time_getroute:2.115779161453247  time_getroute_avg:0.19234356013211337  count_sendpay:11  time_sendpay:42.25611186027527  time_sendpay_avg:3.841464714570479
maxhops:20  msatfactor:1.0  running_for:49  count_getroute:12  time_getroute:2.32531476020813  time_getroute_avg:0.19377623001734415  count_sendpay:12  time_sendpay:46.37626266479492  time_sendpay_avg:3.8646885553995767
maxhops:20  msatfactor:1.0  running_for:51  count_getroute:13  time_getroute:2.5044617652893066  time_getroute_avg:0.19265090502225435  count_sendpay:13  time_sendpay:48.622201442718506  time_sendpay_avg:3.740169341747577
maxhops:20  msatfactor:1.0  running_for:54  count_getroute:14  time_getroute:2.6831555366516113  time_getroute_avg:0.19165396690368652  count_sendpay:14  time_sendpay:51.87213206291199  time_sendpay_avg:3.705152290207999
maxhops:20  msatfactor:1.0  running_for:58  count_getroute:15  time_getroute:2.9157254695892334  time_getroute_avg:0.19438169797261556  count_sendpay:15  time_sendpay:55.00669240951538  time_sendpay_avg:3.6671128273010254

# Now with the currently suggested default values maxhops := 5  and msatfactor :=4
Plugin rebalance initialized with 1000msat base / 10 ppm fee  cltv_final:18  maxhops:5  msatfactor:4.0
maxhops:3  msatfactor:4.0  running_for:3  count_getroute:9  time_getroute:0.06924104690551758  time_getroute_avg:0.00769344965616862  count_sendpay:1  time_sendpay:0.7183606624603271  time_sendpay_avg:0.7183606624603271
maxhops:3  msatfactor:4.0  running_for:4  count_getroute:10  time_getroute:0.1384265422821045  time_getroute_avg:0.01384265422821045  count_sendpay:2  time_sendpay:1.7709376811981201  time_sendpay_avg:0.8854688405990601
maxhops:3  msatfactor:4.0  running_for:5  count_getroute:11  time_getroute:0.20621562004089355  time_getroute_avg:0.018746874549172142  count_sendpay:3  time_sendpay:2.66377592086792  time_sendpay_avg:0.8879253069559733
```

We can see that:
1. CPU on excessive `getroute` usage isn't a big deal, even on a raspberry pi.
2. The route was found in 5 seconds compared to not found (in 60 seconds)
3. Calling `sendpay` is more expensive timewise by a multiple factor. Exact value cannot be determined by the above data, as the successful route was just 3 hops. We see that `sendpay`takes more time than `getroute`, this becoms more painful on longer routes (which no maxhops filter gives early on)

m-schmoock:
I still need your feedback on line 292 if we want such a filter, what approach would be good here and what the default values can be :D

rebalance/rebalance.py
```diff
@@ -186,39 +233,60 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
...
+                # exclude other erroring channels
                 if erring_channel is not None and erring_direction is not None:
                     excludes.append(erring_channel + '/' + str(erring_direction))
+                # count and exclude nodes that produce a lot of errors
+                if erring_node:
+                    if nodes.get(erring_node) is None:
+                        nodes[erring_node] = 0
+                    nodes[erring_node] += 1
+                    if nodes[erring_node] > 3:
+                        excludes.append(erring_node)
```
gallizoltan, line 284: About this erring node filter, could we insert this line (or something similar) somewhere here?
```
rpc_result["erring_node_count"] = sum(n > 3 for n in nodes.values())
```
So we may see in the logs, how useful this erring node filter is.

rebalance/rebalance.py
```diff
@@ -161,23 +165,66 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
...
+                    msatfactor -= 1
+                    if msatfactor == 0:
+                        # when we reached neutral msat factor:
+                        # increase maxhops and restart with msatfactor
+                        maxhops = maxhops + 1
+                        msatfactor = plugin.msatfactor
+                        continue
+                    # we observed that we only have a ~3% success rate on routes that
+                    # are 8 hops or longer. Hence we cut at 6 (+2 in/out).
+                    if plugin.maxhops > 0 and maxhops > plugin.maxhops:
```
gallizoltan, line 221: Don't `continue` in the `if msatfactor == 0:` case. `maxhops` will increased here, so we should check `if plugin.maxhops > 0 and maxhops > plugin.maxhops:`

gallizoltan (approved):
Looks good! Running on my machine, testing/collecting experiences from the wild! :)
Thank you for your good work!

m-schmoock:
> Thank you for your good work!

Thank you for your time testing and discussing this things :D

gallizoltan:
Today's results:
```
running_for:386  count_getroute:242  time_getroute:81.28449010848999  time_getroute_avg:0.3358863227623553  count_sendpay:146  time_sendpay:302.6004514694214  time_sendpay_avg:2.0726058319823384
running_for:379  count_getroute:239  time_getroute:80.07457756996155  time_getroute_avg:0.33504007351448345  count_sendpay:144  time_sendpay:297.20354747772217  time_sendpay_avg:2.0639135241508484
running_for:472  count_getroute:263  time_getroute:97.50493144989014  time_getroute_avg:0.3707411842201146  count_sendpay:167  time_sendpay:373.37960481643677  time_sendpay_avg:2.235806016864891
running_for:657  count_getroute:661  time_getroute:402.4569990634918  time_getroute_avg:0.608860815527219  count_sendpay:124  time_sendpay:250.68265318870544  time_sendpay_avg:2.0216342999089147
running_for:325  count_getroute:262  time_getroute:90.02857279777527  time_getroute_avg:0.3436205068617377  count_sendpay:118  time_sendpay:233.86849546432495  time_sendpay_avg:1.9819364022400419
running_for:105  count_getroute:77  time_getroute:13.510810375213623  time_getroute_avg:0.17546506980796914  count_sendpay:50  time_sendpay:90.6411337852478  time_sendpay_avg:1.8128226757049561
running_for:78  count_getroute:62  time_getroute:13.014460802078247  time_getroute_avg:0.20991065809803625  count_sendpay:35  time_sendpay:64.31830263137817  time_sendpay_avg:1.8376657894679478
running_for:568  count_getroute:348  time_getroute:151.28918766975403  time_getroute_avg:0.4347390450280288  count_sendpay:190  time_sendpay:414.3168137073517  time_sendpay_avg:2.1806148089860615
running_for:419  count_getroute:297  time_getroute:150.79527306556702  time_getroute_avg:0.5077281921399562  count_sendpay:130  time_sendpay:264.88747549057007  time_sendpay_avg:2.0375959653120774
running_for:633  count_getroute:408  time_getroute:211.87850999832153  time_getroute_avg:0.5193100735252979  count_sendpay:185  time_sendpay:416.7726078033447  time_sendpay_avg:2.2528249070451065
running_for:750  count_getroute:388  time_getroute:191.3870129585266  time_getroute_avg:0.4932654973157902  count_sendpay:222  time_sendpay:554.8151588439941  time_sendpay_avg:2.4991673821801537
running_for:577  count_getroute:352  time_getroute:155.53804540634155  time_getroute_avg:0.4418694471771067  count_sendpay:189  time_sendpay:417.30035614967346  time_sendpay_avg:2.2079383923263145
running_for:836  count_getroute:389  time_getroute:217.8616178035736  time_getroute_avg:0.5600555727598293  count_sendpay:225  time_sendpay:613.6381242275238  time_sendpay_avg:2.7272805521223282
```
Working fine on my machine. 😎

m-schmoock:
Nice. Can you also try a day with this feature disabled by  `rebalance-maxhops: 0` and `rebalance-msatfactor: 1`? I would like to know if it makes things better or worse :D

gallizoltan:
Ok, I set up my config and restarted the `lightningd`.
```
2021-06-04T10:07:16.217Z INFO    plugin-rebalance.py: Plugin rebalance initialized with 1000msat base / 400 ppm fee  cltv_final:18  maxhops:0  msatfactor:1.0
```
Will see what will happen by tomorrow.

rebalance/rebalance.py
gallizoltan: By the way, you should really remove this `continue`. With `rebalance-msatfactor: 1` we never reach the maxhops check.

And the `if msatfactor == 0:` could be `if msatfactor < 1:`, as msatfactor is float.

m-schmoock: Thanks. Fixed by 0e5bf03
I moved the maxhops `if plugin.maxhops > 0 and maxhops > plugin.maxhops`in front of the `msatfactor` check. Sounds correct, right?

m-schmoock: Also please don't mind my stupid brain on  8f5f325  :D
I will sqash in the end ^^

rebalance/rebalance.py
```diff
@@ -161,23 +165,64 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
...
+                    # reduce msatfactor to look for smaller channels now
+                    msatfactor -= 1
+                    if msatfactor < 1:
+                        # when we reached neutral msat factor:
+                        # increase maxhops and restart with msatfactor
+                        maxhops = maxhops + 1
+                        msatfactor = plugin.msatfactor
+                    continue
+                else:
+                    raise e
```
gallizoltan, line 224: Almost perfect. First increase `maxhops`, then check if `maxhops` is too big! Otherwise it will run once with `maxhops = plugin.maxhops + 1`.
```suggestion
            except RpcError as e:
                # could not find route -> change params and restart loop
                if e.method == "getroute" and e.error.get('code') == 205:
                    # reduce msatfactor to look for smaller channels now
                    msatfactor -= 1
                    if msatfactor < 1:
                        # when we reached neutral msat factor:
                        # increase maxhops and restart with msatfactor
                        maxhops = maxhops + 1
                        msatfactor = plugin.msatfactor
                    # abort if we reached maxhop limit
                    if plugin.maxhops > 0 and maxhops > plugin.maxhops:
                        rpc_result = {'status':'error', 'message': 'No suitable routes found'}
                        return cleanup(plugin, label, payload, rpc_result)
                    continue
                else:
                    raise e
```

m-schmoock: Okay, I switched the blocks back again. now just without the 'continue' Guess your right. thanks

gallizoltan:
The latest version works with `rebalance-maxhops: 0` and `rebalance-msatfactor: 1`:
```
running_for:3  count_getroute:1  time_getroute:0.015108346939086914  time_getroute_avg:0.015108346939086914  count_sendpay:1  time_sendpay:2.885270357131958  time_sendpay_avg:2.885270357131958
running_for:3  count_getroute:1  time_getroute:0.015320301055908203  time_getroute_avg:0.015320301055908203  count_sendpay:1  time_sendpay:2.700615167617798  time_sendpay_avg:2.700615167617798
running_for:8  count_getroute:1  time_getroute:0.011235952377319336  time_getroute_avg:0.011235952377319336  count_sendpay:1  time_sendpay:8.138787269592285  time_sendpay_avg:8.138787269592285
running_for:2  count_getroute:1  time_getroute:0.010715961456298828  time_getroute_avg:0.010715961456298828  count_sendpay:1  time_sendpay:1.4145050048828125  time_sendpay_avg:1.4145050048828125
running_for:558  count_getroute:115  time_getroute:5.211904287338257  time_getroute_avg:0.04532090684641962  count_sendpay:115  time_sendpay:552.7141962051392  time_sendpay_avg:4.806210401783819
running_for:579  count_getroute:130  time_getroute:6.348092079162598  time_getroute_avg:0.04883147753201998  count_sendpay:130  time_sendpay:571.9680528640747  time_sendpay_avg:4.399754252800575
running_for:622  count_getroute:199  time_getroute:10.911724090576172  time_getroute_avg:0.05483278437475463  count_sendpay:198  time_sendpay:609.8438556194305  time_sendpay_avg:3.080019472825407
running_for:426  count_getroute:198  time_getroute:11.040110111236572  time_getroute_avg:0.055758131874932185  count_sendpay:197  time_sendpay:414.5556945800781  time_sendpay_avg:2.1043436273100413
running_for:416  count_getroute:197  time_getroute:10.938336849212646  time_getroute_avg:0.055524552534074345  count_sendpay:196  time_sendpay:405.0002086162567  time_sendpay_avg:2.0663275949809017
running_for:641  count_getroute:220  time_getroute:13.2664954662323  time_getroute_avg:0.060302252119237724  count_sendpay:219  time_sendpay:627.2657885551453  time_sendpay_avg:2.8642273450006632
```
The results are actually good. But because of the `rebalance-maxhops: 0` it spends much time trying to find a route with channel pairs, where no fitting route exists.

m-schmoock:
> The results are actually good. But because of the `rebalance-maxhops: 0` it spends much time trying to find a route with channel pairs, where no fitting route exists.

I think we can see pros and cons here:
 - Using this `short and big pipes first` can reduce the time to find a route significantly. Maybe helpful for manual rebalances.
 - For long running `rebalanceall` background tasks we may want to disable these filters, since time isn't an issue.

Maybe we disable this optimization for `rebalanceall` and have maxhops 5 + msatfactor 4 as defaults for manual rebalances, so the user can expect to actually finish a manual rebalance within the default timeout of 60secs  (which is currently an issue).

gallizoltan:
I wouldn't disable maxhops filter for `rebalanceall`. In a real-world situation, we barely have a chance to make a successful rebalance with maxhops > 6.

On the other hand, I would be happy if a `rebalanceall` wouldn't take too long. Ideally a few hours, or maybe a day. My node processes transaction forwards, about 5-10 times a day. This ruins the channel balances. It would be good to run a complete `rebalanceall` multiple times a day.

m-schmoock:
@gallizoltan I now did the following:
 - make the rebalance function have maxhops and msatfactor arguments
 - for `rebalanceall` always use maxhops 6  and msatfactor 1
 - change the order of `maxfeepercent` and `retry_for` as I find it more common on the console to retry with a higher timeout, this way we dont need to name all the arguments (outgoing_scid=... incoming_scid=... retry_for=600) but we can just add a 600 at the end of the last command.

Is this better?

rebalance/rebalance.py
```diff
@@ -418,7 +490,11 @@ def maybe_rebalance_pairs(plugin: Plugin, ch1, ch2, failed_channels: list):
         plugin.log(f"Try to rebalance: {scid1} -> {scid2}; amount={amount}; maxfee={maxfee}")
         start_ts = time.time()
         try:
-            res = rebalance(plugin, outgoing_scid=scid1, incoming_scid=scid2, msatoshi=amount, maxfeepercent=0, retry_for=1200, exemptfee=maxfee)
+            res = rebalance(plugin, outgoing_scid=scid1, incoming_scid=scid2,
+                    msatoshi=amount, retry_for=1200, maxfeepercent=0,
+                    exemptfee=maxfee, maxhops=6, msatfactor=1)
```
gallizoltan, line 495: Maybe you meant this?
```suggestion
            res = rebalance(plugin, outgoing_scid=scid1, incoming_scid=scid2,
                    msatoshi=amount, retry_for=1200, maxfeepercent=0,
                    exemptfee=maxfee, maxhops=plugin.maxhops, msatfactor=plugin.msatfactor)
```

gallizoltan (approved):
Looks good.

By the way, do we know, how do the `getroute` works? If `msatfactor: 1` fixed, is there any point to iterate `maxhops`? If there is a short route with 3 hops, then `getroute(maxhops=6)` will not find it anyway?

m-schmoock:
> If there is a short route with 3 hops, then `getroute(maxhops=6)` will not find it anyway?

Yes it will find shorter routes anyway. Iterating and increasing maxhops will simply force trying shorter routes first at the expense to some CPU cycles. Even without msatfactor this can help a bit since shorter routes are less likely to have an error along the route compared to longer routes. I guess

rebalance/rebalance.py
```diff
@@ -124,22 +130,28 @@ def calc_optimal_amount(out_ours, out_total, in_ours, in_total, payload):
 
 @plugin.method("rebalance")
 def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = None,
-              maxfeepercent: float = 0.5, retry_for: int = 60, exemptfee: Millisatoshi = Millisatoshi(5000)):
+              retry_for: int = 60, maxfeepercent: float = 0.5,
+              exemptfee: Millisatoshi = Millisatoshi(5000),
+              maxhops: int = None, msatfactor: float = None):
```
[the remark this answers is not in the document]
m-schmoock: Nope, unfortunately that does not work :)

rebalance/rebalance.py
m-schmoock: Actually this was by intention, to have the long running (non interactive) rebalances for `rebalanceall` using least aggressive values, while the manual (likely interative) commands using this optimization that tries to give faster results.

What values do we want for `rebalanceall`? certainly not plugin defaults of maxhops 5  and msatfactor 4.

m-schmoock:
@gallizoltan 
Note: In the last commit I made the `erring_node` count configureable. I also increased the default to 5 (or more) errors that a node has to raise.

gallizoltan:
LGTM!
Yes, anything can happen in the wild! It's good to have an opportunity to configure the `erring_node` count.

m-schmoock:
Thanks @gallizoltan 

Note: The CI produces stupid errors unrelated tot the plugin, i.e. on `fundchannel`: `You gave bad parameters: feerate_per_kw 253 below minimum 1875` ??).

Still trying to fix that, clean things up and squash. will merge in once ready.

gallizoltan:
OK!
By the way, do you know how to disable `INFO    gossipd: Minimal possible route ...` messages in the logs?
I've got plenty of them recently. :)

PR #249, opened 2021-06-09
rebalance: interchangeable route search strategy

gallizoltan:


m-schmoock:
@gallizoltan

Something like that would be cleaner, yes.
Without going into details, is there a functional difference in using `getroute_basic` or `msatfactor` 1 and `maxhops` 0?

gallizoltan:
> @gallizoltan
> 
> Something like that would be cleaner, yes.
> Without going into details, is there a functional difference in using `getroute_basic` or `msatfactor` 1 and `maxhops` 0?

@m-schmoock Yes, the functional difference is `maxhops` is not iterative in `getroute_basic`. It will not try `getroute(maxhops=1)` and will not fill my logs with `INFO gossipd: Minimal possible route...` messages.

rebalance/rebalance.py
```diff
@@ -625,14 +664,27 @@ def init(options, configuration, plugin):
...
-               f"erringnodes:{plugin.erringnodes}")
+               f"erringnodes:{plugin.erringnodes} "
+               f"getroute: {plugin.getroute.__name__}")
 
 
+plugin.add_option(
+    "rebalance-getroute",
+    "default",
+    "Getroute method for route search"
+    "Can be 'default' or 'basic' for simple search"
```
m-schmoock, line 685: We should point out the differences here a bit and maybe rename default to iterative. Like:
```
"Can be 'basic' or 'iterative'. "
"'basic': Tries all routes sequentially." 
"'iterative': Tries shorter and bigger routes first."
```

rebalance/rebalance.py
```diff
@@ -128,6 +128,49 @@ def calc_optimal_amount(out_ours, out_total, in_ours, in_total, payload):
...
+class NoRouteException(Exception):
+    pass
+
+
+def getroute_basic(plugin: Plugin, targetid, fromid, excludes, msatoshi: Millisatoshi):
+    try:
+        return plugin.rpc.getroute(targetid,
+                                   fromid=fromid,
+                                   exclude=excludes,
+                                   msatoshi=msatoshi,
```
m-schmoock, line 140: The `getroute_basic` method currently ignores the `plugin.msatfactor_effective`.
I know that's likely by intention, just want to point that our at line 534, it will be set to 1 for `rebalanceall`.

Not sure how you see that, maybe we want to set the `getroute_xyz` method as a parameter to the `rebalance` method. I can see the point in using basic for `rebalanceall`.

m-schmoock (commented):
Hey, Sorry it took a while... :)
Here is my first rounds of thoughts.

rebalance/rebalance.py
gallizoltan: It can be a parameter to the `rebalance` method of course. On the other hand, it usually makes the code cleaner, if we do not use too many parameters for functions. And `rebalance` already has 8 parameters.

m-schmoock: Yes I know. But if we add this as a parameter, we can remove the two "maxhops" and "msatfactor" as they were only added in order to have `rebalanceall` setting other values. And we are down to 7 parameters :)

m-schmoock (changes requested):
Better, still lines 536-538 (`rebalanceall`) needs a change to make use of latest change

gallizoltan:
> Better, still lines 536-538 (`rebalanceall`) needs a change to make use of latest change

Yep, sorry, I forgot that.

rebalance/rebalance.py
```diff
@@ -492,7 +535,7 @@ def maybe_rebalance_pairs(plugin: Plugin, ch1, ch2, failed_channels: list):
         try:
             res = rebalance(plugin, outgoing_scid=scid1, incoming_scid=scid2,
                             msatoshi=amount, retry_for=1200, maxfeepercent=0,
-                            exemptfee=maxfee, maxhops=6, msatfactor=1)
+                            exemptfee=maxfee)
```
m-schmoock, line 538: Do we want `getroute_method=getroute_basic` for this? Long running rebalanceall does not need to be optimized, we want to try all routes because we have time, no?

m-schmoock: Twitter dup: Should we add an `getroute_lowfee` option that tries with lower fees first and then increases when no route found? That could also be awesome for rebalanceall.

gallizoltan: > Do we want `getroute_method=getroute_basic` for this? Long running rebalanceall does not need to be optimized, we want to try all routes because we have time, no?

I'm not sure. This way we have an option to select `getroute_method` for `rebalanceall`.

m-schmoock: The reason we used maxhops 6 and msatfactor 1 here before was to partly disable the 'short and big pipes first's' optimization we added per Default. I suggested this because we have plenty of time when using rebalanceall. 

Using 'first and short routes first' does make sense for manual rebalances where the user expect a successful result within 60seconds, which 'getroute_basic' hardly delivers.

I suggest we add a 'getroute_smallfee' iteration function and use that for rebalance all or go with 'getroute_basic'.

gallizoltan: I think it is good this way right now.
1) Most importantly: `getroute_method` is not hard-coded, it can be modified by config.
2) The default method is `iterative`: it uses big channels first. This is a good thing, not because of the time, but because most likely it will not mess up my neighbors' channels.
3) `getroute_smallfee` is a great idea. This PR makes `getroute` interchangeable, and we can implement `getroute_smallfee` later, in a different PR. Even we can make it the default `getroute_method`.

m-schmoock: Agreed, let me test this in production a bit and then we can merge if there are no issues...

m-schmoock:
Concept ACK

m-schmoock:
tACK

PR #250, opened 2021-06-15
rebalance: nits

m-schmoock:


m-schmoock:
@gallizoltan Nothing special I guess, just a second set of eyes as usual...

I wanted to start with the `getroute_lowfee` method but figured that it gets more complicated and we should use gossmap support and own routing for that i.e. by utilizing something like https://github.com/ElementsProject/lightning/pull/4582/files .
`getroute` does not have a good option to control fees.

Maybe we start with MPP or Gossmap custom routing next...

rebalance/rebalance.py
```diff
@@ -655,7 +663,7 @@ def rebalancestop(plugin: Plugin):
     plugin.mutex.acquire(blocking=True)
     plugin.rebalance_stop = False
     plugin.mutex.release()
-    return {"message": "Rebalance stopped"}
+    return {"message": plugin.rebalanceall_msg}
```
gallizoltan, line 666: Wow! What an idea! This will be really informative!

gallizoltan (approved):
LGTM!

PR #253, opened 2021-06-15
rebalance report

gallizoltan:
- `rebalancereport`: shows information about rebalance
- some README update

### DIFF
--- a/rebalance/rebalance.py
+++ b/rebalance/rebalance.py
@@ -674,7 +674,7 @@ def rebalancereport(plugin: Plugin):
     res["rebalanceall_is_running"] = plugin.mutex.locked()
     res["getroute_method"] = plugin.getroute.__name__
     res["maxhops_threshold"] = plugin.maxhops
-    res["msatfactor_threshold"] = plugin.msatfactor
+    res["msatfactor"] = plugin.msatfactor
     res["erringnodes_threshold"] = plugin.erringnodes
     channels = get_open_channels(plugin)
     if len(channels) > 1:
--- a/rebalance/rebalance.py
+++ b/rebalance/rebalance.py
@@ -666,6 +666,36 @@ def rebalancestop(plugin: Plugin):
     return {"message": plugin.rebalanceall_msg}
 
 
+@plugin.method("rebalancereport")
+def rebalancereport(plugin: Plugin):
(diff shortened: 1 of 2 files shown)


rebalance/rebalance.py
```diff
@@ -666,6 +666,46 @@ def rebalancestop(plugin: Plugin):
...
 
+@plugin.method("rebalancereport")
+def rebalancereport(plugin: Plugin):
+    """Show information about rebalance
+    """
+    res = {}
+    res["rebalanceall_is_running"] = plugin.mutex.locked()
+    res["getroute_method"] = plugin.getroute.__name__
+    res["maxhops_threshold"] = plugin.maxhops
+    res["msatfactor_threshold"] = plugin.msatfactor
```
m-schmoock, line 677: this is not a threshold value ;)  We can name it just `msatfactor` or `capacity_factor`. In the latter case we should also rename it in the plugin remaining code. I think its a better speaking name.

m-schmoock:
Nice :+1:

rebalance/rebalance.py
gallizoltan: I started to write `threshold` everywhere, because I thought the `erringnodes` expression is misleading here. :)

m-schmoock: Yes, the names aren't optimal

rebalance/rebalance.py
```diff
@@ -666,6 +666,46 @@ def rebalancestop(plugin: Plugin):
...
+        try:
+            pay = plugin.rpc.listpays(r["bolt11"])["pays"][0]
+            total_amount += pay["amount_msat"]
+            total_fee += pay["amount_sent_msat"] - pay["amount_msat"]
+        except Exception:
+            res["total_successful_rebalances"] -= 1
+    res["total_rebalanced_amount"] = total_amount
+    res["total_rebalance_fee"] = total_fee
+    if total_amount > Millisatoshi(0):
+        res["average_rebalance_fee_ppm"] = round(total_fee/total_amount*10**6, 2)
```
m-schmoock, line 703: That is nice :D

m-schmoock:
tACK

Good to have this, maybe we can improve on this and add more metrics

PR #255, opened 2021-06-17
Update pyln-testing on CI which broke all tests

cdecker:
The root cause was pyln-testing not matching the c-lightning version, and having the fee estimation buckets not matching.

I also went and added some drive-by fixes.

### DIFF
--- a/.ci/test.py
+++ b/.ci/test.py
@@ -95,6 +95,9 @@ def run_one(p: Plugin) -> bool:
 
     assert pytest_path.exists()
 
+    # Update pyln-testing to master since we're running against c-lightning master.
+    install_pyln_testing(pip_path)
+
     print("Current environment packages:")
     subprocess.check_call(
         [pip_path, 'freeze'],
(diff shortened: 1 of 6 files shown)


vincenzopalazzo (approved):
ack https://github.com/lightningd/plugins/pull/255/commits/904dcd4d2c616fddb2193b0e6e9c4babd1f8fa9f

.ci/test.py
```diff
@@ -134,6 +137,23 @@ def run_one(p: Plugin) -> bool:
         print("##[endgroup]")
 
 
+def install_pyln_testing(pip_path):
+    # Update pyln-testing to master since we're running against c-lightning master.
+    dest = Path('/tmp/lightning')
+    repo = 'https://github.com/ElementsProject/lightning.git'
+    if not dest.exists():
+        subprocess.check_call([
+            'git', 'clone', repo, '/tmp/lightning'
```
m-schmoock, line 146: Nit: instead of using the string `/tmp/lightning` you could use the `dest` variable you already set.

.ci/test.py
```diff
@@ -134,6 +137,23 @@ def run_one(p: Plugin) -> bool:
...
+        ])
+    subprocess.check_call([
+        pip_path, 'install', '-U', f'{dest}/contrib/pyln-testing'
+    ])
+
```
m-schmoock, line 151: Nit: I would remove that empty line

paytest/test_paytest.py
```diff
@@ -41,9 +41,9 @@ def test_mpp_pay(node_factory):
     """ l1 send a payment that is going to be split.
     """
     l1, l2 = node_factory.line_graph(2, opts=pluginopt, wait_for_announce=True)
-    res = l1.rpc.paytest(l2.info['id'], 10**8)
+    res = l1.rpc.paytest(l2.info['id'], 2*10**8)
```
m-schmoock, line 44: Out of curiosity: Is this test more stable with 200000000 ?

m-schmoock (approved):
ACK 904dcd4

Thanks!
GitHub CI is happy. Just pointless nits, you can merge.

paytest/test_paytest.py
cdecker: Just needs enough value to force a split which this does.

PR #256, opened 2021-06-22
feeadjuster: fee strategy per channel

m-schmoock:
This will cleanup some code and adds a configurable per channel fee strategy instead of always using global config or defaults. 

The new default strategy selects the median fees from the other peers of a peer.
The rational behind is that our node somehow competes towards that target.
The idea was raised by @ZmnSCPxj a while ago on IRC.

Using these median fees as base values for further calculation also feedbacks and reflects the current network state at that local area when/if the the local peers also use some dynamic methods.

### DIFF
--- a/feeadjuster/feeadjuster.py
+++ b/feeadjuster/feeadjuster.py
@@ -8,8 +8,9 @@ from threading import Lock
 plugin = Plugin()
 # Our amount and the total amount in each of our channel, indexed by scid
 plugin.adj_balances = {}
-# Cache to avoid loads of calls to getinfo
+# Cache to avoid loads of RPC calls
 plugin.our_node_id = None
+plugin.peers = None
 # Users can configure this
 plugin.update_threshold = 0.05
 # forward_event must wait for init
@@ -60,11 +61,26 @@ def get_ratio_hard(our_percentage):
     return 100**(0.5 - our_percentage) * (1 - our_percentage) * 2
 
 
+def get_peer_id_for_scid(plugin: Plugin, scid: str):
+    for peer in plugin.peers:
+        for ch in peer['channels']:
+            if ch['short_channel_id'] == scid:
+                return peer['id']
+    return None
+
+
+def get_local_channel_for_scid(plugin: Plugin, scid: str):
+    for peer in plugin.peers:
+        for ch in peer['channels']:
+            if ch['short_channel_id'] == scid:
+                return ch
+    return None
+
+
 def get_chan_fees(plugin: Plugin, scid: str):
-    channels = plugin.rpc.listchannels(scid)["channels"]
-    for ch in channels:
-        if ch["source"] == plugin.our_node_id:
-            return {"base": ch["base_fee_millisatoshi"], "ppm": ch["fee_per_millionth"]}
+    channel = get_local_channel_for_scid(plugin, scid)
+    assert channel is not None
+    return {"base": channel["fee_base_msat"], "ppm": channel["fee_proportional_millionths"]}
 
 
 def setchannelfee(plugin: Plugin, scid: str, base: int, ppm: int):
@@ -130,7 +146,7 @@ def maybe_adjust_fees(plugin: Plugin, scids: list):
 
 
 def get_chan(plugin: Plugin, scid: str):
-    for peer in plugin.rpc.listpeers()["peers"]:
+    for peer in plugin.peers:
         if len(peer["channels"]) == 0:
             continue
         # We might have multiple channel entries ! Eg if one was just closed
@@ -158,6 +174,7 @@ def forward_event(plugin: Plugin, forward_event: dict, **kwargs):
     if not plugin.forward_event_subscription:
         return
     plugin.mutex.acquire(blocking=True)
+    plugin.peers = plugin.rpc.listpeers()["peers"]
     if forward_event["status"] == "settled":
         in_scid = forward_event["in_channel"]
         out_scid = forward_event["out_channel"]
@@ -183,9 +200,9 @@ def feeadjust(plugin: Plugin):
     Otherwise, the plugin keeps the fees up-to-date.
     """
     plugin.mutex.acquire(blocking=True)
-    peers = plugin.rpc.listpeers()["peers"]
+    plugin.peers = plugin.rpc.listpeers()["peers"]
     channels_adjusted = 0
-    for peer in peers:
+    for peer in plugin.peers:
         for chan in peer["channels"]:
             if chan["state"] == "CHANNELD_NORMAL":
                 scid = chan["short_channel_id"]
--- a/feeadjuster/feeadjuster.py
+++ b/feeadjuster/feeadjuster.py
@@ -64,13 +64,12 @@ def get_chan_fees(plugin: Plugin, scid: str):
     channels = plugin.rpc.listchannels(scid)["channels"]
     for ch in channels:
         if ch["source"] == plugin.our_node_id:
-            return {"base_fee_millisatoshi": ch["base_fee_millisatoshi"],
-                    "fee_per_millionth": ch["fee_per_millionth"]}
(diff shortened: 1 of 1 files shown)


darosior:
I'm not sure to get in what sense a node competes with its peers' channels, they are rather more likely to be in the same route? (And to compete, as a team, against all the other possible routes on the network :p)

m-schmoock:
> I'm not sure to get in what sense a node competes with its peers' channels, they are rather more likely to be in the same route? (And to compete, as a team, against all the other possible routes on the network :p)

The rational is that if our peer can only be reached with high (or low) fees we should adapt to that:
- If the peer is the target (which of course we can't know) the rationale is obvious.
- If the peer is not the target, others will increase or decrease the likelihood of choosing us along the route if they can reflect that by being able to detect high or low fees local areas within the network.

m-schmoock:
Rebased to see if CI is now happy

feeadjuster/feeadjuster.py
```diff
@@ -189,9 +224,10 @@ def feeadjust(plugin: Plugin):
     Otherwise, the plugin keeps the fees up-to-date.
     """
     plugin.mutex.acquire(blocking=True)
-    peers = plugin.rpc.listpeers()["peers"]
+    plugin.peers = plugin.rpc.listpeers()["peers"]
+    plugin.channels = plugin.rpc.listchannels()['channels']
```
gallizoltan, line 228: This will contain all the channels in the Lightning network? The `listchannels` currently returns 85039 channels for me. How much memory will this consume?

feeadjuster/feeadjuster.py
```diff
@@ -305,4 +346,13 @@ def init(options: dict, configuration: dict, plugin: Plugin, **kwargs):
...
 )
+plugin.add_option(
+    "feeadjuster-feestrategy",
+    "median",
+    "Sets the per channel fee selection strategy. "
+    "Can be 'median' to use the median fees from peers of peer "
+    "or 'global' to use global config or default values. "
+    "Default: 'median'.",
+    "string"
+)
```
gallizoltan, line 357: Could we use 'global' for default value?

gallizoltan (commented):
I'm not sure about this. Usually I don't want to compete with other nodes, because I can directly forward payments to my neighbors, but other nodes most likely cannot.
So my node is actually uncomparable, in this mean. :)

Could we use 'global' as the default payment strategy?

And if the payment strategy is 'global', could we avoid initiating `plugin.channels`? To save some memory.

feeadjuster/feeadjuster.py
m-schmoock: yep, I can change it to only do this if `median` is used

feeadjuster/feeadjuster.py
m-schmoock: Sure we can, wanted your opinion on this approach

m-schmoock:
> I'm not sure about this. Usually I don't want to compete with other nodes, because I can directly forward payments to my neighbors, but other nodes most likely cannot.
> So my node is actually uncomparable, in this mean. :)

Can you elaborate a bit on this? 
 - How do you choose your current global fee settings?
 - Does it make sense to use the same values for all channels?

gallizoltan:
> > I'm not sure about this. Usually I don't want to compete with other nodes, because I can directly forward payments to my neighbors, but other nodes most likely cannot.
> > So my node is actually uncomparable, in this mean. :)
> 
> Can you elaborate a bit on this?
> 
> * How do you choose your current global fee settings?
> * Does it make sense to use the same values for all channels?

I've chosen my fee settings by experimenting. I did not want to see too many transactions forwards, because the forwards ruin my liquidity, and if my liquidity is ruined, the forwards start failing.
So I raised my fees until I saw the number of forwards decreased to a reasonable number.

And I use `rebalanceall` regularly. `rebalanceall` uses my default fee setting for rebalances, to fix my liquidity. I raised my fees until I saw `rebalanceall` can fix my liquidity well enough.

My current fees in the config:
```
fee-base=1000
fee-per-satoshi=400
```
I did not use the same fee for all channels. I use feeadjuster in an old-fashioned way: the feeadjuster raises my fees for channels where the liquidity is too low, and lowers the fees for channels where the liquidity is too high.

I think cyberspace is inherently different from meatspace: in the real world, I would assume, two of my neighbors surely know each other, because we live in the same neighborhood. And if I am too greedy, I may be left out of the garden party.
The Lightning Network is different: it has 12,227 nodes with active channels, and an average node has only 8.40 channels. Stats from here: https://1ml.com/statistics
My point is: two of my neighbors usually do not know each other. So no need to compete, because most likely I am the only mutual friend. They have a slim chance to find a working route without me, and that route would contain much more hops.

m-schmoock:
@gallizoltan
Thx
- Do you see some potential benefits in adjusting the fees on a per channel basis?
- If yes, what approach can you imagein?

gallizoltan:
> @gallizoltan
> Thx
> 
> * Do you see some potential benefits in adjusting the fees on a per channel basis?
> * If yes, what approach can you imagine?

Currently I would rather not use it. But of course, you can merge and start to use it. And in time, or based on your experiences, we may figure out later, what advantages this has. Or you mentioned @ZmnSCPxj, maybe he can tell some practical benefits.

m-schmoock:
@gallizoltan 
Can you ACK code-wise?

gallizoltan (approved):
ACK, looks good to me!

ZmnSCPxj:
One *is* still competing with the peers of your peer.

While your position on the network is unique, in a sense, it is also the same as others on the network --- just as the peers of your peer probably can be accessible on the network, *you* are reachable on the network.

While there are nodes that will be nearer to the peers of your peer rather than you, due to capacity issues, at *some* point even *those* nodes will find that the shortest path to your competitors is unusable, and *then* the best *usable* path to the peer of your peer is the same distance (fee-wise) as to you.  Then, in order to reach your peer, the payer has to select between you, or the competing peer of your peer.  Having at *least* a similar feerate as the peer of your peer keeps you in the competition, if your fee is much higher than you will not be forwarded through.  Thus, even though your position on the network is "unique", it is also still "the same" as that of your peers, because the network state is also changing.  It makes sense to look at what others are doing and copy it.  Thus, median.

Rebalancing periodically was also what CLBOSS used to do quite a bit of, but CLBOSS also recorded how much it earned and how much it spent on rebalances, and I found to my chagrin that too much rebalancing often lead to rebalancing spending far outstripping fee earnings.  Thus, recent CLBOSS versions tone down their rebalancing significantly.

CLBOSS also now uses "price theory".  Basically, the theory is that there is a global optimal channel feerate, without any local optima, where you earn the most.  If the price is higher than the global optimal price, the number of customers goes down and earnings are down (because even though your margin is higher, you get fewer customers).  If the price is lower than the global optimal price, the number of customers goes higher, but your margins are too low and you earn too little.  Thus, there is some global optimal price where the number of customers times your earning margins are the highest.  CLBOSS *starts* at the median of the peers of your peer ("it makes sense to look at what others are doing and copy it"), but it also experimentally moves the feerate to variants around its current center price, and recording what it earns at each price point (in time periods of 2880 minutes or thereabouts), then after some time, it looks at which price point earned it most and then moves its center to that price point, then starts a new experiment on the variants around the new price point, ad infinitum, on the basis of this price theory implying *some* global optima exists.

m-schmoock:
@ZmnSCPxj @gallizoltan 
I wonder if we can/should add a feature that takes into account the fees our peers use towards our node.
What I mean by that is that peers can't trick us too easily on our fees, i.e:

1. Our node forwards to a peer with what we currently assume to be a reasonable fee. 
2. Our peers balance goes up.
3. Our peer uses or sticks to too high fees in our direction back that our channel never balances naturally.
4. Eventually we do a manual rebalance and he collects the fees.

ZmnSCPxj:
@m-schmoock  that seems a reasonable logic, however do note that this plugin already adjust fees such that if the balance is mostly to their side, we already bump our own fees.  So naively, it seems we already get a mitigation against that.

PR #257, opened 2021-06-23
rebalance: measure liquidity health

gallizoltan:
`rebalancereport` shows a score that reflects liquidity health
* its value between 0% and 100% depending on liquidity distribution
* running `rebalanceall` may increase the score, and surely not decrease (if no forward event occurs meanwhile)

### DIFF
--- a/rebalance/rebalance.py
+++ b/rebalance/rebalance.py
@@ -348,11 +348,6 @@ def a_minus_b(a: Millisatoshi, b: Millisatoshi):
     return a - b if a > b else Millisatoshi(0)
 
 
-def a_div_b(a: Millisatoshi, b: Millisatoshi):
-    # a divided by b, but don't divide by 0
-    return int(a) / int(b) if int(b) > 0 else 1
-
-
 def must_send(liquidity):
     # liquidity is too high, must send some sats
     return a_minus_b(liquidity["min"], liquidity["their"])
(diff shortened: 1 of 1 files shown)


rebalance/rebalance.py
```diff
@@ -348,6 +348,11 @@ def a_minus_b(a: Millisatoshi, b: Millisatoshi):
     return a - b if a > b else Millisatoshi(0)
 
 
+def a_div_b(a: Millisatoshi, b: Millisatoshi):
+    # a divided by b, but don't divide by 0
+    return int(a) / int(b) if int(b) > 0 else 1
```
m-schmoock, line 353: This method is used to calculate the distortion of channel liquidity.
If `b` is set to `0` it means an extremely distorted scenario. 
Is returning `1` (neutral) a good approach here?

Also, in the code below `b` is given as `liquidity["ideal"]["our"]`, `...["their"]` and `liquidity["min"]`.
Should it be possible/normal these values can be `0`?
If not we can (and should) remove that check and replace it by a check in the calling function.

Also, also, if that check is removed, this function becomes pointless ;)

Also, also, also using recent pyln versions des not require int casts for multiplication and division, maybe we should change `requirements.txt` to recent version and cleanup the code where this also applies.

rebalance/rebalance.py
```diff
@@ -398,9 +403,9 @@ def check_liquidity_threshold(channels: list, threshold: Millisatoshi):
 
 
 def binary_search(channels: list, low: Millisatoshi, high: Millisatoshi):
-    if high - low < Millisatoshi("1sat"):
-        return low
     next_step = (low + high) / 2
+    if high - low < Millisatoshi("1sat"):
+        return next_step
     if check_liquidity_threshold(channels, next_step):
         return binary_search(channels, next_step, high)
```
m-schmoock, line 410: Unrelated to this PR:
Python does not support proper tail recursions. Maybe we should rewrite this to a loop.

Update:
If I see this correct, for  the biggest channel being i.e. `0.05btc` this adds up to 23 recursions until the value goes below 1sat. Not sure if thats a big deal or not, likely not a big deal :D

m-schmoock (commented):
My two thoughts.
I will test this and then ACK later

rebalance/rebalance.py
gallizoltan: In a distorted scenario `liquidity["ideal"]["their"]` can be zero. For example, if someone just started to run a node, opens some channels, and all liquidity belongs to him.

But yes, I will replace this with a check in the calling function.

m-schmoock: If they can be 0 legitimately you don't need to move that check to the caller.
Feel free to choose.

m-schmoock:
LGTM and tested.

Lets try githubs "Squash and merge" button :D

PR #259, opened 2021-06-24
feeadjuster: use listchannels by destination

m-schmoock:
This will add a runtime check to see if the daemon already supports using `lightning-cli listchannels destination=...` feature https://github.com/ElementsProject/lightning/pull/4614 .
If successful this will be less resource intense.

### DIFF
--- a/feeadjuster/feeadjuster.py
+++ b/feeadjuster/feeadjuster.py
@@ -289,11 +289,16 @@ def init(options: dict, configuration: dict, plugin: Plugin, **kwargs):
     if plugin.imbalance > 0.5:
         plugin.imbalance = 1 - plugin.imbalance
 
-    plugin.log(f"Plugin feeadjuster initialized ({plugin.adj_basefee} base / {plugin.adj_ppmfee} ppm) with an "
+    plugin.log(f"Plugin feeadjuster initialized "
+               f"({plugin.adj_basefee} base / {plugin.adj_ppmfee} ppm) with an "
                f"imbalance of {int(100 * plugin.imbalance)}%/{int(100 * ( 1 - plugin.imbalance))}%, "
-               f"update_threshold: {int(100 * plugin.update_threshold)}%, update_threshold_abs: {plugin.update_threshold_abs}, "
-               f"enough_liquidity: {plugin.big_enough_liquidity}, deactivate_fuzz: {plugin.deactivate_fuzz}, "
-               f"forward_event_subscription: {plugin.forward_event_subscription}, adjustment_method: {plugin.get_ratio.__name__}")
+               f"update_threshold: {int(100 * plugin.update_threshold)}%, "
+               f"update_threshold_abs: {plugin.update_threshold_abs}, "
+               f"enough_liquidity: {plugin.big_enough_liquidity}, "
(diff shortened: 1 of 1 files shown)


m-schmoock:
Change to checking the `rpc.help()` output instead of expecting an exception

feeadjuster/feeadjuster.py
```diff
@@ -96,9 +96,12 @@ def get_fees_median(plugin: Plugin, scid: str):
     """
     peer_id = get_peer_id_for_scid(plugin, scid)
     assert peer_id is not None
+    if plugin.listchannels_by_dst:
+        plugin.channels = plugin.rpc.call("listchannels",
+                                          {"destination": peer_id})['channels']
```
gallizoltan, line 101:
```suggestion
    if plugin.channels == None:
        if plugin.listchannels_by_dst:
            plugin.channels = plugin.rpc.call("listchannels",
                                              {"destination": peer_id})['channels']
        else:
            plugin.channels = plugin.rpc.listchannels()['channels']
```

feeadjuster/feeadjuster.py
```diff
@@ -226,7 +229,7 @@ def feeadjust(plugin: Plugin):
     """
     plugin.mutex.acquire(blocking=True)
     plugin.peers = plugin.rpc.listpeers()["peers"]
-    if plugin.fee_strategy == get_fees_median:
+    if plugin.fee_strategy == get_fees_median and not plugin.listchannels_by_dst:
         plugin.channels = plugin.rpc.listchannels()['channels']
```
gallizoltan, line 233:
```suggestion
    plugin.channels = None
```

feeadjuster/feeadjuster.py
```diff
@@ -198,7 +201,7 @@ def forward_event(plugin: Plugin, forward_event: dict, **kwargs):
         return
     plugin.mutex.acquire(blocking=True)
     plugin.peers = plugin.rpc.listpeers()["peers"]
-    if plugin.fee_strategy == get_fees_median:
+    if plugin.fee_strategy == get_fees_median and not plugin.listchannels_by_dst:
         plugin.channels = plugin.rpc.listchannels()['channels']
```
gallizoltan, line 205:
```suggestion
    plugin.channels = None
```

gallizoltan (approved):
LGTM!

Just an idea, doesn't have to be this way:
What if all `plugin.channels` initiation would be in one place, in `get_fees_median`? The old initiations in `forward_event` and `feeadjust` could be just `plugin.channels = None` to ensure, on every occasion, it gets the fresh channels' state. I show you in the comments, what I mean.

feeadjuster/feeadjuster.py
m-schmoock: Well, incase we use listchannels_by_dst we need to set it to None for the next channel after calculating the median fees, otherwise it will use an incorrect list (not for that next channel).

Sidenote: I already figured for ~30 channels on an raspberry pi this takes about 15secs for the initial loop when calling `listchannels(destinaion)` on a per channel basis. But thats not much difference compared to saving all channels once and filtering in python for each channel.

m-schmoock:
Thx. I see what you mean, imho makes the code less readable.

PR #260, opened 2021-06-25
monitor: fix flakyness

m-schmoock:
Sometimes a channels funding allocation was not yet readable
via listpeers which is why this plugin test was flaky.

Also makes use of funding_msat instead of deprecated fields.


vincenzopalazzo (approved):
ack https://github.com/lightningd/plugins/pull/260/commits/e3c5f708842c9e69f8d2ca42fa0c3cdb11d24c45

PR #262, opened 2021-06-26
feeadjuster: fix initial feeadjustment loops

m-schmoock:
I discovered some unexpected behavior in the initial fee adjustment loops.
This currently just add the xfail test. We can add proper fix in this PR as we go...

feeadjuster/test_feeadjuster.py::test_initial_updates

------

Update: I added the plugin `channel_state_changed` notification. When a channel changes to `CHANNELD_NORMAL` this will now also trigger a `maybe_adjust_fees`.

Update2: I fixed remaining tests but it seems like I discovered a bug in clightning's code. When a channels gossip gets updated directly after turning to `CHANNELD_NORMAL` (i.e. by `setchannelfee` called from a `channel_state_changed_hook`) the update first gets (correctly) delayed, but then gets missing so that the gossip is actually never updated. I marked the affected testcode lines with a FIXME comment. The two testcases currently fail just because of that.

### DIFF
--- a/feeadjuster/feeadjuster.py
+++ b/feeadjuster/feeadjuster.py
@@ -234,8 +234,10 @@ def feeadjust(plugin: Plugin):
     channels_adjusted = 0
     for peer in plugin.peers:
         for chan in peer["channels"]:
-            if chan["state"] == "CHANNELD_NORMAL":
-                scid = chan["short_channel_id"]
+            if chan["state"] in ["CHANNELD_NORMAL", "CHANNELD_AWAITING_LOCKIN"]:
+                scid = chan.get("short_channel_id")
+                if scid is None:
+                    continue
                 plugin.adj_balances[scid] = {
                     "our": int(chan["to_us_msat"]),
                     "total": int(chan["total_msat"])
--- a/feeadjuster/feeadjuster.py
+++ b/feeadjuster/feeadjuster.py
@@ -196,6 +196,19 @@ def maybe_add_new_balances(plugin: Plugin, scids: list):
             }
 
 
+@plugin.subscribe("channel_opened")
+def channel_opened_event(plugin: Plugin, channel_opened: dict, **kwargs):
+    plugin.peers = plugin.rpc.listpeers()["peers"]
+    for peer in [p for p in plugin.peers if p['id'] == channel_opened['id']]:
+        for chan in [c for c in peer["channels"] if c['funding_txid'] == channel_opened['funding_txid']]:
+            if chan["state"] in ["CHANNELD_NORMAL", "CHANNELD_AWAITING_LOCKIN"]:
+                scid = chan.get("short_channel_id")
+                if scid is None:
+                    continue
+                maybe_add_new_balances(plugin, [scid])
+                maybe_adjust_fees(plugin, [scid])
+
+
 @plugin.subscribe("channel_state_changed")
 def channel_state_changed_event(plugin: Plugin, channel_state_changed: dict, **kwargs):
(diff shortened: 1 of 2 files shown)


gallizoltan (approved):
LGTM!

feeadjuster/feeadjuster.py
```diff
@@ -234,8 +262,10 @@ def feeadjust(plugin: Plugin):
     channels_adjusted = 0
     for peer in plugin.peers:
         for chan in peer["channels"]:
-            if chan["state"] == "CHANNELD_NORMAL":
-                scid = chan["short_channel_id"]
+            if chan["state"] in ["CHANNELD_NORMAL", "CHANNELD_AWAITING_LOCKIN"]:
+                scid = chan.get("short_channel_id")
```
gallizoltan, line 266: `CHANNELD_AWAITING_LOCKIN` is here because of a clightning bug? So this meant to be temporal till the bug gets fixed?

gallizoltan (approved):

feeadjuster/feeadjuster.py
m-schmoock: Actually this was more like testing. `channel_state_changed` will not trigger for new channels (meaning new_state `CHANNELD_AWAITING_LOCKIN` and old_state `UNKNOWN???`. We need to use the `channel_opened` notification for this, buuuut when a channel is new (not locked in) we don't have a `scid` but only the `funding_txid`.

Ideally we call `setchannelfee` when a new channel it is still in `CHANNELD_AWAITING_LOCKIN`, because then we don't send two updates shortly after one another, one with default fees, and then the modified fees (which can cause problems for the gossip network). But in order to be able to do this we need to wait for `channel_opened` notification for which we only have `funding_txid` and no `scid`. So we need to change more code of this plugin so it deals with `funding_txid` instead of `scid`.

I will convert this to draft until this is fiddled out. But still good to have some discussion about this...

rustyrussell:
> Update2: I fixed remaining tests but it seems like I discovered a bug in clightning's code. When a channels gossip gets updated directly after turning to `CHANNELD_NORMAL` (i.e. by `setchannelfee` called from a `channel_state_changed_hook`) the update first gets (correctly) delayed, but then gets missing so that the gossip is actually never updated. I marked the affected testcode lines with a FIXME comment. The two testcases currently fail just because of that.

I cannot reproduce this, BTW, in c-lightning master, using the following test (though very slow under !DEVELOPER).

```
def test_setchannelfee_delayed(node_factory, bitcoind):
    l1, l2 = node_factory.line_graph(2)

    l1.rpc.setchannelfee("all", 0xDEAD, 0xBEEF)
    l1.daemon.wait_for_log('delaying .* secs')

    wait_for(lambda: only_one(l2.rpc.listchannels(source=l1.info['id'])['channels'])['base_fee_millisatoshi'] == 0xDEAD)
```

However, you may be hitting the other effect: we ratelimit everyone's updates, *including* our own, when we process them (we allow 1 per day average, with a burst of up to 4 per day).  That's more strict than this, and can cause us to throw away our own updates!

Perhaps we should give our own updates a free pass?

chrisguida:
Needs rebase, `feeadjuster` has been removed from the actively maintained plugins list to make the CI pass again, feel free to add it back :)

chrisguida:
Feeadjuster has been added back to the maintained plugins and has had [a fairly large cleanup](https://github.com/lightningd/plugins/pull/525) merged as well. Closing as this PR is probably not relevant anymore.

Please feel free to reopen if it is :)

PR #263, opened 2021-06-29
rebalance: rebalancestop also return last runs message

m-schmoock:
Will make `rebalancestop` also return the last `rebalanceall_msg` event if there was nothing to stop.
This is helpful as `rebalanceall` jobs tend to run very long in the background. This way
the user can still get the result of that job.

### DIFF
--- a/rebalance/rebalance.py
+++ b/rebalance/rebalance.py
@@ -657,7 +657,10 @@ def rebalancestop(plugin: Plugin):
     """It stops the ongoing rebalanceall.
     """
     if not plugin.mutex.locked():
-        return {"message": "No rebalance is running, nothing to stop"}
+        if plugin.rebalanceall_msg is None:
+            return {"message": "No rebalance is running, nothing to stop."}
+        return {"message": f"No rebalance is running, nothing to stop. "
+                           f"Last 'rebalanceall' gave: {plugin.rebalanceall_msg}"}
     plugin.rebalance_stop = True
     plugin.mutex.acquire(blocking=True)
     plugin.rebalance_stop = False
@@ -736,6 +739,7 @@ def init(options, configuration, plugin):
     plugin.msatfactor = float(options.get("rebalance-msatfactor"))
     plugin.erringnodes = int(options.get("rebalance-erringnodes"))
     plugin.getroute = getroute_switch(options.get("rebalance-getroute"))
+    plugin.rebalanceall_msg = None
 
     plugin.log(f"Plugin rebalance initialized with {plugin.fee_base} base / {plugin.fee_ppm} ppm fee  "
                f"cltv_final:{plugin.cltv_final}  "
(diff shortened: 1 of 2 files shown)


rebalance/rebalance.py
```diff
@@ -719,7 +722,7 @@ def rebalancereport(plugin: Plugin):
     res["total_rebalanced_amount"] = total_amount
     res["total_rebalance_fee"] = total_fee
     if total_amount > Millisatoshi(0):
-        res["average_rebalance_fee_ppm"] = round(total_fee/total_amount*10**6, 2)
+        res["average_rebalance_fee_ppm"] = round(total_fee / total_amount * 10**6, 2)
```
gallizoltan, line 725: How do you set up/run flake8? I've checked this before committing this code, but it did not say anything to me.

gallizoltan (approved):
Looks good to me!

rebalance/rebalance.py
m-schmoock: I have a file `ln_setup.sh` which I source when developing lightning an plugins.
It sets various aliases, in this case:

```
alias ln_check_python="flake8 --ignore=E501,E731,W503,W504"
```

Then I just use i.e. `ln_check_python rebalance/*.py` when I made changes in rebalance.

-------------

Also you might love that:

```
# tests a specified file and optional testcase
function ln_test() {
    if [ -z $1 ]; then
        echo "Usage: needs at least a pyton testfile as first argument or optionally a specific testcase name as second argument or both of them concatenated with '::'"
        exit 1
    fi
    if [[ "$1" =~ "::" ]]; then
        set -- $(echo "$1" | awk '{ gsub(/::+/, "\n" ); print; }' | head -n1) $(echo "$1" | awk '{ gsub(/::+/, "\n" ); print; }' | tail -n1)
    fi
    if [ -z $2 ]; then
        py.test -n 8 -d -v $1
    else
        py.test -n 8 -d -v $1 -k $2
    fi
}

# run 50 times on ${MAKE_PAR} threads
function ln_test_flaky() {
    if [ -z $1 ]; then
        echo "Usage: needs at least a pyton testfile as first argument or optionally a specific testcase name as second argument or both of them concatenated with '::'"
        exit 1
    fi
    if [[ "$1" =~ "::" ]]; then
        set -- $(echo "$1" | awk '{ gsub(/::+/, "\n" ); print; }' | head -n1) $(echo "$1" | awk '{ gsub(/::+/, "\n" ); print; }' | tail -n1)
    fi
    if [ -z $2 ]; then
        py.test --count 100 -x -n ${MAKE_PAR} -v $1
    else
        py.test --count 100 -x -n ${MAKE_PAR} -v $1 -k $2
    fi
}

# Converts Python dict outputs in a way that JSON jq likes it 
function ln_jsonify() {
     echo $1 | sed  "s/'/\"/g" | sed "s/False/false/g" | sed "s/True/true/g" | sed "s/None/null/g" | sed "s/\([^\"]\)\([0-9]\+msat\)\([^\"]\)/\1\"\2\"\3/gp" | jq
}
```

m-schmoock:
F.y.i Just "noise" plugin test flakes. Will merge as this is rather trivial

rebalance/rebalance.py
gallizoltan: Wow, thank you, I will try these!

PR #271, opened 2021-07-12
rebalance: fix example usage in readme

billygarrison:
The parameters were in incorrect order, causing an error when user doesn't use with `-k`


LightningHelper:
This is correct.
@m-schmoock ?

rustyrussell (approved):
Yep, this matches the source.

m-schmoock:
Whoops. Forgot about that since I was in vacation :)

You are right, I changed the order a while ago and forgot about the doc.

PR #273, opened 2021-07-22
exposed a method to cancel all outstanding test plugins

renepickhardt:
especially uncertain about line 240
because I don't know what actually happens on a protocol level if one tries to cancel an htlc that has already been canceled. would be bad if the channel gets failed. it wasn't clearly to me specified at: https://github.com/lightningnetwork/lightning-rfc/blob/master/02-peer-protocol.md#removing-an-htlc-update_fulfill_htlc-update_fail_htlc-and-update_fail_malformed_htlc

if that was a problem anyway we would still have race condition problems if the call comes roughly at the same time when the timeout timer triggers anyway

### DIFF
--- a/paytest/paytest.py
+++ b/paytest/paytest.py
@@ -240,10 +240,10 @@ def timeout(plugin, secret):
     del plugin.pending[secret]
         
         
-@plugin.method("timeoutallpendingtestpays")
-def timeoutallpendingtestpays(plugin, **kwargs):
+@plugin.method("rejectpaytests")
+def rejectpaytests(plugin, **kwargs):
     for secret in plugin.pending.keys():
-        timout(plugin,secret)
+        timeout(plugin,secret)
 
 @plugin.async_hook("htlc_accepted")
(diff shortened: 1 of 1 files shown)


renepickhardt:
this method does not work! in the current form as it triggers this exception: 

```
INFO    plugin-paytest.py:   File \"/home/rpickhardt/lnplugins/plugins/paytest/paytest.py\", line 245, in rejectpaytests
INFO    plugin-paytest.py:     for secret in plugin.pending.keys():
INFO    plugin-paytest.py: RuntimeError: dictionary changed size during iteration
```

the `timout` function is killing the secrete I have a fix will update the PR soonish

cdecker:
@Mergifyio rebase

chrisguida:
Needs rebase, `paytest` has been removed from the actively maintained plugins list to get CI running again, but feel free to add it back :)

chrisguida:
Closing this as `paytest` is [no longer maintained](https://github.com/lightningd/plugins/pull/484). Feel free to resurrect it by opening a PR that moves it out of archived :)

PR #275, opened 2021-07-23
datastore: plugin to replicate datastore commands before next release.

rustyrussell:
See https://github.com/ElementsProject/lightning/pull/4674
(I manually tested that it passes all the tests there, too!)

When they finally upgrade the node, this automatically puts any
datastore data into the inbuilt datastore and shuts down.

### DIFF
--- a/datastore/README.md
+++ b/datastore/README.md
@@ -1,0 +1,47 @@
+# Datastore Plugin
+
+The next version of c-lightning (0.10.2?) has a built-in datastore.
+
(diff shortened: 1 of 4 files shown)


vincenzopalazzo (approved):
ack https://github.com/lightningd/plugins/pull/275/commits/4cfdf899cf5ef20a0d20aefdfc26990106932997

The only thing that I don't like is the big if-else, I preferer to use a map of function or something like that. I can try to propose somethings after. But it LGTM.

PR #276, opened 2021-07-23
rebalance, feeadjuster, drain, jitrebalance: fix for next release.

rustyrussell:
We now insist on payment_secret, so rebalance fails!


m-schmoock:
Thanks, we also need to apply this for: drain, feeadjuster (tests), jitrebalance and probe.

m-schmoock:
@rustyrussell The `payment_secret` was added in https://github.com/ElementsProject/lightning/commit/3f5d5a1de , so this change will not run older clightning version. I will add a PR that addresses this backwards compatible also for the other plugins

m-schmoock (approved):
tACK d96e84f

Successfully tested with versions lower and higher than https://github.com/ElementsProject/lightning/commit/3f5d5a1de that have and have not the `paymanet_secret` in the invoice. This works for both because `payment_secret` is an optional parameter for `sendpay` and can be (intentionally) `None` if it was not contained in the invoice before.

rustyrussell:
Good catch, did those too!

PR #280, opened 2021-07-25
commando: reckless?  Try going commando!

rustyrussell:


carnhofdaki:
```
$ pwd
~/src/plugin
$ git rev-parse HEAD
e61c70645624317723fe7a131ec48424bcbb0da9
$ lightning-cli plugin start commando/commando.py commando_reader=024b9a1fa8e006f1e3937f65f66c408e6da8e1ca728ea43222a7381df1cc449605 # this node is BLUEIRON
{
   "code": -32602,
   "message": "Extra parameters must be in object"
}
```

commando/README.md
```diff
@@ -0,0 +1,27 @@
...
+## Options:
+
+Each of these can be specified more than once:
+
+* --commando-reader: a node id which can execute `list` and `get` commands
+* --commando-writer: a node id which can execute any commands.
+
+## Example Usage
+
+$ l1-cli plugin start commando.py commando_reader=0266e4598d1d3c415f572a8488830b60f7e744ed9235eb0b1ba93283b315c03518
```
carnhofdaki, line 23: @rustyrussell This does not work.

carnhofdaki (changes requested):

commando/commando.py
```diff
@@ -0,0 +1,156 @@
...
+    """Messages are form [8-byte-id][data]"""
+    msg = (msgtype.to_bytes(2, 'big')
+           + idnum.to_bytes(8, 'big')
+           + bytes(contents, encoding='utf8'))
+    plugin.rpc.call(plugin.msgcmd, {'node_id': peer_id, 'msg': msg.hex()})
+
+
+def send_result(plugin, peer_id, idnum, res):
+    # We can only send 64k in a message.
+    parts = textwrap.wrap(json.dumps(res), 65000)
```
FreePietje, line 52: 65000 seems like a *magic number* and in the test is another *magic number*, but then it's 65535.
The latter is close to `64*1024=65536`, but not the same.
Maybe use a constant and do `CONSTANT -1` where appropriate (with comment)?

commando/README.md
rustyrussell: Thanks, you're right.  It needs to be a keyword-based argument list, like so:

```
$ l1-cli plugin subcommand=start plugin=`pwd`/commando.py commando_reader=0266e4598d1d3c415f572a8488830b60f7e744ed9235eb0b1ba93283b315c03518
```

(And path to commando.py needs to be absolute).

commando/commando.py
rustyrussell: I will define it.  The entire packet can be 65535, but there's overhead in the message itself.

Thanks!

rustyrussell:
Thanks, feedback addressed, and I've added rune support!

carnhofdaki:
@rustyrussell Thank you! Now my node (IRATEMONKEY) listens to read commando commands from the current BLUEIRON (024b9…) hoping that is still your node ;) anyway

rustyrussell:
> @rustyrussell Thank you! Now my node (IRATEMONKEY) listens to read commando commands from the current BLUEIRON (024b9…) hoping that is still your node ;) anyway

Hmm, are you still running it?  I don't get a response...

carnhofdaki:
@rustyrussell Yes, still running. There are a few things worth mentioning:

 1. My systems is based on musl-libc (Alpine Linux) and so I used e61c706 because it worked and did not have strange requirements like in f914ad1. The same issue with c27e59d\*.
 2. This is the script I use to set-up commando and no idea how I would include all that options in the `config` file of `lightningd`:
```shell
#!/bin/sh

cd ~/src/plugins
git rev-parse --short HEAD
lightning-cli plugin stop $PWD/commando/commando.py
lightning-cli plugin \
  subcommand=start \
  plugin=$PWD/commando/commando.py \
  commando_reader=024b9a1fa8e006f1e3937f65f66c408e6da8e1ca728ea43222a7381df1cc449605
```

\* requirements issue with sha256:
```
sha256.c:3077:25: warning: '_PyUnicode_get_wstr_length' is deprecated [-Wdeprecated-declarations]
```

I see that `hashlib` can not be used as you need midstate access… hmm. Running the old e61c706 which is the only one that works for me so far.

PR #284, opened 2021-08-05
autopilot.py: Moved network initialization to new function to fix issues with timeout

bit0fun:
Initialization for the autopilot plugin would time out due to how long it would take to execute, and lightningd would kill the plugin before finishing. This commit moves the graph calculation routine out of the initialization function to execute after the plugin was initialized. Initialization only reads in options passed from config file or on command line when lightningd was run.

### DIFF
--- a/autopilot/autopilot.py
+++ b/autopilot/autopilot.py
@@ -1,8 +1,8 @@
 #!/usr/bin/env python3
 
-from bech32 import bech32_decode, CHARSET, convertbits
+from bech32 import bech32_decode, convertbits
 from lib_autopilot import Autopilot, Strategy
-from pyln.client import LightningRpc, Plugin, RpcError
+from pyln.client import Plugin, RpcError
 import random
 import threading
 import math
@@ -92,7 +92,7 @@ class CLightning_autopilot(Autopilot):
             print("Attempt RPC-call to download channels from the lightning network")
             channels = self.__rpc_interface.listchannels()["channels"]
             print("Number of retrieved channels: {}".format(len(channels)))
-        except ValueError as e:
+        except ValueError:
             print("Channel list could not be retrieved from the peers of the lightning network")
             return False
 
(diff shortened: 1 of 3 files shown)


cdecker:
This would be better done in the background in a thread instead of exposing yet another method, that'd cause issues if users forget to call it.

cdecker:
Moved the initialization into a separate thread, with a `threading.Event()` to synchronize when we need it.

autopilot/autopilot.py
```diff
@@ -126,8 +127,17 @@ def init(configuration, options, plugin):
...
+    print('Initialized autopilot function')
 
-    plugin.autopilot = CLightning_autopilot(plugin.rpc)
+    def intialize_autopilot():
+        plugin.autopilot = CLightning_autopilot(plugin.rpc)
+        plugin.initialized.set()
+
+    # Load the autopilot in the background and have it notify
+    # dependents once we're finished.
+    threading.Thread(initialize_autopilot, daemon=True).start()
```
m-schmoock, line 140: We need to name the `target` argument, otherwise it thinks that this is `group` which, if set, must be `None`.
```
    threading.Thread(target=initialize_autopilot, daemon=True).start()
```

autopilot/autopilot.py
```diff
@@ -126,8 +127,17 @@ def init(configuration, options, plugin):
     plugin.num_channels = int(options['autopilot-num-channels'])
     plugin.percent = int(options['autopilot-percent'])
     plugin.min_capacity_sat = int(options['autopilot-min-channel-size-msat']) / 1000
+    plugin.initialized = threading.Event()
+    plugin.autopilot = None
+    print('Initialized autopilot function')
 
-    plugin.autopilot = CLightning_autopilot(plugin.rpc)
+    def intialize_autopilot():
```
m-schmoock, line 134: There is a typo in the method name `intialize_autopilot`, this should be `initialize_autopilot` so the Thread.init call couple of lines below can start it.

m-schmoock (changes requested):
Besides the two issues below, we need to:
- Fix the requirements to work for python 3.9 (networkx 2.3 pulls in code that is not compatible anymore). 
- Remove the try/catch for a very old clighting 0.7.2 in line ~120

I will try to commit a fix ontop on bit0fun/plugins to address those

m-schmoock:
@cdecker I 'fixupped' change e43c925c which resulted in erasing you out of the authors. Sorry. If you want to re-fix it yourself just re-apply my two commits. Otherwise you can resolve my review and merge

m-schmoock:
Okay, now its polished and I added basic tests.

I added try/expect in the init Thread so if something fails the plugin does not crash but print/return proper errors. This is required, because the plugin tries to lookup DNS seeds and such which may fail (and did so in testing environment).

m-schmoock (approved):
Fixed stuff and (self)approved my ontop changes  ;)

-> push da button

m-schmoock:
self ACK edbc1f7

PR #286, opened 2021-08-13
WIP backup: optimize restore and backup-compact

SimonVrouwe:
Since I run the [clboss](https://github.com/ZmnSCPxj/clboss) plugin on CL, my `backup` grew to 7GB in a few weeks and calling `backup-compact` takes ages, blocks CL and shows no progress bar. The `bacup-cli restore` command is similarly slow (just 10 it/s). This is problematic because you can run out of disk space on the (backup) backend and also restore should not take days.

This PR wants to fix that:

- [x] optimize `compact` and `restore` by committing to db only once at the end, should be ok as no concurrent use of the db during compact or restore (now >1000 it/s !).

Other ideas:
- [ ] add `auto-compacting` option: when backup grows above certain size, call `backup-compact` (maybe in separate threat or when CL is not busy?).
- [x] the `backup-cli restore` progress bar is not working because `tqdm` cannot get length of the iterator [here](https://github.com/lightningd/plugins/blob/1f6f701bf1e60882b8fa61cb735e7033c8c29e3c/backup/backend.py#L121) (I think).

### DIFF
--- a/backup/backend.py
+++ b/backup/backend.py
@@ -6,18 +6,18 @@ from typing import Iterator
 import sqlite3
 from tqdm import tqdm
 
-# A change that was proposed by c-lightning that needs saving to the
-# backup. `version` is the database version before the transaction was
-# applied. The optional snapshot reqpresents a complete copy of the database,
-# as it was before applying the `transaction`. This is used by the plugin from
-# time to time to allow the backend to compress the changelog and forms a new
-# basis for the backup.
+# A 'transaction' that was proposed by c-lightning and that needs saving to the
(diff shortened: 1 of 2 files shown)


SimonVrouwe:
BTW, I tested c79d2d4 `restore` and `backup-compact` on a (offline) copy of my 7GB backup and compared db files before/after using `sqldiff`.

cdecker:
ACK https://github.com/lightningd/plugins/pull/286/commits/7ac1594df868c5a635af8b9afaff5c30d5252240

PR #287, opened 2021-08-16
rebalance: add average_forward_fee_ppm for 1s 7d and 30d interval to rebalancereport

m-schmoock:
Using dynamic feeadjuster this can help us better understand our current usage

### DIFF
--- a/rebalance/rebalance.py
+++ b/rebalance/rebalance.py
@@ -684,6 +684,31 @@ def health_score(liquidity):
     return score * coefficient
 
 
+def get_avg_forward_fees(plugin: Plugin, intervals):
+    now = time.time()
+    max_interval = max(intervals)
+    total = [0] * len(intervals)
+    fees = [0] * len(intervals)
+    res = [0] * len(intervals)
+    all_forwards = list(filter(lambda fwd: fwd.get("status") == "settled"
+                               and fwd.get("resolved_time", 0)
+                               + max_interval * 60 * 60 * 24 > now,
+                               plugin.rpc.listforwards()["forwards"]))
+
+    # build intermediate result per interval
+    for fwd in all_forwards:
+        for idx, i in enumerate(intervals):
+            if now > fwd["resolved_time"] + i * 60 * 60 * 24:
+                continue
+            total[idx] += fwd["out_msat"]
+            fees[idx] += fwd["fee_msat"]
+
+    # return average intermediate
+    for idx, i in enumerate(res):
+        res[idx] = fees[idx] / total[idx] * 10**6
+    return res
+
+
 @plugin.method("rebalancereport")
 def rebalancereport(plugin: Plugin):
     """Show information about rebalance
@@ -727,6 +752,12 @@ def rebalancereport(plugin: Plugin):
         res["average_rebalance_fee_ppm"] = round(total_fee / total_amount * 10**6, 2)
     else:
         res["average_rebalance_fee_ppm"] = 0
+
+    avg_forward_fees = get_avg_forward_fees(plugin, [1, 7, 30])
+    res['average_forward_fee_ppm_1d'] = avg_forward_fees[0]
+    res['average_forward_fee_ppm_7d'] = avg_forward_fees[1]
+    res['average_forward_fee_ppm_30d'] = avg_forward_fees[2]
+
     return res
 
 


rebalance/rebalance.py
```diff
@@ -684,6 +684,36 @@ def health_score(liquidity):
...
+def get_avg_forward_fees(plugin: Plugin, intervals):
+    now = time.time()
+    max_interval = max(intervals)
+    total = [0] * len(intervals)
+    fees = [0] * len(intervals)
+    res = [0] * len(intervals)
+    # TODO: next commented out line dows not work. pyln binding bug?
+    #all_forwards = plugin.rpc.listforwards(status="settled")["forwards"]
+    all_forwards = list(filter(lambda fwd: fwd.get("status") == "settled",
+                               plugin.rpc.listforwards()["forwards"]))
```
gallizoltan, line 696: Maybe `fwd.get("status") == "settled" and fwd.get("resolved_time", 0) + max_interval * 60 * 60 * 24 > now`, or something like that?
This way the following code may be cleaner, because of no need for `sort` and `break` in the loop.

gallizoltan (approved):
Looks good to me!

rebalance/rebalance.py
m-schmoock: hm, yes but we actually must fix pyln to accept the "status" when binding. When this is done the line 703 that checks for "resolved_time" later on must keep in place anyway.

m-schmoock: I also checked, when status is resolved, there is always a "resolved_time" in my database

gallizoltan: > I also checked, when status is resolved, there is always a "resolved_time" in my database

Not always. Can you check really old forwards, before v0.7.0?
My database is continuous since 2018 July. 😎

m-schmoock: :D will add this. do you have an idea why `plugin.rpc.listforwards(status="settled")` does not work?

m-schmoock:
@gallizoltan
applied the patch for old version.
do you have an idea why plugin.rpc.listforwards(status="settled") does not work?

gallizoltan:
> @gallizoltan
> applied the patch for old version.
> do you have an idea why plugin.rpc.listforwards(status="settled") does not work?

I don't know, sorry.

rebalance/rebalance.py
```diff
@@ -684,6 +684,37 @@ def health_score(liquidity):
...
+    fees = [0] * len(intervals)
+    res = [0] * len(intervals)
+    # TODO: next commented out line dows not work. pyln binding bug?
+    #all_forwards = plugin.rpc.listforwards(status="settled")["forwards"]
+    all_forwards = list(filter(lambda fwd: fwd.get("status") == "settled"
+                                  and fwd.get("resolved_time", 0) + max_interval * 60 * 60 * 24 > now,
+                               plugin.rpc.listforwards()["forwards"]))
+
+    # sort by age, newest first
+    all_forwards.sort(key=lambda fwd: fwd["resolved_time"], reverse=True)
```
gallizoltan, line 700: No need for this.

rebalance/rebalance.py
```diff
@@ -684,6 +684,37 @@ def health_score(liquidity):
...
+
+    # build intermediate result per interval
+    for fwd in all_forwards:
+        if now > fwd["resolved_time"] + max_interval * 60 * 60 * 24:
+            break
```
gallizoltan, line 705: No need for this, because we filter out older forwards.

gallizoltan (approved):

m-schmoock:
> > @gallizoltan
> > applied the patch for old version.
> > do you have an idea why plugin.rpc.listforwards(status="settled") does not work?
> 
> I don't know, sorry.

It gives me:
TypeError: listforwards() got an unexpected keyword argument 'status'
Does not seem right, as other rpc calls with named arguments work exactly like this

rebalance/rebalance.py
m-schmoock: Right, now thats pointless

rebalance/rebalance.py
m-schmoock: yep, artifact of prior internal version I guess

PR #288, opened 2021-08-16
feeadjuster: disable dynamic base fee scaling

m-schmoock:
This disables scaling of the base fee based on channel distortion and
instead always sets global config base fee.

The rationale behind this is that the base fee is meant to cover the
'fixed risk' of accepting an HTLC. In reality the risk may vary,
but this is unrelated on channel balance distortion but other factors.

We can have dynamic base fee per channel if we implent a better way
of competing and cost/risk coverage for this value.

### DIFF
--- a/feeadjuster/feeadjuster.py
+++ b/feeadjuster/feeadjuster.py
@@ -104,9 +104,8 @@ def get_fees_median(plugin: Plugin, scid: str):
                         and ch['source'] != plugin.our_node_id]
     if len(channels_to_peer) == 0:
         return None
-    fees_base = [ch['base_fee_millisatoshi'] for ch in channels_to_peer]
     fees_ppm = [ch['fee_per_millionth'] for ch in channels_to_peer]
-    return {"base": statistics.median(fees_base), "ppm": statistics.median(fees_ppm)}
+    return {"base": plugin.adj_basefee, "ppm": statistics.median(fees_ppm)}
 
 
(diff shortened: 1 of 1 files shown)


gallizoltan (approved):
The rationale makes sense, and the PR reaches its purpose.
Besides, it leaves some unused code: base fee local variable and function parameter could be deleted from some places.

m-schmoock:
@gallizoltan 
Thanks. I left these by intention, yet we may refactor and remove later if we don't come up with a way of applying dynamic base fees. What are your opinions on base fee selection?

All I can imagine right now is: 
 - Tracking how many failed htlcs (thus unilateral closed channel we encounter)
 - Summing up the Fees for this over the past
 - Tracking how many non-failed HTLCs (and thus their base fees) we had
 - Do the math i.e.:   base_fee = fees_lost_for_failed_htlcs / num_successful_htlcs  
 
 This can be done by tracking only a recent timeframe, i.e. the last 3 month

Do you have an idea to weight certain channels more or less in this calculation? Meaning is there any metric where we can say the risk of loosing an HTLC on a specific channel is bigger or smaller compared to other channels or htlcs ?

gallizoltan:
@m-schmoock Probably there is no point in dealing with the base fee. I think #zeroBaseFee is the future.
https://twitter.com/Snyke/status/1418109408438063104

m-schmoock:
Nah   ;)

PR #290, opened 2021-08-18
explain that plugin can also be added to the config

LightningHelper:


cdecker:
ACK https://github.com/LightningHelper/plugins/commit/e4e48aeb7e3697418b5406d6b7f4eb0fa667ce51

PR #291, opened 2021-08-18
rebalance: division by zero fix

gallizoltan:
this occurs if I don't have any transaction forward for a day


m-schmoock:
ACK d9033fd

Though I didnt know we had the bitcoin-bot/changelog check. We can ignore that I guess

PR #292, opened 2021-08-20
Add lightning-qt in tree from darosior/lightning-qt

darosior:
It deserves more love than i can afford to give it at the moment, hopefully it can be better maintained here as it appears that people are actually using it.

However it's *ancient*: last time i contributed to it was more than 2 years ago, since then i've been only merging bugfixes with months of delay... So i still need to clean it up and test it. I also need to change the way it's installed to get rid of committing big blobs to the repo.


cdecker:
Awesome work @darosior, happy to move this over here and get some more eyes on it.

Just some due diligence questions:
 - What license are the assets under?
 - I assume the code is ok with a permissive BSD-MIT license?
 - Any pending issues that should be replicated here?

darosior:
> What license are the assets under?

I had a [README for the icons](https://github.com/darosior/lightning-qt/blob/master/res/icons/icons_readme.txt) (which i should remove in this PR btw) indicating their origin. Most were taken from the Bitcoin Core project and then modified by myself (to have them in Yellow, much Lightning designing) a while back. The Lightning one was taken from https://paomedia.github.io/small-n-flat/ and [licensed under CC0](https://iconarchive.com/icons/paomedia/small-n-flat/readme.txt). The black ones recently contributed by @grazhopr are available on https://www.freepng.fr . For instance the [thrash icon is](https://www.freepng.fr/png-uy884o/) and they claim to provide these icons free for use in personal projects not intended to be used as commercial products.

> I assume the code is ok with a permissive BSD-MIT license?

Yes. The repository was previously [under a BSD 3-clause license](https://github.com/darosior/lightning-qt/blob/master/LICENCE). Pinging the contributors for full disclosure: @vincenzopalazzo @grazhopr @rsbondi @jb55 @bruuuuuuhhhhhhhh @vadorovsky FYI i'm upstreaming [lightning-qt](https://github.com/darosior/lightning-qt) to which you contributed under a BSD 3-clause licence to the [lightningd/plugins repo](https://github.com/lightningd/plugins) where the code could be given a BSD or MIT license.

> Any pending issues that should be replicated here?

A few that my update should fix, another one that i would probably open here, and a lot of "good first issues" i can think of to open here :)

darosior:
Ok, so i did another round of due diligence for the icons:
- Removed the icons i could not find the provenance of (have been replaced)
- Added the provenance of each and every icon, added a note to the Contributing section

I also did another round of updates and made some basic testing. The code is not super clean, but the software is functional.

grazhopr:
Hi, really sorry taken a long long time to get back on this. For the icons then they are all taken from Bitcoin Core and coloured the same yellow. The only icon that wasnt was the payment_expired which I'd planned to create a new one from combining the clock.png and warning.png (but I didnt :) I just forgot and temporarily found one that looked similar. I'll create i anyway if of use. apologies again for late response and for not being clearer on teh icons...

darosior:
Oh just after i replaced them all :sob: 

> apologies again for late response and for not being clearer on teh icons...

Sure, thanks for the head up!

grazhopr:
> Oh just after i replaced them all 
> 
> > apologies again for late response and for not being clearer on teh icons...
> 
> Sure, thanks for the head up!

Damn, sorry !

darosior:
I had forgotten about this PR.. I'll get back to rebase it. If anyone cares and gets to it before i do please feel free to consider this as up for grab. (And maybe take @icota's fix from https://github.com/darosior/lightning-qt/pull/20.)

chrisguida:
Needs rebase, CI should be working now

chrisguida:
I actually think we should move away from putting things into the repo tree and favor leaving them as submodules. That way the maintainer is clearly defined. What do you think @cdecker ?

cdecker:
Sounds good to me. We can then just have an action that updates submodules to the latest version, and scraping metadata into a repository descriptor that can be read by plugin managers. That would make this repo more of an index, and allow maintainers to move at their own speed.

PR #293, opened 2021-08-29
feeadjuster: feature to adjust a specific channel

m-schmoock:
This adds an optional paramter to the `feeadjust` method which is `scid`
that will, if set, only adjust one channel. Default remains None which
will adjust all channels.


m-schmoock:
@gallizoltan 
simple addition, do we want this?

gallizoltan (approved):
Looks good to me!

PR #294, opened 2021-08-29
datastore: Skip the migration test when we have built-in datastore

cdecker:
The datastore tests were broken if the c-lightning built-in commands were available, which they now are.


rustyrussell:
I think it's just easier to test directly that the datastore plugin tells us it's builtin.  Testing now...

PR #298, opened 2021-09-17
paytest: Fix an error when we unexpectedly get a LegacyPayload

cdecker:
Reported-by: Rene Pickhardt <@renepickhardt>


renepickhardt (approved):

renepickhardt:
didn't test it but it looks good to me. However another fix would be to just record it with a new payment key / secrete. 

Every partial payment would get its own counter but since they are bound through the payment hash and since in experiments I often have full control this might also be a solution.

PR #301, opened 2021-09-24
historian: Instead of exiting just skip large messages

cdecker:
This check was introduced to protect against the length being
corrupted but we now get some node_announcements with huge
feature-bits, triggering this rule. Let's skip them instead.

### DIFF
--- a/historian/historian.py
+++ b/historian/historian.py
@@ -12,6 +12,10 @@ import gossipd
 import struct
 import time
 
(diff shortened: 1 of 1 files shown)


cdecker:
ACK https://github.com/lightningd/plugins/pull/301/commits/3d85f09da93579af900efd596abaea680dc4ee61

PR #302, opened 2021-09-25
Fix Github Actions

cdecker:
c-lightning changed a bit under our feet. This catches up with it.


cdecker:
ACK https://github.com/lightningd/plugins/pull/302/commits/aab969e58cf7950d7a6ffa64f977fff540fbc267

PR #305, opened 2021-10-03
requestinvoice: Make bind address configurable, make default localhost

laanwj:
Add a environment variable `REQUEST_INVOICE_ADDRESS` to configure what address to bind on.

This defaults to `127.0.0.1` (localhost). Especially as the README talks about Tor hidden service usage, I think this is safer than anyaddr binding that was the default before. No one uses unwrapped HTTP over the public internet anymore anyway.


laanwj:
I guess it would be even better to make these command-line options instead of environment variables?

laanwj:
> I guess it would be even better to make these command-line options instead of environment variables?

Added a commit (by @kvaciral) that implements this.

PR #311, opened 2021-11-15
rebalancereport calls listpays once

gallizoltan:
I had several successful rebalances, and when I call `rebalancereport` it takes some time to collect info about them.
A possible performance improvement: `rebalancereport` does not call `listpays` for every rebalance one by one, but only once.
In my case, this changes the execution time from ~140 secs to ~7 secs.

### DIFF
--- a/rebalance/rebalance.py
+++ b/rebalance/rebalance.py
@@ -742,9 +742,13 @@ def rebalancereport(plugin: Plugin):
     total_fee = Millisatoshi(0)
     total_amount = Millisatoshi(0)
     res["total_successful_rebalances"] = len(rebalances)
+    # pyln-client does not support the 'status' argument as yet
+    # pays = plugin.rpc.listpays(status="complete")["pays"]
+    pays = plugin.rpc.listpays()["pays"]
+    pays = [p for p in pays if p.get('status') == 'complete']
     for r in rebalances:
         try:
-            pay = plugin.rpc.listpays(r["bolt11"])["pays"][0]
+            pay = next(p for p in pays if p["payment_hash"] == r["payment_hash"])
             total_amount += pay["amount_msat"]
             total_fee += pay["amount_sent_msat"] - pay["amount_msat"]
         except Exception:
--- a/rebalance/rebalance.py
+++ b/rebalance/rebalance.py
@@ -448,7 +448,7 @@ def feeadjust_would_be_nice(plugin: Plugin):
 
 
 def get_max_amount(i: int, plugin: Plugin):
-    return max(plugin.min_amount, plugin.enough_liquidity / (4**(i + 1)))
+    return max(plugin.min_amount, plugin.enough_liquidity / (4**i))
 
 
 def get_max_fee(plugin: Plugin, msat: Millisatoshi):


m-schmoock (approved):
Hi Zoltan, the code works. Some notes from my side:

1. Maybe you want to add something like the following lines at the end of the testcase `rebalance/test_rebalance.py::test_rebalance_manual` to check that the report works:
```Python

    # briefly check rebalancereport works
    report = l1.rpc.rebalancereport()
    assert report.get('rebalanceall_is_running') == False
    assert report.get('total_successful_rebalances') == 2
```

2. I noticed that when you add the same lines at the end of `rebalance/test_rebalance.py::test_rebalance_all` we see that it does not 2 but 8 rebalances to complete the `rebalanceall` command, which I find odd because `l1` just has 2 channels.

3. A note about the change change: Before it was calling `listpays` for each bolt11, which let lightningd perform the filtering and cause some RPC delay. With this change we request the full `listpays` once which can be quite big. For a reference I did some commandline testing, getting the full list on my Raspi node takes ~0.6 seconds. Getting a specific `listpays` by a bolt11 takes ~0.02s  (that per rebalance). Meaning this change performs better after about ~30 rebalances. I agree that we should go for your change, because most user will reach 30 rebalances and it will only getting slower afterwards.

m-schmoock:
Also CI is broken with python deps :/

gallizoltan:
> 2. I noticed that when you add the same lines at the end of `rebalance/test_rebalance.py::test_rebalance_all` we see that it does not 2 but 8 rebalances to complete the `rebalanceall` command, which I find odd because `l1` just has 2 channels.

That is a good one! I will fix this today.

> 3. A note about the change change: Before it was calling `listpays` for each bolt11, which let lightningd perform the filtering and cause some RPC delay. With this change we request the full `listpays` once which can be quite big. For a reference I did some commandline testing, getting the full list on my Raspi node takes ~0.6 seconds. Getting a specific `listpays` by a bolt11 takes ~0.02s  (that per rebalance). Meaning this change performs better after about ~30 rebalances. I agree that we should go for your change, because most user will reach 30 rebalances and it will only getting slower without your fix afterwards.

I currently have 5713 successful rebalances. 😎

gallizoltan:
> 2. I noticed that when you add the same lines at the end of `rebalance/test_rebalance.py::test_rebalance_all` we see that it does not 2 but 8 rebalances to complete the `rebalanceall` command, which I find odd because `l1` just has 2 channels.

Thanks for noticing!
Just one line of code. I don't open a separate PR for this if I don't have to:
https://github.com/lightningd/plugins/pull/311/commits/dc3a8b08eeda780a536bb17119005ea634c7fbc1

m-schmoock:
Nice. Will do final checks and then merge. If you want to extend the test like I mentioned in 1. Go for it. Otherwise I will add it anyway since we currently don't test the report. :)

gallizoltan:
> Nice. Will do final checks and then merge. If you want to extend the test like I mentioned in 1. Go for it. Otherwise I will add it anyway since we currently don't test the report. :)

When I last tried to write and run tests, it was kind of complicated. I'd rather miss it this time if it is not a big problem.

m-schmoock:
ACK dc3a8b0

PR #324, opened 2021-12-16
Add listmempoolfunds plugin

andrewtoth:
This plugin is motivated by https://github.com/ElementsProject/lightning/issues/2104. Using `listmempoolfunds` instead of `listfunds`, a user can monitor received unconfirmed txs.


cdecker:
Aw shute, didn't see this PR in time, and added your plugin myself. I'd like to keep the new plugins as just a link, so we don't get as many submodule issues, and users know where any issues and improvements should be sent.

PR #329, opened 2022-01-25
skip channels that are not (yet) locked in

svewa:
feeadjuster crashes when trying to read an non-existing short_channel_id. So better make sure the channel does have one before reading it.


cdecker:
ACK https://github.com/lightningd/plugins/pull/329/commits/e0d05b3c010dd9cc4c7dd5f6e4c46d59c0ba42dd

Issue #331, opened 2022-02-06
Noise plugin hangs while trying to receive the most recent message


PestToast:
recvmsg with no id works as it should, but using the most recent received message id it behaves the same way, as if I didn't enter an id at all.
I am able to retrieve older messages, and the output lists the correct number of messages, I am just incapable of reading the most recent one. I have some commands I've been running with outputs, if this helps.

```
user@user-ThinkPad-T420:~/lightning$ l3-cli sendmsg 03c1002e4832efc5528907ac08d4b628c454dfe4507192fbcf219aa5ef979872e2 "test 1"
{
   "route": [
      {
         "id": "03645cd14343cb9a325d1244c396bfb023b6ba49ec23e16bf2e8beb41dfd445273",
         "channel": "378x1x1",
         "direction": 0,
         "msatoshi": 11,
         "amount_msat": "11msat",
         "delay": 15,
         "style": "tlv"
      },
      {
         "id": "03c1002e4832efc5528907ac08d4b628c454dfe4507192fbcf219aa5ef979872e2",
         "channel": "378x2x1",
         "direction": 0,
         "msatoshi": 10,
         "amount_msat": "10msat",
         "delay": 9,
         "style": "tlv"
      }
   ],
   "payment_hash": "b0b158521ee9ed956e5707c2708e39c8a02f8281fdc6888b636581b0f0d854b3",
   "attempt": 1
}
user@user-ThinkPad-T420:~/lightning$ l1-cli recvmsg 0
^C
user@user-ThinkPad-T420:~/lightning$ l3-cli sendmsg 03c1002e4832efc5528907ac08d4b628c454dfe4507192fbcf219aa5ef979872e2 "test 2"
{
   "route": [
      {
         "id": "03645cd14343cb9a325d1244c396bfb023b6ba49ec23e16bf2e8beb41dfd445273",
         "channel": "378x1x1",
         "direction": 0,
         "msatoshi": 11,
         "amount_msat": "11msat",
         "delay": 15,
         "style": "tlv"
      },
      {
         "id": "03c1002e4832efc5528907ac08d4b628c454dfe4507192fbcf219aa5ef979872e2",
         "channel": "378x2x1",
         "direction": 0,
         "msatoshi": 10,
         "amount_msat": "10msat",
         "delay": 9,
         "style": "tlv"
      }
   ],
   "payment_hash": "4876098fa27d5d82b9cd706631e357884f066438ba1921fc6d43bb8b4205d9bd",
[...]

PR #332, opened 2022-02-06
Fixes a small bug with noise.py that prevented user from seeing the most recent message id

PestToast:
Changelog-Fixed: This fixes the problem described in issue #331 . Users are now able to view the latest received message using `recvmsg`


cdecker:
ACK 2ad70d3a3ecccd5a97352a0aa2413b02dda01766

m-schmoock:
@PestToast @cdecker 
I re-opened #331 as this fix introduces other problems.
See my comment:  https://github.com/lightningd/plugins/issues/331#issuecomment-1046217384

PR #340, opened 2022-02-21
fix: autopilot

m-schmoock:
fixes several (not all) testing issues we had with the autopilot plugin.

- fixed a test annotation and CI detection
- fixed a dns.resolver.query deprecation
- added a remark for a possible division by zero (not fixed yet)
- I set the repeated "unable to connect" message to just log-level "warn" so they don't appear as teardown BROKEN errors.


vincenzopalazzo (approved):
ack https://github.com/lightningd/plugins/pull/340/commits/e5226a037e5477fa268b5df64b46203499559ebc

m-schmoock:
Lets do this as is and address the CI issues as a follow up.

PR #344, opened 2022-03-03
commando: change underscore to dash

### LINKED ISSUE #343: commando-reader=<id>: unrecognized option (is commando_reader)
No matter if used as a command-line parameter with `--commando-reader=<id>` or in config file as `commando-reader=<id>`

The option which works is `commando_reader` in the config file and `--commando_reader` in command line of `lightningd`.

So no dash, but underscore. Will do a PR for the readme in a moment.

jsarenik:
Fixes #343

### DIFF
--- a/commando/commando.py
+++ b/commando/commando.py
@@ -362,8 +362,8 @@ def commando_rune(plugin, rune=None, restrictions=[]):
 def init(options, configuration, plugin):
     plugin.out_reqs = {}
     plugin.in_reqs = {}
-    plugin.writers = options['commando_writer']
-    plugin.readers = options['commando_reader']
+    plugin.writers = options['commando-writer']
+    plugin.readers = options['commando-reader']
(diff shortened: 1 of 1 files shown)


jsarenik:
Just a note that I was not trying for a while to use commando since its Python dependency could not be compiled on Alpine Linux. Now it works well. Kudos to @jb55 for triggering my retry (by doing [lnlink](https://github.com/jb55/lnlink)).

laanwj:
TBH I'd fix it the other way around (fix the plugin instead of the docs), it's more consistent with other plugin's arguments.

jsarenik:
> TBH I'd fix it the other way around (fix the plugin instead of the docs), it's more consistent with other plugin's arguments and UNIX argument spelling in general.

Done. Thank you for feed-back.

laanwj:
Thanks, looks good to me now!

PR #346, opened 2022-03-06
Add TypeScript API

AaronDewes:
This adds c-lightning.ts. It is designed to be mostly compatible with clightningjs, but introduces some new features and works better with TypeScript.


vincenzopalazzo (approved):
ACK https://github.com/lightningd/plugins/pull/346/commits/690fcf481a5968010d4fd8c640db1aada824434b

Well typed library are always good :) Thanks!

Issue #347, opened 2022-03-07
commando message hook blocking?


jb55:
I noticed long `pay` commands can hang all commando requests, it looks like the custommsg hook is blocking? Is there a way to make it non blocking?

I found https://github.com/ElementsProject/lightning/pull/2326, not sure if this will help

cc @cdecker

cdecker:
I don't think that the `custommsg` hook is synchronous (otherwise `pay` would also not proceed), it could rather be the consuming plugin being single-threaded, which would mean that any commands that come in while the hook hasn't returned would be delayed. Which client lib are you using?

jb55:
@cdecker wrote:
>Which client lib are you using?

I'm using [lnsocket](http://git.jb55.com/lnsocket). All it does is open a lightning connection and send commando messages. It's pretty easy to block commando by making a request with a large response:

To reproduce:

```sh
$ git clone http://git.jb55.com/lnsocket && cd lnsocket
$ make lnrpc
$ export LN_NODE=nodeid LN_HOST=host LN_RUNE=rune
$ ./lnrpc $LN_NODE $LN_HOST listchannels $LN_RUNE
error: rpc request timeout
$ ./lnrpc $LN_NODE $LN_HOST getinfo $LN_RUNE
error: rpc request timeout

# eventually once the listchannels is done...
$ ./lnrpc $LN_NODE $LN_HOST getinfo $LN_RUNE
{"result": {"id": "03f3c108ccd536b8526841f0a5c58212bb9e6584a1eb493080e7c1cc34f82dad71", "alias": "jb55.com", "color": "ff0000", "num_peers": 24, "num_pending_channels": 1, "num_active_channels": 22, "num_inactive_channels": 0, "address": [{"type": "ipv4", "address": "24.84.152.187", "port": 9735}], "binding": [{"type": "ipv4", "address": "0.0.0.0", "port": 9735}], "version": "v0.10.2-194-g52b1f5a-modded", "blockheight": 726321, "network": "bitcoin", "msatoshi_fees_collected": 13220067, "fees_collected_msat": "13220067msat", "lightning-dir": "/home/jb55/.lightning-bitcoin/bitcoin"}}
```

It looks like commando is probably buffering the large `listchannels` result into memory? making the request eventually timeout. It will block all subsequent RPC calls on other connections like `getinfo` until listchannels is finished. `listchannels` could be a long running command like `pay`, it's the same problem.

jb55:
[...]

PR #348, opened 2022-03-16
commando: make custommsg hook async.

rustyrussell:
Otherwise it blocks everything while it's working, which can be a
while for a slow command.

Fixes: #347 (I think?)
Signed-off-by: Rusty Russell <rusty@rustcorp.com.au>

### DIFF
--- a/commando/commando.py
+++ b/commando/commando.py
@@ -161,8 +161,8 @@ def try_command(plugin, peer_id, idnum, method, params, runestr):
     send_result(plugin, peer_id, idnum, res)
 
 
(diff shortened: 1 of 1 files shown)


jb55:
This doesn't fix #347 but it doesn't hurt anything either...

rustyrussell:
Reproduced.  You're right... hacking...

rustyrussell:
Fixed.

jb55:
thanks! works! although it's weird that processes stay open even after
the connection is closed. we'll save that for another day 😅

PR #350, opened 2022-03-24
include node alias in node metrics

svewa:
I understand that the alias is unreliable and everything, but having it for channels but not for nodes is what I don't get. IF alias is used, it should be as a property of the node, which this introduces.
Also having to work with hundreds of SCIDs is very difficult, so I like the alias.


chrisguida:
Needs rebase

chrisguida:
Same comment as here: https://github.com/lightningd/plugins/pull/351#issuecomment-2146393553

PR #370, opened 2022-06-30
Update Dockerfile

w3irdrobot:
This PR fixes a few things in the Dockerfile.

- allows hardcoding of the lightningd image to avoid breaking the Dockerfile randomly
- updates pip after install to fix issues with Pillow install later
- updates the location of the plugins in the image since the upstream lightningd image moved them
- update python requirements install to allow it to fail when dependency install fails

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,6 @@
-FROM elementsproject/lightningd
+ARG CLN_VERSION="0.11.1"
+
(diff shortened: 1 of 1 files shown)


cdecker:
ACK https://github.com/lightningd/plugins/pull/370/commits/d68a901c19f1fc1ae88e4bc6cdc146ad0bbb9e94

Issue #165, opened 2020-11-30
feeadjuster: a blacklist option


darosior:
Just a reminder for me to implement it, but if a new contributor wants to jump on it i can (and Michael too probably) offer guidance

swissrouting:
@darosior any update on this? I see that openoms already created a PR implementing the feature that's waiting to be reviewed. What will be the next steps?

PR #375, opened 2022-07-17
add option to exclude channels with nodes

openoms:
skip setting channels with the IDs in the `feeadjuster-exclude.list` file

Fixes: #165

The simplest solution I could come up with to achieve not setting all channels when running:
`lightning-cli feeadjust`

To not create a new file a config option would need to be created for each pubkey something like:
```
  "options": [
    {
      "name": "exclude",
      "type": "string",
      "default": 0,
      "description": "A pubkey of a node to not adjust the channels with.",
	  "multi": true
    }
  ```
  then would need to parse the multiple entries into a list.
  
I think it is more useful to have a separate list so the list of pubkeys don't need to be deleted if one stops using the feeadjuster temporarily.

### DIFF
--- a/feeadjuster/feeadjuster.py
+++ b/feeadjuster/feeadjuster.py
@@ -225,13 +225,19 @@ def feeadjust(plugin: Plugin, scid: str = None):
 
     This method is automatically called in plugin init, or can be called manually after a successful payment.
     Otherwise, the plugin keeps the fees up-to-date.
+
+    To stop setting the channels with a list of nodes place a file called `feeadjuster-exclude.list` in the
+    lightningd data directory with a simple line-by-line list of pubkeys.
     """
     plugin.mutex.acquire(blocking=True)
     plugin.peers = plugin.rpc.listpeers()["peers"]
     if plugin.fee_strategy == get_fees_median and not plugin.listchannels_by_dst:
         plugin.channels = plugin.rpc.listchannels()['channels']
     channels_adjusted = 0
+    with open('feeadjuster-exclude.list') as file:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: feeadjuster/feeadjuster.py [hunk 1]
```diff
@@ -225,13 +225,19 @@ def feeadjust(plugin: Plugin, scid: str = None):
...
+
+    To stop setting the channels with a list of nodes place a file called `feeadjuster-exclude.list` in the
+    lightningd data directory with a simple line-by-line list of pubkeys.
     """
     plugin.mutex.acquire(blocking=True)
     plugin.peers = plugin.rpc.listpeers()["peers"]
     if plugin.fee_strategy == get_fees_median and not plugin.listchannels_by_dst:
         plugin.channels = plugin.rpc.listchannels()['channels']
     channels_adjusted = 0
+    with open('feeadjuster-exclude.list') as file:
```
gallizoltan, line 237: 1) The average user does not have a `feeadjuster-exclude.list` file. And without that, this line throws an exception:
`FileNotFoundError: [Errno 2] No such file or directory: 'feeadjuster-exclude.list'`

2) I'm not sure it will be obvious to everyone that the default folder for `feeadjuster-exclude.list` is `~/.lightning/bitcoin/`.

gallizoltan (commented):
Looks good! Thank you very much!
But please make it work without `feeadjuster-exclude.list` file also!

feeadjuster/feeadjuster.py
openoms: good point, will post a change.

PR #376, opened 2022-07-19
summary: fix attribute errors

andrewtoth:
I was getting two `AttributeError`s due to `int`s not being converted to `Millisatoshi`s. Not sure why nobody else was having this issue.
On line 221 - `'int' object has no attribute 'millisatoshis'`
On line 169 - `'int' object has no attribute 'to_btc_str'`

Converting to millisatoshis in two spots fixes the issue.


rustyrussell:
Ack https://github.com/lightningd/plugins/pull/376/commits/9f60f71f30bbc3ed28b164470e8ef026367716b3

PR #379, opened 2022-07-22
Replaces python-teos by rust-teos

sr-gi:
With the last release of `rust-teos`, I have just discontinued `python-teos`, so it's best to replace the watchtower-plugin link.


cdecker:
ACK https://github.com/lightningd/plugins/pull/379/commits/952d3283932518a0cff7fe903fadc0974e870e44

PR #380, opened 2022-07-29
Add referenced requirements.txt for backup plugin

sethforprivacy:
This file is referenced in https://lightning.readthedocs.io/BACKUP.html#backup-plugin-and-remote-nfs-mount, but does not exist, making installation and usage of backup plugin much more tricky.

This commit simply adds the two required packages so that the docs are correct now and the plugin can be more easily installed.

### DIFF
--- a/backup/requirements.txt
+++ b/backup/requirements.txt
@@ -1,0 +1,2 @@
+pyln-client
+psutil


cdecker:
ACK https://github.com/lightningd/plugins/pull/380/commits/c7f25abb029bb4a69a0531883d617f94ae2c8f56

PR #383, opened 2022-08-25
Fixes helpme.py improper calculation of num invoices paid and available out of channel funds

### LINKED ISSUE #382: Helpme crashes when using with node with open channel on regtest
Helpme looks for the 'value' field for each output in plugin.rpc.listfunds()['outputs'], but this field does not exist. Here is an example output from the plugin.rpc.listfunds()['outputs']:

`{'txid': 'e735cbc96a35b38cd0ba90663738f33e049e7200e6ac2083f65f767e7b9c31da', 'output': 0, 'amount_msat': 99899846000, 'scriptpubkey': '00145e6daefe03fcc33d6fad922e2401f0d21e7e2249', 'address': 'bcrt1qtek6alsrlnpn6madjghzgq0s6g08ugjfwa7c9g', 'status': 'confirmed', 'blockheight': 103, 'reserved': False}`

And here is the offending code, on line 457 of helpme.py:
https://github.com/lightningd/plugins/blob/eb03e4fbedc4f6e658611847ccea6d755a9e1cb6/helpme/helpme.py#L457

I guess amount_msat should be being used directly instead, or is the "value" field uniquely not present when using regtest?

### LINKED ISSUE #384: Helpme shows incorrect number of paid invoices
Helpme.py checks for an invoice status being 'complete' when it should be looking for 'paid'. This results in the following message always showing 0:
`STAGE 5 (receiving payments): COMPLETE (0 payments received)`

michaeljohnclancy:
fixes #382 and fixes #384.

### DIFF
--- a/helpme/helpme.py
+++ b/helpme/helpme.py
@@ -499,7 +499,7 @@ node!  Be prepared to lose your funds (but please report a bug if you do!)
     elif len([s for s in invoices if s['status'] == 'paid']) == 0:
         r += "INCOMPLETE (No payments succeeded)"
     else:
-        r += "COMPLETE ({} payments received)".format(len([s for s in invoices if s['status'] == 'complete']))
+        r += "COMPLETE ({} payments received)".format(len([s for s in invoices if s['status'] == 'paid']))
         stages['invoices'] = True
 
     r += "\nSTAGE 6 (adding bling): "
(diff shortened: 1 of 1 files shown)


rustyrussell (approved):
Ack 2367ee92ea47b7bb359827ad3146f8d7e8f7f9a1

Thanks, this clearly needed some love!

PR #386, opened 2022-08-27
feeadjuster: make exclude list non-required and verbose

openoms:
As mentioned in: https://github.com/lightningd/plugins/pull/375#discussion_r956562872 the exclude list needs to be non-required to  run feeadjuster.

Also added the output about the `feeadjuster-exclude.list`file not present or the list of pubkeys.


cdecker:
Would be better to check if the file exists before trying to open it, otherwise any number of exception cause the same message (file permissions, corruptions, etc).

Applying this as is since it fixes a breaking change

PR #387, opened 2022-08-29
rebalance: listpays use status argument

gallizoltan:
It is a performance improvement: `rebalancereport` running time changes from 40 secs to 14 secs on my node.


gallizoltan:
Hi @m-schmoock , what do you think, has the time already come for this?

m-schmoock:
Definitifely yes.
Sorry for the huge delay, this repo really needs some love.

m-schmoock:
I'll merge this in, but will ad a cln version check to do this depending on used cln version with one of my commits in the 'fix alot' branch...

PR #391, opened 2022-09-01
Add `circular` to the list of plugins

giovannizotta:
`circular` is a plugin that helps routing nodes rebalance their channels. It features a custom pathfinding algorithm that gathers liquidity information from failed payments and uses it to be as effective as possible.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -34,3 +34,6 @@
 [submodule "poncho"]
 	path = poncho
 	url = https://github.com/fiatjaf/poncho
+[submodule "circular"]
+	path = circular
+	url = https://github.com/giovannizotta/circular.git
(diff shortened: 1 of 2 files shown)


cdecker:
ACK https://github.com/lightningd/plugins/pull/391/commits/f578f09ca9486b59c7b2406b2d4627fb113921e1****

giovannizotta:
I noticed there were some conflicts because another plugin was added after I did this PR. I resolved them, is there anything else I need to do?

cdecker:
I went ahead and rebased on top of `master`, so it applies cleanly now.

giovannizotta:
@cdecker do I have to do something to merge this? I'm not allowed to author commits, so I guess you should do it. Let me know if I can help

cdecker:
That's quite alright, looks like we need a rebase due to a conflict on `README.md`, but I can do that.

PR #394, opened 2022-09-09
Align README table.

fiatjaf:


fiatjaf:
This should be good now.

PR #396, opened 2022-09-16
fix: a lot of tests and broken stuff

m-schmoock:
This is a general cleanup PR that aims to fix issues that got raised by outdated code and upstream changes.


m-schmoock:
... will still fix the tests for backup and feeadjuster plugin

rustyrussell (commented):
Note that the msatoshi->amount_msat changes in route break everything before 0.12!  You probably want to do a test for this at init, and set a `plugin.msatfieldname`?

Probably best to test for version >= 0.12 (ideally use semver, but just split by '.' works too!).

m-schmoock:
@cdecker we need to synchronize some efforts here with your recent `fix-gci` branch

cdecker:
Yeah, was looking mostly to fix from an infra point of view (test runner script, virtualenvs setup, etc), not so much on fixing individual plugins, though I did try to fix some things on the side

m-schmoock:
> Yeah, was looking mostly to fix from an infra point of view (test runner script, virtualenvs setup, etc), not so much on fixing individual plugins, though I did try to fix some things on the side

Not sure now who wants to rebase, but no problem.

m-schmoock:
@rustyrussell
I used the semver, though its kinda overkill, especially since it does not even filter out/ignore the 'v' in front of the version string.
Some ppl might complain if they don't update their pip environment for a module that only does a string split :D

@gallizoltan

see [4544d51](https://github.com/lightningd/plugins/pull/396/commits/4544d51d5887d3a2033b346508e5fb3e6cfc2c85) on the listpays status version check... 

@ anyone else
If there is no objections on this PR (even though CI is not clean, but thats a topic for another day) I will merge this.

PR #400, opened 2022-11-11
drain: README.md fix parameter order

feelus:
Fix flipped parameter order in documentation so that it matches actual plugin methods https://github.com/lightningd/plugins/blob/master/drain/drain.py#L451


m-schmoock:
@feelus right, thanks...

PR #404, opened 2022-11-20
feeadjuster: Add option to scale median fee adjustment strategy

elsirion:
Currently there is no way to over/underbid the median fees of peers, this PR adds a scaling factor to do so (e.g. set it to `0.5` to calculate fees as if the other peers had 50% lower fees).


m-schmoock:
@elsirion Hey, sorry for the very late reply. This looks neat! I'm runinng tests and reviewing it right now...

elsirion:
I've been running it on my node for quite a while now, seems to work well.

PR #406, opened 2022-12-02
fix: rpc version check for new format

m-schmoock:
This adds a little helper that can be extended with other useful stuff to the plugins that do a version check (rebalance, drain and feeadjuster). Maybe we can move this to a separate package or maybe even into pyln eventually


cdecker:
ACK https://github.com/lightningd/plugins/pull/406/commits/ef8882f881325e738c407d4e368d261a1c6470a2

vincenzopalazzo (approved):
ACK https://github.com/lightningd/plugins/pull/406/commits/ef8882f881325e738c407d4e368d261a1c6470a2

Lets land this so we can get some tests from the community and in the mean while we put this feature in pyln-client, what do you think?

cdecker:
Absolutely agreed, the more we can decouple them, the better it is. Also allows us to play with the API before committing to it by migrating it into the far wider used `pyln-client` package.

PR #410, opened 2022-12-04
fix: autopilot

m-schmoock:
Several upstream incompatibility issues addressed.
What remains is a possible division by zero bug `autopilot/lib_autopilot.py` line `247`.

Since I dont understand the code (I just added the comment a while ago), maybe @cdecker wants to have a look and fix this. When this is fixed, the plugins rudimentary testcase should work again.

### DIFF
--- a/autopilot/autopilot.py
+++ b/autopilot/autopilot.py
@@ -2,7 +2,7 @@
 
 from bech32 import bech32_decode, convertbits
 from lib_autopilot import Autopilot, Strategy
-from pyln.client import Plugin, RpcError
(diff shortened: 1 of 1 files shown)


cdecker:
ACK 8ea5b204d813aada42f310b0cb18c8ada6fb4927

m-schmoock:
@cdecker 

Maybe we should think about adding something like an 'archive' plugin folder with outdated or more or less unused plugins that are no longer included in CI. 

When we move a plugin there, its not gone, the dev can still fix/update/reactivate it.

PR #415, opened 2022-12-13
rebalance: multithreading

gallizoltan:
`rebalanceall` can run on multiple threads for faster results

### DIFF
--- a/rebalance/rebalance.py
+++ b/rebalance/rebalance.py
@@ -3,12 +3,24 @@ from clnutils import cln_parse_rpcversion
 from datetime import timedelta
 from functools import reduce
 from pyln.client import Plugin, Millisatoshi, RpcError
-from threading import Thread, Lock
+import threading
 import time
 import uuid
+import concurrent.futures
 
 plugin = Plugin()
-plugin.rebalance_stop = False
+plugin.rebalance_stop_by_user = False
+plugin.rebalance_stop_by_thread = False
+plugin.threadids = {}
+
+
+def rebalance_stopping():
+    return plugin.rebalance_stop_by_user or plugin.rebalance_stop_by_thread
+
+
+def get_thread_id_str():
+    max_digits = len(str(plugin.threads - 1))
+    return f"{plugin.threadids.get(threading.get_ident(), 0):{'0' + str(max_digits)}}"
 
 
 # The route msat helpers are needed because older versions of cln
@@ -65,6 +77,7 @@ def peer_from_scid(short_channel_id, my_node_id, payload):
             return ch['destination']
     raise RpcError("rebalance", payload, {'message': 'Cannot find peer for channel: ' + short_channel_id})
 
+
 def get_node_alias(node_id):
     node = plugin.rpc.listnodes(node_id)['nodes']
     s = ""
@@ -73,6 +86,7 @@ def get_node_alias(node_id):
     else:
         s += node_id[0:7]
     return s
+
 
 def find_worst_channel(route):
     if len(route) < 4:
@@ -217,7 +231,7 @@ def waitsendpay(payment_hash, start_ts, retry_for):
             return result
         except RpcError as e:
             if e.method == "waitsendpay" and e.error.get('code') == 200:
-                if plugin.rebalance_stop:
+                if rebalance_stopping():
                     raise e
                 if int(time.time()) - start_ts >= retry_for:
                     raise e
@@ -299,7 +313,7 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
     time_sendpay = 0
 
     try:
-        while int(time.time()) - start_ts < retry_for and not plugin.rebalance_stop:
+        while int(time.time()) - start_ts < retry_for and not rebalance_stopping():
             count += 1
             try:
                 time_start = time.time()
@@ -333,7 +347,6 @@ def rebalance(plugin, outgoing_scid, incoming_scid, msatoshi: Millisatoshi = Non
                 excludes.append(worst_channel['channel'] + '/' + str(worst_channel['direction']))
                 continue
 
-
             rpc_result = {
                 "sent": msatoshi + fees,
                 "received": msatoshi,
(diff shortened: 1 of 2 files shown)


gallizoltan:
@m-schmoock I'm working on this. It's a work in progress, not yet finished, please, do not merge it yet. But I think it's a good concept.

rebalance/rebalance.py
```diff
@@ -585,7 +598,7 @@ def maybe_rebalance_pairs(ch1, ch2, failed_channels: list):
             return result
         amount = min(amount, get_max_amount(i, plugin))
         maxfee = get_max_fee(amount)
-        plugin.log(f"Try to rebalance: {scid1} -> {scid2}; amount={amount}; maxfee={maxfee}")
+        plugin.log(f"Thread{get_thread_id_str()} tries to rebalance: {scid1} -> {scid2}; amount={amount.to_satoshi_str()}; maxfee={maxfee.to_satoshi_str()}")
```
m-schmoock, line 601: you also need to fix the testcase `rebalance/test_rebalance.py::test_rebalance_all` for this, as this is still currently waiting just for "Try to rebalance: ..."

rebalance/rebalance.py
```diff
@@ -620,16 +633,68 @@ def maybe_rebalance_pairs(ch1, ch2, failed_channels: list):
...
+            if ch1["lock"].acquire(blocking=False):
+                if ch2["lock"].acquire(blocking=False):
+                    idle_count = 0
+                    pair[2] = True
+                    result = maybe_rebalance_pairs(ch1, ch2, failed_channels)
+                    ch2["lock"].release()
+                ch1["lock"].release()
+            if result["success"]:
+                plugin.log(f"Thread{get_thread_id_str()} restarts rebalance threads after successful rebalance")
+                plugin.rebalance_stop_by_thread = True
```
m-schmoock, line 659: Question: Do I get it right, that the first successful rebalanced pair of any thread will stop the entire pool execution?

rebalance/rebalance.py
```diff
@@ -620,16 +633,68 @@ def maybe_rebalance_pairs(ch1, ch2, failed_channels: list):
...
+            ch1 = pair[0]
+            ch2 = pair[1]
+            processed = pair[2]
+            if processed:
+                continue
+            if ch1["lock"].acquire(blocking=False):
+                if ch2["lock"].acquire(blocking=False):
+                    idle_count = 0
+                    pair[2] = True
+                    result = maybe_rebalance_pairs(ch1, ch2, failed_channels)
```
m-schmoock, line 654: Understanding: Is it correct that the result does not need to get add-up here, as only successful results will be handled by the thread pool execution futures (further below) anyway and the unsuccessful runs do not need to be accounted for?

rebalance/rebalance.py
```diff
@@ -620,16 +633,68 @@ def maybe_rebalance_pairs(ch1, ch2, failed_channels: list):
     return result
 
 
+def rebalance_pair_picker(threadid, channel_pairs: list, failed_channels: list):
+    plugin.threadids[threading.get_ident()] = threadid
+    result = {"success": False, "fee_spent": Millisatoshi(0)}
+    idle_count = 0
```
m-schmoock, line 639: Can you explain (to me) the usage of this `idle_count` variable?
It looks to me like it is increased whenever an unfitting or already processed pair is found. Its then set to `0` when it will actually do something. It will log a message when it was `1`, and wait 10 secs if its greater than `0` in the end. It is never decreased (Other than explicitly setting it to `0` once) . Is this correct?

Maybe my understanding of Threads in Python is not reflecting reality :D, but isn't `idle_count` (and all the other variables in this function, like also `result`) on a per-thread basis so that any thread has its own distinct version of this variable?

m-schmoock (dismissed):
Hi Zoltan, it looks really good!
Just some questions and a broken test.

rebalance/rebalance.py
gallizoltan: The last time I tried (it was more than a year ago) it was so difficult to make the tests work. But OK, I will try.

rebalance/rebalance.py
gallizoltan: Yep, we restart the pool. We need to get the channels again, as the balances changed, and we should continue with the new balances.

It's important to note this is not a complete restart, because `failed_channels` remains the same. So if we cannot rebalance a pair, we will not try again.

rebalance/rebalance.py
gallizoltan: Yes, we cannot do anything with the unsuccessful runs. It's the same as before.

rebalance/rebalance.py
gallizoltan: `idle_count`: it is on a per-thread basis, so every thread has its own `idle_count`.
If the thread can try at least one rebalance, the `idle_count` become zero. If the thread checked all possible pairs, and cannot do anything, then `idle_count` increases. Means that this thread is idle. So the thread goes to sleep (as nothing to do), and logs that it is idle (but only once, when the `idle_count` equals 1, to not to spam the log).

The idea on the background: it is difficult to get the ideal number of threads. My node has 40 active channels, and in the beginning of a `rebalanceall` up to 12 threads can work parallel. After a time, some threads cannot find a job. Nearing the end, it's only 1 or 2 threads are working, because of the few possibilities. But this is not a continuous decrease in the number of working thread, because sometimes as a rebalance attempt finishes, two others can start, due to the channels being released. So it is not a good concept to terminate the idling thread, as they can be useful later.

rebalance/rebalance.py
m-schmoock: Okay, how is it restarted then? I can only see that it sets `rebalance_stop_by_thread` to `True`, which will stop it.
Which code does the restarting?

m-schmoock:
ACK 58ed3cf

rebalance/rebalance.py
gallizoltan: After all the threads stopped, the execution returns into `maybe_rebalance_once`, which last line is: `plugin.rebalance_stop_by_thread = False`
Then the `rebalanceall_thread` will call `maybe_rebalance_once` again, and `maybe_rebalance_once` gets fresh channel states in the first line.

It's almost the same as before this modification. `maybe_rebalance_once` was returning with a successful result and starting again with fresh channel states – before this commit too.

m-schmoock: Ahhhh, thanks :D

PR #416, opened 2022-12-27
cleanup: backup

m-schmoock:
Just code nits and minor changes, but since its an important plugin, we make a proper PR :)

### DIFF
--- a/backup/backend.py
+++ b/backup/backend.py
@@ -93,7 +93,7 @@ class Backend(object):
         return stmt
 
     def _restore_transaction(self, tx: Iterator[str]):
-        assert(self.db)
+        assert self.db
         cur = self.db.cursor()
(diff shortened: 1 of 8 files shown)


m-schmoock:
I have 2 more nits, but I don't know how to handle them.
```
./backup/backends.py:12:18: F821 undefined name 'Mapping'
./backup/backends.py:30:9: F821 undefined name 'kill'
```

- I don't know what `Mapping` is supposed to be
- `kill('foo')` is defined statically in `backup.py`, but if I import it properly, we have a circular dependency

cdecker:
ACK https://github.com/lightningd/plugins/pull/416/commits/57b3c60bdc998d8ddf4a2975e567d4b8a52a2705

PR #417, opened 2022-12-27
github: fix CI workflows

m-schmoock:
We have a couple of workflow config issues running CI since a long time.
Lets try to address them.

PR #418, opened 2022-12-29
Fixups on some of @cdecker's plugins

cdecker:
Builds on top of #417


request-invoice/requestinvoice.py
```diff
@@ -1,125 +0,0 @@
-#!/usr/bin/env python3
```
m-schmoock, line 1: Why not moving this into the `/archive` like I did for the old `commando` plugin ( 748d50a ) instead?

backup/test_backup.py
```diff
@@ -49,13 +49,17 @@ def test_start_no_init(node_factory, directory):
         options=opts, cleandir=False, may_fail=True, start=False
     )
 
-    with pytest.raises(ValueError):
```
m-schmoock, line 52: See https://github.com/lightningd/plugins/pull/416/commits/5892bdc13d27e4dceed41795bdae22845232be90 on how I solved that. 

The test still works correctly, we just need to wait for the 30secs timeout. Maybe we can change the test by setting the `wait_for` `TIMEOUT` value specically lower for this test.

m-schmoock: This commit is part of the PR https://github.com/lightningd/plugins/pull/416

request-invoice/requestinvoice.py
cdecker: Because it is rather annoyingly useless and pretty much requires a full rewrite anyway.

backup/test_backup.py
cdecker: Merged that PR, now we need to rebase here :-)

cdecker:
Rebased on top of `master` after applying #416

m-schmoock:
Wow, actually green tests 🎆

cdecker:
Nice, looks like things are passing now. @m-schmoock do you want to rebase #417 on top of `master` so and then I can merge the two back to back?

cdecker:
By the way, further work I intend to do includes:

 - [ ] Downloading nightly and stable releases of CLN instead of building them
 - [ ] Test against stable and nightly daily, to have a canary for breaking plugins when we deprecate things
 - [ ] Split testing into separate jobs, to further isolate the runs, hopyfully we find a way to automatically file issues as well when a plugin fails, calling out the maintainers.

m-schmoock:
> Nice, looks like things are passing now. @m-schmoock do you want to rebase #417 on top of `master` so and then I can merge the two back to back?

Done

backup/requirements.txt
```diff
@@ -1,2 +0,0 @@
-pyln-client>=0.12
```
m-schmoock, line 1: Agree... But whats the trick to tell CLN to use poetry on a plugin?

The reason I'm still using pip requirements is that when I install them `--user` (and they don't clash with deps each other) CLN is able to start it without loading a special environment.

Maybe I'm missing information, if so we also need to update `README.md` on how to run plugins correctly with poetry. Currently when I have a poetry plugin I do `poetry export --without-hashes -f requirements.txt --output requirements.txt ` and install that again via `pip install --user -r requirement.txt`.  :D

noise/noise.py
```diff
@@ -53,13 +53,27 @@ def to_dict(self):
...
+        if i < 2**8:
+            return struct.pack("!b", i)
+        elif i < 2**16:
+            return struct.pack('!H', i)
+        elif i < 2**32:
+            return struct.pack("!I", i)
+        else:
+            return struct.pack("!Q", i)
+
+    print("XXX", repr(scid))
```
m-schmoock, line 67: I see there is still debug output.
I repeat the review when your ready ;)

autopilot/lib_autopilot.py
```diff
@@ -246,7 +246,7 @@ def __get_long_path_pdf(self, skew=True, smooth=False):
 
         s = sum(path_pdf.values())
         # FIXME: next line can raise a division by zero
```
m-schmoock, line 248: This comment can be deleted I guess

m-schmoock (changes requested):

backup/requirements.txt
cdecker: Yeah, it's a bit tricky: if instead of a full she-bang pointing to a specific `python3` we use `/use/bin/env python3` then the plugin executable will use the `$PATH` envvar. Both `venv` and `poetry run` will prefix the `$PATH` with their `python3` location (allowing to enter and exit an env). So either `python3 -m pip install` or `python3 -m poetry install` will install things in an env. With these changes we get equivalence between the two, by disallowing poetry's virtualenv logic, and allocating a throwaway env for the test.

Long story short: use `python3 .ci/test.py [pluginname]` yo tun tests as they will be run on CI

cdecker:
ACK https://github.com/lightningd/plugins/pull/418/commits/6887ecd347983206a75fcdec9085f37966a8fd6c

backup/requirements.txt
m-schmoock: Thx. How does cln itself start a python plugin? Does it do a similar trick to check if it needs to load a special env or is it just running `python3 plugin.py` ?

PR #429, opened 2023-02-23
summary: request ascii mode

m-schmoock:
This adds an optional parameter `ascii` to hte `summary` method as well as an plugin option flag `summary-ascii`.
This gives the user the option to use ascii output if requested.
Sometimes useful for fixed fonts and such.


vincenzopalazzo (approved):
ACK https://github.com/lightningd/plugins/pull/429/commits/3d44b823ebcde073e087429deca34fd939ec0a82

PR #453, opened 2023-05-30
nostr-control

joelklabo:
Adding nostr-control and removing Nostrify (now unmaintained, re-implemented as nostr-control)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -40,9 +40,6 @@
 [submodule "python-teos"]
 	path = python-teos
 	url = https://github.com/talaia-labs/python-teos
-[submodule "nostrify"]
-	path = nostrify
-	url = https://github.com/joelklabo/nostrify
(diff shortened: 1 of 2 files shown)


chrisguida:
Needs rebase

chrisguida:
@joelklabo do you still want me to merge this? There are conflicts that need to be resolved.

If nostr-control is not being actively maintained, can you remove it from this PR and keep only the part that removes nostrify?

If you are unavailable to do this, please let me know so I can do it instead.

chrisguida:
Whoops, I forgot [we already removed nostrify and @joelklabo stated that he isn't maintaining the newer nostr-control](https://github.com/lightningd/plugins/pull/486) either.

Thinking it might be nice to keep everything in an archived list that isn't in the README, even plugins that never made it in here, just so users can search through old code that worked for someone at some point. 

Closing.

PR #454, opened 2023-06-01
Summars

daywalker90:
Another take on the summary plugin with these key differences (not necessarily pros or cons, that's in the eye of the user):

- optionally show recent forwards, paid invoices and/or payments done
- always one copy&paste for a command away to get more details on what you see (e.g. copy peer id and do listpeerchannels, payment_hash for pay's and invoice label for listinvoices)
- show channel balance as numbers instead of bars
- no fiat

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -46,3 +46,6 @@
 [submodule "sling"]
 	path = sling
 	url = https://github.com/daywalker90/sling
+[submodule "summars"]
+	path = summars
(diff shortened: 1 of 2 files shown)


daywalker90:
updatet to v2.0 for cln 23.11+ using paging for speedup and more filtering options.

daywalker90:
rebased, squashed and fixed tests for master

daywalker90:
updated summars to v3.1.0 for cln 24.02

Issue #450, opened 2023-05-19
prometheus: update for CLN v23.05 with msat purge


bleetube:
Update prometheus.py for for CLN v23.05 with msat purge

```
UNUSUAL plugin-prometheus.py: ----------------------------------------
UNUSUAL plugin-prometheus.py: Exception occurred during processing of request from ('127.0.0.1', 45976)
UNUSUAL plugin-prometheus.py: Traceback (most recent call last):
UNUSUAL plugin-prometheus.py:   File \"/usr/lib/python3.11/socketserver.py\", line 691, in process_request_thread
UNUSUAL plugin-prometheus.py:     self.finish_request(request, client_address)
UNUSUAL plugin-prometheus.py:   File \"/usr/lib/python3.11/socketserver.py\", line 361, in finish_request
UNUSUAL plugin-prometheus.py:     self.RequestHandlerClass(request, client_address, self)
UNUSUAL plugin-prometheus.py:   File \"/usr/lib/python3.11/socketserver.py\", line 755, in __init__   
UNUSUAL plugin-prometheus.py:     self.handle()
UNUSUAL plugin-prometheus.py:   File \"/usr/lib/python3.11/http/server.py\", line 432, in handle
UNUSUAL plugin-prometheus.py:     self.handle_one_request()
UNUSUAL plugin-prometheus.py:   File \"/usr/lib/python3.11/http/server.py\", line 420, in handle_one_request
UNUSUAL plugin-prometheus.py:     method()
UNUSUAL plugin-prometheus.py:   File \"/home/lightning/.local/lib/python3.11/site-packages/prometheus_client/exposition.py\", line 151, in do_GET
UNUSUAL plugin-prometheus.py:     output = encoder(registry)
UNUSUAL plugin-prometheus.py:              ^^^^^^^^^^^^^^^^^
UNUSUAL plugin-prometheus.py:   File \"/home/lightning/.local/lib/python3.11/site-packages/prometheus_client/openmetrics/exposition.py\", line 14, in generate_latest
UNUSUAL plugin-prometheus.py:     for metric in registry.collect():
UNUSUAL plugin-prometheus.py:   File \"/home/lightning/.local/lib/python3.11/site-packages/prometheus_client/registry.py\", line 75, in collect
UNUSUAL plugin-prometheus.py:     for metric in collector.collect():
[...]

PR #451, opened 2023-05-22
prom: Fix breakage due to msat purge

cdecker:
Closes #450

PR #459, opened 2023-06-23
prom: Handle channel attribute name changes and fix msat conversion

bleetube:
This is my own swing at https://github.com/lightningd/plugins/pull/451

I noticed that `to_satoshi()` was a method defined in pyln-client for the Millisatoshi type, so I made sure to cast values as that type so the conversion works as intended.

This closes https://github.com/lightningd/plugins/issues/450 which was due to changes in various attribute names for the v23.05 release of CLN.

### DIFF
--- a/prometheus/prometheus.py
+++ b/prometheus/prometheus.py
@@ -32,12 +32,12 @@ class NodeCollector(BaseLnCollector):
             value=blockheight,
         )
 
-        fees_msat = info["msatoshi_fees_collected"]
-        yield GaugeMetricFamily(
-            'lightning_fees_collected_msat',
-            'How much have we been paid to route payments?',
-            value=fees_msat,
-        )
+#       fees_msat = info["msatoshi_fees_collected"]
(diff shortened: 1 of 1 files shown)


cdecker:
Thanks for the PR, I fixed up #451 and we should be good for now, but in case you want to test things locally you can try out the following command:

```
python3 .ci/test.py prometheus
```

This will limit the CI runner to the `prometheus` directory. For each plugin in the CI run the runner will:

 1. Determine whether it is a Python/requirements.txt or a Python/Poetry plugin (skipping if it is neither for the moment, but support for non-python plugins is definitely planned).
 2. Create a `virtualenv` to install the dependencies in and to run the tests in
 3. Install additional packages we determine to be useful (`pytest`, `pytest-timeout`, ...)
 4. Run the `pytest` tests in the virtualenv

We use virtualenvs extensively in CI in the hopes of creating a reproducible environment that devs can also run locally.

PR #484, opened 2024-01-30
Fix CI

fmhoeger:
This PR makes the CI tests pass again. 

* Support Python versions 3.8, 3.9, 3.10, 3.11, and 3.12
* Move plugins whose tests do not pass into new `archived` subdirectory. These are: `autopilot`,  `backup`,  `commando`, `donations`,  `drain`, `feeadjuster`, `helpme`,  `historian`, `jitrebalance`, `noise`, `paytest`,  `probe`,  `prometheus`,  `rebalance`,  `summary`.
* Update Dockerfile `CLN_VERSION` to 23.11.2
* Add `bitcoind-version` 26.0 to matrix 
* Add `cln-version` v23.11 and master branch to matrix
* Remove all references to `DEVELOPER` env var
* Use ubuntu-22.04 image for CI completion job
* Update packages in pyproject.toml and poetry.lock
* Update Python versions in README


chrisguida (changes requested):
Thanks for working on this!

I think I want to move the unmaintained plugins into a section called "Unmaintained". "Archive" doesn't really seem like the right place for plugins that we want to remove from testing but may become maintained again, once the maintainers return.

From @cdecker :

>No, I think we should a) ping authors when things break, b) exclude from testing if it doesn't get fixed in a timely manner (it is distracting for other plugin devs), followed by c) archival (move it out of the main directly) and finally d) deletion if there is no volunteer to maintain the plugin.

>I think of us as repo maintainers, not plugin maintainers. Our job is to guide others through the process of developing and publishing their plugin, ensuring quality standards and safety concerns are addressed, but we aren't the ones responsible for addressing them. Does that make sense?

So, I think the best strategy is to move the plugins into "Unmaintained" for now, and create a new section on the README that identifies these plugins.

At the same time we can ping the plugin maintainers to update their plugins, and hopefully they will return.

.github/workflows/main.yml
```diff
@@ -12,41 +12,40 @@ on:
 
 jobs:
   build-and-test:
-    name: Test PY=${{ matrix.python-version }}, DEV=${{ matrix.developer }}, EXP=${{ matrix.experimental }}, DEP=${{ matrix.deprecated }}
-    runs-on: ubuntu-latest
+    name: Test PY=${{ matrix.python-version }}, BCD=${{ matrix.bitcoind-version }}, EXP=${{ matrix.experimental }}, DEP=${{ matrix.deprecated }}
+    runs-on: ubuntu-22.04
```
chrisguida, line 16: Why?

.github/workflows/main.yml
```diff
@@ -12,41 +12,40 @@ on:
...
     timeout-minutes: 60
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.7, 3.8, 3.9]
-        developer: [0,1]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        bitcoind-version: ["22.0", "26.0"]
```
chrisguida, line 22: Testing two different versions of bitcoind is probably overkill, but kinda neat. Thoughts @cdecker ?

.github/workflows/main.yml
```diff
@@ -12,41 +12,40 @@ on:
...
 
-    - name: Checkout c-lightning@master
-      uses: actions/checkout@v3
+    - name: Checkout c-lightning@v23.11
+      uses: actions/checkout@v4
       with:
         repository: 'ElementsProject/lightning'
         path: 'lightning'
-        ref: 'master'
+        ref: 'v23.11'
```
chrisguida, line 34: I'm not sure if this is a good idea, testing against master will make the CI fail faster when stuff breaks, so things will break more often, but we'll know sooner.

Thoughts @cdecker ?

.github/workflows/main.yml
```diff
@@ -12,41 +12,40 @@ on:
...
         submodules: 'recursive'
         fetch-depth: 0  # Required for pyln versions to be recognized
 
     - name: Download runtime dependencies
       run: |
-        export BITCOIND_VERSION="0.20.1"
-        wget https://storage.googleapis.com/c-lightning-tests/bitcoin-${BITCOIND_VERSION}-x86_64-linux-gnu.tar.bz2
-        tar -xjf bitcoin-${BITCOIND_VERSION}-x86_64-linux-gnu.tar.bz2
+        export BITCOIND_VERSION=${{ matrix.bitcoind-version }}
+        wget https://bitcoincore.org/bin/bitcoin-core-${BITCOIND_VERSION}/bitcoin-${BITCOIND_VERSION}-x86_64-linux-gnu.tar.gz
```
chrisguida, line 41: Not sure why we used to use a Google storage bucket here and avoided using bitcoincore.org, but this is definitely the simplest fix for this.

chrisguida:
@fmhoeger another thing, can you revert the changes I made to the demoted plugins? It will be disorienting for maintainers when they come back to fix their plugins and the code has changed.

.github/workflows/main.yml
fmhoeger: I thought it would make sense to pin the image to a certain version to prevent surprises when `ubuntu-latest` gets updated to a newer version. But I changed my mind. It actually _does_ make sense if the tests stop passing for some reason with the latest Ubuntu :+1:
Fixed

daywalker90:
I added some commits in a PR here https://github.com/fmhoeger/plugins/pull/1 is that a good way to contribute?

chrisguida:
@daywalker90 yes, this is very helpful, thanks!! Let's add that in as a separate PR, because the focus of this PR is just to get the CI up and running again. New features should go in separately.

fmhoeger:
> > @fmhoeger this is great! Unfortunately it looks like someone has updated `rebalance.py` on master in the meantime... Can you go ahead and rebase this please? Maybe `rebalance` no longer needs to be considered unmaintained?
> 
> @chrisguida OK, rebased just now. Will check if the tests pass for `rebalance`.

@chrisguida Tests for `rebalance` still don't pass after the rebase. So we need to keep it in 'Unmaintained' for the time being.

chrisguida:
We could add back this note as well: https://github.com/lightningd/plugins/commit/cf96eb63c1687644042ea85acfd3785527fd98d4#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L184-L186

chrisguida:
@cdecker this is ready to go.

For the merge, should we do:
1) Merge commit as is
2) Merge commit after squashing some things
3) Squash into one commit

?

chrisguida (approved):

m-schmoock:
> Oh hmm, I guess mergify merged it after my approval.... whoops

:D

m-schmoock:
@chrisguida The workflow badge still shows failing, even though the last master run was green. Maybe the badge should be changed, not shure though: https://github.com/lightningd/plugins/pull/489

gallizoltan:
What good should I do to get back the `rebalance` plugin?

PR #485, opened 2024-02-03
Checkout submodules for CI

daywalker90:
based on https://github.com/lightningd/plugins/pull/484

Plugins in submodules will get checked out, so they can get tested.


daywalker90:
CI will fail because nostrify is still in repo and is now getting tested by this pr. See https://github.com/lightningd/plugins/pull/486

chrisguida:
ACK 4536d8d469cadb48b3029e595314ba241aff72ef

PR #486, opened 2024-02-03
remove unmaintained nostrify

daywalker90:
According to the plugins README this plugin is unmaintained/obsolete. It's also failing it's tests.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -40,6 +40,3 @@
 [submodule "python-teos"]
 	path = python-teos
 	url = https://github.com/talaia-labs/python-teos
-[submodule "nostrify"]
-	path = nostrify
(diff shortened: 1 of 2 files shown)


chrisguida:
Looks like `nostr-control` is the replacement? https://github.com/joelklabo/nostr-control

@joelklabo do you have any comment?

joelklabo:
Yeah, that's right. To be honest I haven't tested either in a while though.

cdecker:
Got it :+1:

cdecker:
I'm happy with removing plugins if the former maintainer abandons them. The plugin code is still there, just not linked here. If it were a plugin whose code is in this repo I'd be tempted to move them to archived, just so the code remains readable, and potentil maintainers can come and re-activate the plugin's life again.

PR #487, opened 2024-02-04
Rust plugin CI support

daywalker90:
based on #484 
This is a first draft with the goal to add support for rust plugins to be tested by the CI. 

What i still like to add is a way to let these rust plugins run some "init" to setup the compiled binary or some other test dependencies, e.g. generate grpc bindings. Maybe look for a bash script "ci.sh" and run that?

### DIFF
--- a/.ci/test.py
+++ b/.ci/test.py
@@ -56,6 +56,10 @@ def enumerate_plugins(basedir: Path) -> Generator[Plugin, None, None]:
     ]
     print(poetry_pytest)
 
+    rust_pytest = [
+        x for x in plugins if (x / Path("Cargo.toml")).exists()
+    ]
+
     for p in sorted(pip_pytest):
         yield Plugin(
             name=p.name,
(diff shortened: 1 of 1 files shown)


chrisguida:
Very cool! let's merge this once #484 is in :)

daywalker90:
Are you sure? What about the problem of setting up the test environment (compiling/ python test dependencies etc.)?

chrisguida:
I guess what I meant to say is, let's hold off on merging this until #484 is in.

Can you explain to me more what the problem is? Does this work for `holdinvoice`?

daywalker90:
Half of the tests work, other half (grpc tests) do not because i'm missing the grpc code and before i figure out a way to get that in the CI i wanted to see on what solution we settle on regarding CI-test-setup.

Right now i download the plugin binary from github [see here](https://github.com/daywalker90/holdinvoice/blob/fe4e7f64de4927813cf1a933599026721ace0b80/tests/util.py#L19). I don't know if you even want other plugins to compile during CI (would make it run really long if it's all rust plugins). 

Also i'm sure we should atleast look for a "requirements-dev.txt" or "requirements-tests.txt" or something so we can install python dependencies like we do for python plugins. (can't be "requirements.txt" because then the plugin gets detected as "python")

chrisguida:
Ok thanks for the details!

Yeah I think maybe just adding a link to a binary may be the way to go for now until we have a more general solution.

Plugin devs will assert that the binary they're using is indeed the binary that is built at the commit the submodule link points to.

chrisguida:
And yes, your suggestion of adding requirements-tests.txt or requirements-dev.txt is good.

I'd be fine with either since I don't know which would be more standard.

Perhaps @fmhoeger knows?

chrisguida:
Is this replaced by #491?

daywalker90:
> Is this replaced by #491?

Pretty much for now, yes

PR #491, opened 2024-02-07
Generic plugin support

daywalker90:
I've thought more about https://github.com/lightningd/plugins/pull/487 and since CLN supports basically any language for plugins  i propose to take this generic approach instead.
Non-python plugins put their `requirements.txt` and `setup.sh` in the `tests` folder (the one we already search for executing tests). The bash script will allow python developer to setup the plugin environment as they see fit.

Example [setup.sh](https://github.com/daywalker90/summars/blob/bash-test-setup/tests/setup.sh) from my summars plugin and then a simple switch in my [utils.py](https://github.com/daywalker90/summars/blob/5068ba0a5420be2c9f1147cb824819ea7a89e545/tests/util.py#L15) for local dev and CI

### DIFF
--- a/.ci/test.py
+++ b/.ci/test.py
@@ -200,7 +200,7 @@ def prepare_generic(p: Plugin, directory: Path):
     if p.details['setup'].exists():
         print(f"Running setup script from {p.details['setup']}")
         subprocess.check_call(
-            ['bash',  p.details['setup']],
+            ['bash',  p.details['setup'], f'TEST_DIR={directory}'],
             stderr=subprocess.STDOUT,
         )
     install_pyln_testing(pip_path)
--- a/.ci/test.py
+++ b/.ci/test.py
@@ -56,6 +56,8 @@ def enumerate_plugins(basedir: Path) -> Generator[Plugin, None, None]:
     ]
     print(poetry_pytest)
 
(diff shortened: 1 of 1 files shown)


chrisguida:
Neat! So we just require plugins to provide a release binary via a `setup.sh` script, and then we use that for the pytests?

chrisguida:
If I run the CI now, will it test anything?

daywalker90:
No because there is no non-python plugin that has passing tests in the repo merged atm. But i tested it on my fork (where i merged #485 and #486 and a special branch of #454 ) here:
https://github.com/daywalker90/plugins/actions/runs/7820798111/job/21336309829
ignore the other tests they had an error compiling cln or something

daywalker90:
I was in need of getting the python bin that is using the specific test env into the setup.sh. I think there could be some other use cases for this, so i added the env variable TEST_DIR.

daywalker90:
rebased

chrisguida:
ACK fb1637d5ef055f3cf75224d77be7cefb0d0f56f8

chrisguida:
@mergifyio queue

chrisguida:
@mergifyio refresh

chrisguida:
@mergifyio queue

PR #499, opened 2024-03-02
Pin poetry to 1.7.1 to fix tests for poetry-based plugins

sip-21:
This PR is a temporary fix for the poetry-based backup and donations plugins whose tests are currently failing in CI. Recently released poetry v1.8 seems to cause problems in the current test setup. Pinning poetry to v1.7.1 makes the tests pass again.

### DIFF
--- a/.ci/test.py
+++ b/.ci/test.py
@@ -120,7 +120,7 @@ def prepare_env_poetry(p: Plugin, directory: Path) -> bool:
     subprocess.check_call(['which', 'python3'])
 
     subprocess.check_call([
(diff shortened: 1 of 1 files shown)


chrisguida:
@cdecker are you okay with this change? We may need to update poetry to a newer version eventually, but it seems like 1.7.1 is ok for now?

chrisguida:
@sip-21 any idea how to fix the broken plugins if we unpin later?

sip-21:
> @sip-21 any idea how to fix the broken plugins if we unpin later?

At this point, no. It might well be a problem with the new Poetry version, not sure.
It looks like the way Poetry 1.8 installs the dependencies in the context of the test setup has changed.

PR #512, opened 2024-04-21
holdinvoice v1.0.2

daywalker90:
# Changelog

### Fixed

- Race condition when restarting a node while holding alot of invoices. Holdstate of invoices previously in ACCEPTED state can now temporarily be in OPEN state during a node restart. We would previously CANCEL invoices if the amount of sats in ACCEPTED htlcs was dropping below the amount of the invoice.


chrisguida:
ACK e56487e7db2ca83bfdc07aeb76a3ec952046731c

chrisguida:
@mergifyio queue

chrisguida (dismissed):
Approving for auto-merge

chrisguida (approved):

PR #517, opened 2024-04-22
archive plugin sparko

daywalker90:
@fiatjaf  This plugin is archived and has no tests. It should therefore be removed.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,9 +10,6 @@
 [submodule "lightning-qt"]
 	path = lightning-qt
 	url = https://github.com/darosior/pylightning-qt.git
-[submodule "sparko"]
-	path = sparko
-	url = https://github.com/fiatjaf/sparko.git
 [submodule "trustedcoin"]
 	path = trustedcoin
 	url = https://github.com/fiatjaf/trustedcoin.git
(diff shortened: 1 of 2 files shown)


fiatjaf:
I agree.

chrisguida:
ACK 58de1ed0595a24a53510dc89ede96a00acca648e

Are there any other plugins that do something similar?

Maybe we can keep around a list of old plugins that are no longer maintained but do something useful and might serve as examples for new devs?

vincenzopalazzo (approved):
ACK https://github.com/lightningd/plugins/pull/517/commits/58de1ed0595a24a53510dc89ede96a00acca648e

chrisguida:
I think keeping old plugins around in the "archived" section is useful for new devs looking for old plugins that do something interesting and which no new plugin currently does. Otherwise they just disappear forever.

Closing unless a compelling argument can be made as to why sparko is causing harm by remaining in the archived section.

chrisguida (changes requested):
I think we should move this to the archived section rather than removing it entirely.

chrisguida (approved):

chrisguida (approved):
ACK ebc44fd

PR #518, opened 2024-04-22
add missing submodule cln-ntfy

daywalker90:
It's in the README but not in .gitmodules. Don't know why but it should either be added with this PR or removed from README.


chrisguida:
LGTM

ACK d051fa01773da17f66214e5a61d2d335f0edc164

Issue #457, opened 2023-06-04
feeadjuster setting base fees of 1000 msats even though fee-base is set to 0


dazl:
After updating CLN to v23.05

with no other config changes

fee adjuster now sets base fees of 1000 msats for all channels 

before updating it would adjust ppm fees but leave base fees unchanged at zero.

```
$ lightning-cli listconfigs |grep "\"fee-base\":"
   "fee-base": 0,
```

wtogami:
feeadjuster needs to be itself adjusted for the deprecations in v23.05?

chrisguida:
Fixed by #525

PR #467, opened 2023-07-17
feeadjuster: basefee switch

m-schmoock:
- make dynamic adjustments of the base fee optional and off per default.
Currently only affects median strategy

- fixes some bugs, breakages and testcases along the way

Issue #199, opened 2021-01-15
feeadjuster: assert can trigger


gallizoltan:
**Description**

I've seen a curious line in the lightningd log:
```
2020-12-06T20:32:06.150Z **BROKEN** plugin-feeadjuster.py: Adjusting fees:
```
I've investigated the situation and figured out what happened. It caused by the assert at line 122 in feeadjuster/feeadjuster.py:
```
assert 0 <= percentage and percentage <= 1
```
**Steps to reproduce**

1. Let's assume I have a channel with a total size of 100,000 Sats. 20,000 Sats are mine in that channel.
2. I call a `feeadjust`. It saves my balance in `plugin.adj_balances[scid]` at line 198.
3. Then an incoming payment happens: 30,000 Sats arrive into the channel. (In my case it was actually a circular payment, rebalance, but it does not matter.)
At this point I have 50,000 Sats in the channel, but `plugin.adj_balances[scid]` still stores the 20,000 Sats value. The `feeadjuster` is not aware of the incoming payments.
4. After these a forward event occurs: 40,000 Sats leave the channel. 
The `feeadjuster` is subscribed to the forward event, so it reduces my stored balance by 40,000 Sats at line 173. The result will be negative: -20,000 Sats. This will trigger the assert.

**A possible solution**

The easy way: it will not happen again if I turn off the automatic fee updates in `feeadjuster`.

m-schmoock:
Good that we put that assert in ;)

m-schmoock:
@gallizoltan  Do we need to fix this by i.e. by replacing the `forward_event_subscription` early return in `forward_event` to check for this in line 178 where `maybe_adjust_fees` is called? Can you make a test for this?

Partly addressed by https://github.com/lightningd/plugins/pull/198  (only rebalance)

gallizoltan:
I'm not sure what would be the best solution. Maybe we should write a `@plugin.subscribe("invoice_payment")` and a `@plugin.subscribe("sendpay_success")`.

chrisguida:
Fixed by #525

PR #525, opened 2024-05-08
feeadjuster: various updates/fixes

daywalker90:
Ok so i was bored and went over the feeadjuster plugin because @grubles said there were problems with it.

I found a compatibility issue in #457 with `listconfigs` so i made a wrapper function to support both the old and new structure of `listconfigs`.
I went over all the feeadjuster issues that were reporting bugs and most of them are probably users who updated their node but not their `pyln-client`, since we can't raise the version in `requirements.txt` ourselves because that would also break compatibility (maybe fixed by [this](https://github.com/ElementsProject/lightning/pull/7173) ?).

But i also found an actual issue in #199 which i fixed by always getting the channel balance from a rpc call to `listpeers`/`listpeerchannels` instead of doing math with the data from the `forward_event`. We need to do this because between forwards there can be invoices getting paid, payments sent out or channel splices. Listening for these events is rather pointless since they don't contain data about what channel balances changed, so we would have to do rpc calls anyways. Unfortunately `listpeers`/`listpeerchannels` does not update quite as fast as the `forward_event` so we have to wait a bit.

Then @grubles pointed me to the PR #467 from @m-schmoock and i incorporated it together with my fixes.

Then i tested `feeadjuster` against CLN [23.02.2](https://github.com/daywalker90/plugins/actions/runs/9008084960), [23.05.2](https://github.com/daywalker90/plugins/actions/runs/9008084965), [23.08.1](https://github.com/daywalker90/plugins/actions/runs/9008084979), [23.11.2](https://github.com/daywalker90/plugins/actions/runs/9008084968) and [24.02.2](https://github.com/daywalker90/plugins/actions/runs/9008084957)

I had to adjust the tests a bit to accomodate for the bad gossip retransmission in older CLN versions (it's still slightly flaky but who cares)

<sub>just rewrite it in rust lul</sub>

### DIFF
--- a/feeadjuster/README.md
+++ b/feeadjuster/README.md
@@ -22,3 +22,4 @@ the median fees from peers of peer.
 'median'. This allows over- or underbidding other nodes by a constant factor (default: 1.0).
 - `feeadjuster-max-htlc-steps` Default 0 (turned off). Sets the number of max htlc adjustment steps. If our local channel balance drops below a step level
 it will reduce the max htlc to that level, which can reduce local routing channel failures.  A value of 0 disables the stepping.
+- `feeadjuster-basefee` Default False, Also adjust base fee dynamically. Currently only affects median strategy.
--- a/feeadjuster/feeadjuster.py
+++ b/feeadjuster/feeadjuster.py
@@ -22,6 +22,17 @@ plugin.mutex = Lock()
 plugin.mutex.acquire()
 
 
+def read_excludelist():
+    try:
+        with open('feeadjuster-exclude.list') as file:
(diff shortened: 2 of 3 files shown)


chrisguida:
Excellent work!

chrisguida:
@m-schmoock care to approve this?

chrisguida:
@telelvis care to approve this?

telelvis:
fantastic work by @daywalker90 , imo. 
I am all for better feeadjuster!

chrisguida:
Closes #457 and #199

PR #530, opened 2024-06-02
Refactor CI

sip-21:
* Refactor Python code:
  * Move duplicate code from _test.py_ and _update_badges.py_ into new file _utils.py_
  * Add some type annotations
* Fix `setup-python` action in workflow files
* Cleanup

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -119,12 +119,12 @@ jobs:
       - uses: actions/checkout@v4
         with:
           submodules: recursive
-      - name: Set up Python ${{ matrix.python-version }}
+      - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: "3.12"
       - name: Complete
         run: |
-          echo "Updating badges data..."
+          echo "Updating badges data for main workflow..."
           python3 .ci/update_badges.py main 5  # We test for 5 distinct Python versions
-          echo "CI completed"
+          echo "CI completed."
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -104,12 +104,12 @@ jobs:
       - uses: actions/checkout@v4
         with:
           submodules: recursive
-      - name: Set up Python ${{ matrix.python-version }}
+      - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: "3.12"
       - name: Complete
         run: |
-          echo "Updating badges data for nightly..."
+          echo "Updating badges data for nightly workflow..."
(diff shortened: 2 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -119,12 +119,12 @@ jobs:
       - uses: actions/checkout@v4
         with:
           submodules: recursive
-      - name: Set up Python ${{ matrix.python-version }}
+      - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: "3.12"
```
chrisguida, line 125: Why are we changing to only test 3.12?

.github/workflows/nightly.yml
```diff
@@ -104,12 +104,12 @@ jobs:
       - uses: actions/checkout@v4
         with:
           submodules: recursive
-      - name: Set up Python ${{ matrix.python-version }}
+      - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: "3.12"
```
chrisguida, line 110: Same comment here

.github/workflows/main.yml
sip-21: This is just the Python version that is used in the `gather` step for running _update_badges.py_ and NOT the one for running the tests. We could also keep the version unspecified, and the version of Python currently in `PATH` will be used, I guess.

sip-21: We want to get rid of this warning here: https://github.com/lightningd/plugins/actions/runs/9331506874
image

chrisguida: Ah ok, I see. So we want to keep `update_badges.py` in sync with the latest version of Python? What version is in the PATH? Is this not populating because `matrix.python_version` only exists in the `build-and-test` step and not in the `gather` step?

sip-21: Yes, `matrix.python_version` is outside the scope of the `gather` step and therefore the value for `python-version` was missing, as I see it. It is clearly a bug. Here is another log output to clarify:
image

To fix this, I specified the latest stable Python version. 3.12. In fact, the recommended way is to set the Python version explicitly:
> The python-version input is optional. If not supplied, the action will try to resolve the version from the default .python-version file. If the .python-version file doesn't exist Python or PyPy version from the PATH will be used. The default version of Python or PyPy in PATH varies between runners and can be changed unexpectedly so we recommend always setting Python version explicitly using the python-version or python-version-file inputs.

https://github.com/actions/setup-python

chrisguida: Ok sounds good!

chrisguida:
ACK `latest commit`

chrisguida (approved):

PR #519, opened 2024-05-02
add bolt12-prism as submodule

gudnuf:

PR #532, opened 2024-06-04
Add badges for currencyrate plugin

sip-21:
* Add `main` and `nightly` badges for `currencyrate` plugin on _README.md_


chrisguida:
Do we need to do an update like this every time we add tests?

Where are you getting the new badge URLs from?

Perhaps we can just mark all the plugins with missing tests as failing?

sip-21:
We could add URLs for all plugins, but it will not look nice. It will display a red badge with the "Resource missing" or similar text. 

But yes, thinking more about it, we could add default JSON data in the _.badges_ branch for displaying badges for those plugins that still have no tests.

chrisguida:
Let's add badges for `bolt12-prism` as well https://github.com/lightningd/plugins/pull/519

chrisguida:
Merging this for now, we can do something more complicated later :)

ACK 5911628

chrisguida (approved):

PR #534, opened 2024-06-05
holdinvoice: update to v2.0.0

daywalker90:
# Changelog

### Added

- Newly supported invoice fields ``exposeprivatechannels`` and ``created_index``

### Changed

- Error on not decodeable ``fallbacks`` in ``HoldInvoiceRequest``'s instead of skipping them
- ``HoldState`` is now a primitive inside primitives.proto
- ``HoldState`` is now consistently UPPERCASE everywhere if represented as a string

### Removed

- ``DecodeBolt11``, it is now supported by ``cln-grpc`` to have routes in the decode/decodepay responses

### Fixed

- Crash in cln v24.05 caused by missing `failure_message` in hook response with `result`:`fail`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -81,6 +81,27 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
+    - name: Get changed files
+      id: get_changed_files
+      if: ${{ github.event_name == 'pull_request' }}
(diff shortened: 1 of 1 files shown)


chrisguida (approved):
ACK dda4677

This should get the plugins repo CI back to passing!

PR #535, opened 2024-06-05
sling: update to v2.0.0

daywalker90:
# Changelog

### Added

- ``sling-refresh-gossmap-interval``: How often to read ``gossip_store`` updates in seconds. Default is every ``10``s

### Changed

- ``sling`` now reads the ``gossip_store`` file directly instead of polling the whole graph via ``listchannels``. This drastically reduces CPU usage on and especially after the first read. Also, since ``sling`` is reading incremental changes to the ``gossip_store`` file, it can poll it much more often to keep a closer track of any gossip changes.
- Options code refactored. All options are now natively dynamic and there is no longer any manual reading of config files. Read the updated README section on how to set options for more information

### Removed

- ``sling-refresh-graph-interval``, replaced by ``sling-refresh-gossmap-interval``

### Fixed

- Restored backward compatibility to v23.11
- When running ``paralleljobs``>1 there were still duplicate routes taken sometimes


daywalker90:
let me fix that flaky test

daywalker90:
the one time regtest in ci is too fast instead of too slow 😄  should be fixed now

chrisguida:
ACK 1c89ad3

PR #539, opened 2024-06-05
circular: add tests

daywalker90:
apparently also update it


chrisguida:
ACK 8e52447

Note that this will break the CI on both main and nightly, so hopefully @giovannizotta will fix the `circular-push` bug soon.

If not, we'll move `circular` to archived in the next week or two.

daywalker90:
This will require a general fix too. Notice how the test here fails earlier than the push test because circular needs a general update for cln 24.05

PR #542, opened 2024-06-05
currencyrate: add badges

daywalker90:
We already have tests for currencyrate, just missing the badges. Will run tests on full repo so  expect failure.


chrisguida:
Duplicate of #532

PR #543, opened 2024-06-05
CI: ignore some paths from tests

daywalker90:
Lets not run tests on these files/directories. There is no point


daywalker90:
ironically this will run the full repo tests since the ci file got changed (we want this tho)

chrisguida:
This will save a ton of time on non-plugin fixes!

chrisguida:
ACK 34f99c0

PR #546, opened 2024-06-05
new plugins manager category

daywalker90:
We were still pointing to the old external reckless repo 💀 

So i took the opportunity to create an extra plugins manager category and moved reckless there with a link to the cln docs. I also added coffee to the new category.


vincenzopalazzo (approved):
ACK f836e4d21076e036cd33a5f7642c23af1fef67bb

chrisguida (approved):
Yep, it makes sense to move reckless off the list of plugins, since it isn't a plugin.

And creating a new category for plugin managers makes sense because there are multiple plugin managers and these are powerful tools for interacting with plugins.

ACK 2590655

PR #548, opened 2024-06-06
CI: refactor to return success on ignored paths

daywalker90:
I moved the check for changed files up and used it to check if on a PR there are files outside the ignored ones changed. Then we onlry run steps if we are pushing or a pr with files changed outside the ignored ones. This gets the "README pr"'s down to <1min, usually ~30s


chrisguida:
ACK 7d928c4

PR #549, opened 2024-06-06
CI: add cache for cln, bitcoind and python dependencies

daywalker90:
Right now we download bitcoind, cln and python dependencies 5x each main CI run. 

In this PR i introduce three caches for bitcoind, cln and python dependencies. For this we explicitly state the cln version to use. This will also enable us in the future to add more cln versions but we have to bump it on a new release.

Usually the save cache action is not needed but because our `pytest_tests` step might regularly fail we need the early save to keep the cache.

This brings down main CI runs by 1-2 minutes and saves alot of bandwith for everyone.

The python dependencies of each plugin do not get cached as that should get tested each run.


daywalker90:
Oh and the cache gets only saved on push events because only caches on master branch can be used by all PR's

daywalker90:
I made a small mistake that needs cleanup: under the actions page go to caches and delete the ones not belonging to the master branch

chrisguida (approved):
Great work!

ACK a01966d226b3d219da2f4cf58fe1fb0a4adcc8c5

PR #551, opened 2024-06-06
CI: add retry logic for push

daywalker90:
nightly was having troubles pushing the results to the badges branch. (notice how holdinvoice is still marked failing on nightly, despite all tests passing). This retry logic will fix this.


daywalker90:
btw main is just hiding this problem by running more often, they use the same code

chrisguida (approved):
LGTM

ACK 6abe780

PR #552, opened 2024-06-07
CI: support make check test framework

daywalker90:
Addresses discussion in https://github.com/LNOpenMetrics/go-lnmetrics.reporter/pull/147

Adds support for real (real because some use Makefiles but don't have a check command) `Makefile` enjoyers. They can have a `tests/setup.sh` incase they need to do something before running `make check`.

Currently `go-lnmetrics.reporter` is the only plugin in the repo which will match this new type.


daywalker90:
someone show me `make check` testing for compatibility with current cln releases, until then i'll close this

PR #556, opened 2024-06-10
Add default badges to README

sip-21:
This PR adds badge links for all plugins that still don't have badges displayed to the README file, e.g. because they are missing tests.

Another PR https://github.com/lightningd/plugins/pull/555 provides the JSON data for the corresponding plugins inside the `badges` branch. 

image


chrisguida (approved):
Is there any way to keep these dynamically updated? ie if tests are removed?

This is fine for now.

ACK d0a43f4

sip-21:
> Is there any way to keep these dynamically updated? ie if tests are removed?
> 
> This is fine for now.
> 
> ACK [d0a43f4](https://github.com/lightningd/plugins/commit/d0a43f4cc1cedf232040ffe6b7428e16f4bc2524)

Actually, I thought to add some badges update logic to this PR. So, let's do it in a separate one :+1:

Issue #494, opened 2024-02-09
[feat]: Display failing / non-tested plugins in a matrix with dynamic badges instead of moving to archive


chrisguida:
Research whether it's possible to add some columns to the current "available plugins" list, where one column is "passing CI" and the other is "has tests", and dynamically update these for each plugin so users can quickly determine the state of each plugin's maintenance at a glance.

If they're failing CI we obviously need a way to remove them from testing so other plugins can use the CI.

cdecker says:

>I wonder if we can trick the badge generators into treating each plugin as a separate branch or tag

chrisguida:
Part of this is done in #509 

We should decide what we want to do with failing plugins.

Currently we archive them after a short time because they cause the CI job to fail.

But we could remove the CI job failure and simply allow them to display a red failure badge.

I'm partial to keeping it how it is now, but I'm not sure which is best.

Perhaps we should add a few more prior versions of CLN to the CI (and remove some Python versions) and let plugins fail for a bit longer before we count them as archived?

@cdecker @vincenzopalazzo @m-schmoock @daywalker90 @sip-21 thoughts?

daywalker90:
This is what we probably want: CI only fails if a plugin goes from previously passing->failing. If it was already failing, CI passes. In this way we get alerted once and then CI is chill again.

I agree we should test more cln versions and add them to the table. If a plugin is failing the last x versions it gets archived. In this way users running older cln versions can get value from the plugins repo aswell.

chrisguida:
Hmm that definitely sounds useful, I wonder how difficult it would be to implement, do you want to take a crack at it?

>I agree we should test more cln versions and add them to the table. If a plugin is failing the last x versions it gets archived. In this way users running older cln versions can get value from the plugins repo aswell.

Cool, I'll add another version or two

PR #559, opened 2024-06-15
CI: test more cln versions

daywalker90:
Replaces: #545
Fixes: #494

I included some drive-by fixes for the push retry logic and uncommented the check that makes sure all tested python versions passed the tests.

Also added some code to auto-create the badges for plugins that have no tests. For this i had to add the remaining missing submodules and fix their names.

With this we now test the latest three CLN versions but only two python versions (oldest and newest from before) when running the CI for pull requests and commits to master. This is done to keep a reasonable CI matrix size.

Nightly remains unchanged and tests master cln with five python versions.

Preview:

image


daywalker90:
Changed the way `get_testfiles` to mirror the way pytest detects test files aswell. Since there could be plugins with a tests folder but no pytest tests

cdecker (approved):

chrisguida:
Looks great! Just has some conflicts

chrisguida (approved):
ACK https://github.com/lightningd/plugins/commit/6a4b72bf27c0f7f9fac8fbb51eb9b2bd03e84288

PR #563, opened 2024-06-19
refactor get_testfiles

daywalker90:
Currently we are checking the _badges_ branch for testfiles when calling `collect_gather_data`-> `has_testfiles`. That branch is and will not be up2date with the master branch so we should avoid that. And instead of switching branches around i rather save the testfiles once inside the plugin.

### DIFF
--- a/.ci/test.py
+++ b/.ci/test.py
@@ -8,8 +8,7 @@ import json
 from itertools import chain
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


cdecker:
Uh, nice catch, I hadn't thought about the branches as context 🤔

chrisguida:
ACK https://github.com/lightningd/plugins/commit/ba3636c68196f12f1904e89b143f62811efc4221

PR #564, opened 2024-06-21
CI: downgrade main python to 3.11, remove unwanted test results

daywalker90:
Also invalidates the cache a bit more often: on minor python version updates and whenever ubuntu-latest goes to a new version. 

Also fixes two bugs: 

- we were calling `enumerate_plugins` on the badges branch and not on the master branch
- duplicate `git push` command after the retry loop exits


chrisguida:
ACK https://github.com/lightningd/plugins/pull/564/commits/fee25bf3b9c9f2e6f74d9801712b7b614e0bc1f8

PR #567, opened 2024-06-30
CI: nightly fixes

daywalker90:
Right now all plugins fail the nightly python 3.12 tests and are therefore marked as failing for master despite most plugins actually working fine on master.

This is because python dependency management is a mess. The problem was coincurve 18 no longer working in python 3.12.4. We got it updated in CLN but until there is a new python packages release this alone doesn't do anything yet. We should install the dependencies in development mode, which installs them from source. I think this is what we actually want since the purpose is to test the plugins against any changes in CLN and the python framework on master.

You'd think that a simple `poetry install --with=dev` would suffice but you'd be mistaken. Poetry will only install top level dependencies in dev mode and not sub-dependencies. With this you'll still get coincurve 18. The way i did it was to use the requirements.txt from the `poetry export --with=dev` command to setup the pip and generic plugin's test environment.

If a plugins fails to setup it's environment it will now fail the test instead of cancelling the nightly workflow.

I've added the `workflow_dispatch` trigger, so we can run nightly manually for testing purposes.

### DIFF
--- a/.ci/test.py
+++ b/.ci/test.py
@@ -21,18 +21,18 @@ global_dependencies = [
 pip_opts = ['-qq']
 
 
-def prepare_env(p: Plugin, directory: Path) -> bool:
+def prepare_env(p: Plugin, directory: Path, env: dict, workflow: str) -> bool:
     """ Returns whether we can run at all. Raises error if preparing failed.
     """
     subprocess.check_call(['python3', '-m', 'venv', '--clear', directory])
     os.environ['PATH'] += f":{directory}"
 
     if p.framework == "pip":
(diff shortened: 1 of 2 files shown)


daywalker90:
https://github.com/daywalker90/plugins/actions/runs/9744577335 this is a nightly run on [b18f2f4](https://github.com/lightningd/plugins/pull/567/commits/b18f2f4cef6d4a8aef65ba22c8d343824d1fcc2d)

chrisguida (approved):
Thanks for fixing!

ACK b18f2f4cef6d4a8aef65ba22c8d343824d1fcc2d

PR #578, opened 2024-07-31
Add torq plugin

Jhoyola:
I hope it's ok to have the repository name different from the submodule directory name. We already have a repository called Torq. Otherwise I can rename the plugin.


cdecker:
Thanks @Jhoyola, I'm looking forward to giving the plugin a go. It's perfectly fine to have the plugin name not match the repository name, but maybe add a `-plugin` suffix so people don't get the impression that the plugin is the entirety of torq rather than a way to hook torq into CLN more closely.

Jhoyola:
@cdecker done!

Jhoyola:
Btw. For now it has just "channel open interceptor" = channel acceptor. But we will for sure add htlc interceptor later too.

chrisguida:
Hi @Jhoyola, this plugin looks very cool! Would you mind adding some integration tests with CLN so the CI can automatically notify us when compatibility with the plugin breaks on master? We're trying to move towards a situation where all new plugins (and eventually all old plugins) include such tests. Right now our plugins use pytest with pyln-testing, but feel free to use any framework that suits your needs if you would like to integrate it into our CI.

Jhoyola:
@chrisguida Sure. I'll add some.

chrisguida:
@Jhoyola thanks so much! Now let's see if we can get some testers to let us know what they think :)

I agree, pyln-testing is amazing for writing tests, IMO tests are actually what Python is best at :)

Jhoyola:
It seems that the CI errors with:
`error starting plugin '/home/runner/work/plugins/plugins/torq-plugin/torq-plugin.py': opening pipe: Permission denied`
Any idea why?

chrisguida:
@Jhoyola try making the plugin executable in git

ca-ruz:
This plugin seems to be a fallback for when torq is down. Is that correct? Does this plugin do anything else or are there plans to add more functionality in the future?

Jhoyola:
The goal of the plugin is to add features to Torq that cannot be done with the CLN gRPC (that Torq normally uses with CLN). The only current feature is the "channel acceptor" with which it's possible to create automation in Torq to dynamically filter for which incoming channels are allowed to be opened.

We aim to add more features in the future, like htlc interceptor, with which you can choose to block some htlcs based on Torq automations.

The automations look like this in Torq https://docs.torq.co/hc/docs/articles/1705659548-how-to-use-automation-workflows

Jhoyola:
@chrisguida I added the execute permission and squashed into 1 commit. Could you run the CI again please.

Jhoyola:
CI seemed to fail with 23.11. I think its this https://github.com/ElementsProject/lightning/issues/6078 . `join_nodes `in pyln-testing seems to not work for 23.11. I tried locally testing just that and it gives the same error.

I pushed one more little fix for plugin to exit gracefully. I hope it's ok to merge even with 23.11. failing.

chrisguida:
@Jhoyola yep, that's fine, 23.11 will just show as a red X on the main README so people know not to torq-plugin with that version. CI working as designed :)

Eventually if recent CLN versions start failing tests that will give us a heads up that the plugin has stopped being maintained.

ACK https://github.com/lightningd/plugins/pull/578/commits/2653eeef916fb5e989823b89f5efb3954debae41

@cdecker thoughts?

cdecker:
Apologies for the delay, I'm loving this, let's merge as soon as CI is green, and then we can iterate should there be any nits to be addressed.

Jhoyola:
I made a fix to go around the problem with pyln-testing `node_factory.join_nodes` hanging on v23.11. Now all versions should pass.

Jhoyola:
@cdecker The CI should now be all green. If you could run the CI, and hopefully merge.

cdecker:
Done ^^

PR #580, opened 2024-08-06
clnrod: fix tests for 24.08+

daywalker90:
Master gets upset now because `experimental-anchors` is now a default enabled option. I've added a version check to the tests and only enable it explicitly on older versions.


chrisguida (approved):
ACK https://github.com/lightningd/plugins/pull/580/commits/731979437fa4c0e3e4397243fdf144e10a8ab5e5

PR #583, opened 2024-08-22
Add hold plugin

michael1011:
@chrisguida because I did not use `pyln-testing`, I am not sure what I am supposed to do about the CI icons


chrisguida:
For the badges on the README, just follow the pattern the other plugins use for each icon. These will be created automatically for `hold` by the CI.

As for whether we should merge this without working integration tests, I'm reluctant to introduce new plugins to this repo where it's not possible to tell whether the plugin has been abandoned. We already have a ton of unmaintained/broken plugins in this repo and this just increases the load on maintainers while reducing the trust people have in the repo.

@cdecker what do you think?

cc @daywalker90

cdecker:
Very nice addition, and very welcome given that many have been asking for this. I agree with @chrisguida, that we should at least have some smoke tests to ensure it doesn't bitrot silently.

michael1011:
@daywalker90 the limiting factor for our old hold invoice plugin was not CPU usage but responsiveness. It was using datastore as persistent storage, which started to act up and have random slowdowns after hundreds of thousands of invoices in it. Not sure why, but a much sensible approach is the plugin using its own SQLite or PostgreSQL database where one can easily clean cancelled and expired invoices and have much more flexibility than dumping everything as KV store in the database of CLN itself.

I did not test with hundreds of HTLCs but my educated guess would be 0% CPU usage because I am not using a busy lock with a sleep like your plugin, but the oneshot channels of Tokio to respond to HTLC acceptance hooks.

michael1011:
> I agree with @chrisguida, that we should at least have some smoke tests to ensure it doesn't bitrot silently.

I guess I can add some basic functionality tests for `pyln-testing`. Is there a way to tell the CI to only execute certain Python files for testing or will I have to work with environment variables (eg only run my regtest tests if BOLTZ_REGTEST=1)?

daywalker90:
> @daywalker90 the limiting factor for our old hold invoice plugin was not CPU usage but responsiveness. It was using datastore as persistent storage, which started to act up and have random slowdowns after hundreds of thousands of invoices in it. Not sure why; anyways a much sensible approach is the plugin using its own SQLite or PostgreSQL database where one can easily clean cancelled and expired invoices and have much more flexibility than dumping everything as KV store in the database of CLN itself.
> 
> I did not test with hundreds of HTLCs but my educated guess would be 0% CPU usage because I am not using an endless loop with a sleep like your plugin, but the oneshot channels of Tokio to respond to HTLC acceptance hooks.

Ah very intersting. My plugin cleans up any extra data from cln's database after the invoice was paid or expired, unless we are talking about hundreds of thousands of active, unpaid invoices at the same time 😄. How does your plugin handle invoice expiries  / HTLC expiries while holding HTLC's?



> I guess I can add some basic functionality tests for `pyln-testing`. Is there a way to tell the CI to only execute certain Python files for testing or will I have to work with environment variables (eg only run my regtest tests if `BOLTZ_REGTEST=1`)?

The CI of this repo will try to run `pytest` on any `.py` files that start with `test_` in either the root or `tests` directory.

michael1011:
> My plugin cleans up any extra data from cln's database after the invoice was paid or expired

We keep data about paid invoices to be able to figure which channels are useful for receiving and just in case there is some bug, you really want all information you can get your hands on. Expired ones we could delete I guess (unless they have information about cancelled HTLCs which could also be useful; but maybe that just me being a data hoarder), but that's tricky because I don't want to iterate over the entirety of my namespace in datastore and don't want to maintain a secondary index. A `DELETE FROM invoices WHERE status = 'expired'` is perfect -> SQL database it is. And I guess I should open an issue for pagination for `listdatastore`; could be useful in the future.

> How does your plugin handle invoice expiries / HTLC expiries while holding HTLC's?

It holds on for dear life until force close. You have to tell it to settle or cancel via RPC or gRPC

michael1011:
> The CI of this repo will try to run pytest on any .py files that start with test_ in either the root or tests directory.

That works. I will just move my, non `pyln-testing`, tests into a folder that is not called `tests`

michael1011:
@daywalker90 there is no Rust compiler in the CI, so how did you manage to start the plugin? I browsed through your repo and you are downloading artifacts? Is that the best way to go about this?

daywalker90:
Yes, we don't want to compile plugins in this repo. It would make the CI runs take hours eventually. You can use a `setup.sh` inside your `tests` folder to get the binary for your tests.

michael1011:
@chrisguida I added some very basic tests with `pyln-testing` to ensure compatibility with different CLN versions and rebased against master to resolve the conflict. Could you trigger another CI run?

cdecker:
> @daywalker90 the limiting factor for our old hold invoice plugin was not CPU usage but responsiveness. It was using datastore as persistent storage, which started to act up and have random slowdowns after hundreds of thousands of invoices in it. Not sure why; anyways a much sensible approach is the plugin using its own SQLite or PostgreSQL database where one can easily clean cancelled and expired invoices and have much more flexibility than dumping everything as KV store in the database of CLN itself.
> 
> I did not test with hundreds of HTLCs but my educated guess would be 0% CPU usage because I am not using an endless loop with a sleep like your plugin, but the oneshot channels of Tokio to respond to HTLC acceptance hooks.

Very interesting indeed for the datastore API. Did you make use of the append modes and CAS primitives to append historical data to a key? Did you try to shard over multiple keys? I mean having a monthly or daily key in the datastore, then append each time, and you can still query ranges depending on your key granularity. I'm not surprised that for heavy usage datastore isn't perfect, and using a fully fledged DB for professional applications is probably the best idea, but we might learn something from your experience here and maybe formulate some good practices for the small to medium use cases.

michael1011:
> Did you make use of the append modes and CAS primitives to append historical data to a key?

I used `must-create` explicitly to save new hold invoices and `must-replace` to update records.

> Did you try to shard over multiple keys?

I was thinking about it, but decided not to. hold having it's own SQL database seemed more reasonable than somehow trying to make datastore work. A plugin having it's own database is not unheard of; I am already doing it for mpay and bkpr, which ships with CLN, does it too. The main intention is interacting *as little as possible* with CLN in the plugin. We talked about the "database pauses" of CLN before (basically CLN freezes for a couple seconds when PostgreSQL does autovacuum). That is minimized by hold having it's own database because it does not need to talk to datastore and there being less data in the database of CLN.

chrisguida:
Looks great @michael1011!!

ACK https://github.com/lightningd/plugins/pull/583/commits/78b5d23ab2a9c1e37a1d3128530f4351f314c111

Merging this in the next day or two unless I hear any objections

cc @cdecker @daywalker90 @vincenzopalazzo

PR #584, opened 2024-09-03
sauron: Add mutinynet and tests

ca-ruz:
Adds mutinynet support to sauron. Mutinynet.com uses the mempool.space REST API, which has a slightly different spec from blockstream's Esplora so there is new code to handle that.

We pinned pyln-client to <=24.5 because version 24.8 has some extra requirements that sauron does not meet, we should probably fix this but for now pinning 24.5 should work fine.


sauron/sauron.py
```diff
@@ -38,11 +34,12 @@ def fetch(url):
 
     return session.get(url)
 
-
 @plugin.init()
-def init(plugin, options, configuration, **kwargs):
-    plugin.api_endpoint = options["sauron-api-endpoint"]
-    if not plugin.api_endpoint:
+def init(plugin, options, **kwargs):
+    plugin.sauron_api_endpoint = options.get("sauron-api-endpoint", None)
```
chrisguida, line 39: Why rename this variable?

sauron/sauron.py
```diff
@@ -58,11 +55,11 @@ def init(plugin, options, configuration, **kwargs):
     plugin.log("Sauron plugin initialized")
     plugin.log(sauron_eye)
 
-
 @plugin.method("getchaininfo")
 def getchaininfo(plugin, **kwargs):
-    blockhash_url = "{}/block-height/0".format(plugin.api_endpoint)
-    blockcount_url = "{}/blocks/tip/height".format(plugin.api_endpoint)
+    blockhash_url = "{}block-height/0".format(plugin.sauron_api_endpoint)
```
chrisguida, line 60: Why remove the slash? (`/`)

sauron/sauron.py
```diff
@@ -90,27 +87,30 @@ def getchaininfo(plugin, **kwargs):
         raise SauronError("Unsupported network")
     plugin.sauron_network = chains[genesis_req.text]
 
-    # We wouldn't be able to hit it if its bitcoind wasn't synced, so
-    # ibd = false and headercount = blockcount
```
chrisguida, line 94: Why remove this comment?

sauron/sauron.py
ca-ruz: I renamed it because I saw `options["sauron-api-endpoint"]` and I thought the variable name needed to match that. Now I understand the options specify sauron for the user to know they are related to sauron since its running on core lightning but there was no need to change the variable name since that is only in the sauron code. Will revert back those changes.

sauron/sauron.py
ca-ruz: I removed them because there is slash in the API address when running sauron, /api/. I thought there could be an issue when making the API calls,  ex. /block-height.  I did some testing and it does not matter, the API call works fine but I still removed it because I thought it was redundant. Should I leave them as before for readability?

sauron/sauron.py
ca-ruz: That was just a mistake. I´ll add it again.

sauron/sauron.py
chrisguida: Yeah I would say generally don't change stuff unless you're sure it's an improvement.

Seems like if the user leaves off the slash in the option and there's no slash added here either, there could be an annoying error where the user doesn't realize they need to add a slash at the end of the URL they pass into the `sauron-api-endpoint` option. But if there are two slashes, I think Python figures this out. So I think there is more of a reason to leave as is than to change it.

ca-ruz: Yeah, that makes sense. Python does figure it out and it works just fine with two slashes. I hadn't thought  of the scenario where the user would not add the slash to the endpoint option and then none of the API calls would work. I'll add them as it was before since it would always work regardless of how the user pass the endpoint.

sauron/sauron.py
```diff
@@ -225,11 +290,11 @@ def estimatefees(plugin, **kwargs):
         "feerates": feerates
     }
 
-
 plugin.add_option(
     "sauron-api-endpoint",
     "",
     "The URL of the esplora instance to hit (including '/api').",
+    "The URL of the mutinynet instance to hit (including '/api').",
```
chrisguida, line 297:
```suggestion
```

sauron/sauron.py
```diff
@@ -154,64 +165,118 @@ def sendrawtx(plugin, tx, **kwargs):
...
+
+    # Construct the final fee rates structure
     feerates = [
         {"blocks": 2, "feerate": very_urgent},
         {"blocks": 6, "feerate": urgent},
         {"blocks": 12, "feerate": normal},
         {"blocks": 144, "feerate": slow}
     ]
 
+    # Return the estimated fees
```
chrisguida, line 279:
```suggestion
```

sauron/sauron.py
```diff
@@ -154,64 +165,118 @@ def sendrawtx(plugin, tx, **kwargs):
...
-        very_urgent = int(feerates["2"] * 10**3)
+        slow = int(feerates["144"] * multiply_factor)
+        normal = int(feerates["12"] * multiply_factor)
+        urgent = int(feerates["6"] * multiply_factor)
+        very_urgent = int(feerates["2"] * multiply_factor)
 
-    feerate_floor = int(feerates["1008"] * 10**3)
+    # Handle missing '1008' key with a default value (e.g., `slow` as fallback)
+    feerate_floor = int(feerates.get("1008", slow) * multiply_factor)
+
```
chrisguida, line 270:
```suggestion
```

sauron/sauron.py
```diff
@@ -154,64 +165,118 @@ def sendrawtx(plugin, tx, **kwargs):
...
+    # Construct the final fee rates structure
```
chrisguida, line 271:
```suggestion
```

sauron/sauron.py
```diff
@@ -154,64 +165,118 @@ def sendrawtx(plugin, tx, **kwargs):
...
-        normal = int(feerates["12"] * 10**3)
-        urgent = int(feerates["6"] * 10**3)
-        very_urgent = int(feerates["2"] * 10**3)
+        slow = int(feerates["144"] * multiply_factor)
+        normal = int(feerates["12"] * multiply_factor)
+        urgent = int(feerates["6"] * multiply_factor)
+        very_urgent = int(feerates["2"] * multiply_factor)
 
-    feerate_floor = int(feerates["1008"] * 10**3)
+    # Handle missing '1008' key with a default value (e.g., `slow` as fallback)
```
chrisguida, line 268:
```suggestion
```

sauron/sauron.py
```diff
@@ -130,16 +134,23 @@ def getrawblock(plugin, height, **kwargs):
...
+        "blockhash": block_hash,
+        "block": block_data,
     }
 
 
 @plugin.method("sendrawtransaction")
 def sendrawtx(plugin, tx, **kwargs):
     sendtx_url = "{}/tx".format(plugin.api_endpoint)
-
+
```
chrisguida, line 153:
```suggestion

```

sauron/sauron.py
```diff
@@ -99,18 +98,23 @@ def getchaininfo(plugin, **kwargs):
         "ibd": False,
     }
 
-
 @plugin.method("getrawblockbyheight")
 def getrawblock(plugin, height, **kwargs):
+    # Step 1: Get the block hash by height
```
chrisguida, line 103:
```suggestion
```

sauron/sauron.py
```diff
@@ -99,18 +98,23 @@ def getchaininfo(plugin, **kwargs):
...
     blockhash_url = "{}/block-height/{}".format(plugin.api_endpoint, height)
+
```
chrisguida, line 105:
```suggestion
```

sauron/sauron.py
```diff
@@ -58,11 +57,11 @@ def init(plugin, options, configuration, **kwargs):
     plugin.log("Sauron plugin initialized")
     plugin.log(sauron_eye)
 
-
 @plugin.method("getchaininfo")
 def getchaininfo(plugin, **kwargs):
     blockhash_url = "{}/block-height/0".format(plugin.api_endpoint)
     blockcount_url = "{}/blocks/tip/height".format(plugin.api_endpoint)
+
```
chrisguida, line 64:
```suggestion
```

sauron/sauron.py
```diff
@@ -2,24 +2,22 @@
...
 plugin.sauron_network = "test"
 
-
 class SauronError(Exception):
     pass
 
-
 def fetch(url):
-    """Fetch this {url}, maybe through a pre-defined proxy."""
+    """Fetch the given {url}, maybe through a pre-defined proxy."""
```
chrisguida, line 20:
```suggestion
    """Fetch this {url}, maybe through a pre-defined proxy."""
```

sip-21:
You need to pin the version of `pyln-client` in requirements.txt:
`pyln-client==24.5`

sip-21:
You might want to add some instructions in the README for how to start sauron with Mutinynet. Below command worked for me once I had `pyln-client==24.5` installed (see comment above):
`lightningd --signet --disable-plugin bcli --plugin $PWD/sauron.py --sauron-api-endpoint https://mutinynet.com/api/`

sauron/sauron.py
```diff
@@ -38,14 +36,33 @@ def fetch(url):
...
+    plugin.log("plugin.api_endpoint = %s" % plugin.api_endpoint)
+    
     if not plugin.api_endpoint:
         raise SauronError("You need to specify the sauron-api-endpoint option.")
         sys.exit(1)
 
+    # Test for Esplora or mempoolspace
+    try:
+        # MutinyNet API
+        feerate_url = "{}/v1/fees/recommended".format(plugin.api_endpoint)
```
chrisguida, line 51: I would swap the order of mutinynet and blockstream, hitting blockstream first, since that will be the default for most users.

sauron/sauron.py
```diff
@@ -154,64 +182,106 @@ def sendrawtx(plugin, tx, **kwargs):
...
+        # Blockstream API
+        gettx_url = "{}/tx/{}".format(plugin.api_endpoint, txid)
+        status_url = "{}/tx/{}/outspend/{}".format(plugin.api_endpoint, txid, vout)
 
+    # Fetch the list of UTXOs for the given address
     gettx_req = fetch(gettx_url)
     if not gettx_req.status_code == 200:
         raise SauronError(
-            "Endpoint at {} returned {} ({}) when trying to " "get transaction.".format(
+            "Endpoint at {} returned {} ({}) when trying to get transaction.".format(
```
chrisguida, line 199: :+1:

sauron/sauron.py
```diff
@@ -225,11 +295,11 @@ def estimatefees(plugin, **kwargs):
         "feerates": feerates
     }
 
-
 plugin.add_option(
     "sauron-api-endpoint",
     "",
     "The URL of the esplora instance to hit (including '/api').",
+    "The URL of the mutinynet instance to hit (including '/api').",
```
chrisguida, line 302:
```suggestion
    "The URL of the esplora or mempool.space instance to hit (including '/api').",
```

chrisguida (changes requested):
See comments

sauron/requirements.txt
```diff
@@ -1,2 +1,2 @@
-pyln-client>=23.2
+pyln-client==24.5
```
chrisguida, line 1:
```suggestion
pyln-client>=23.2,<=24.5
```

chrisguida (changes requested):

sauron/requirements.txt
ca-ruz: Done!

chrisguida:
Btw @sip-21 noticed that this PR breaks mainnet on blockstream.info. Can you look into this @ca-ruz @iorch ?

ca-ruz:
@sip-21 Can you please give me more details about the tests you wrote and what were you trying to do when Sauron failed? I am trying to replicate it to fix it. Thanks!

sip-21:
@ca-ruz The failing GitHub actions/tests are here: https://github.com/sip-21/plugins/actions/runs/11116007452/job/30885476517?pr=15

```
[gw0] [ 20%] PASSED test_sauron.py::test_rpc_getchaininfo 
[gw1] [ 40%] ERROR test_sauron.py::test_rpc_sendrawtransaction_invalid 
[gw2] [ 60%] ERROR test_sauron.py::test_rpc_getrawblockbyheight 
[gw3] [ 80%] ERROR test_sauron.py::test_rpc_getutxout 
[gw1] [ 80%] ERROR test_sauron.py::test_rpc_sendrawtransaction_invalid 
[gw2] [ 80%] ERROR test_sauron.py::test_rpc_getrawblockbyheight 
[gw4] [100%] ERROR test_sauron.py::test_rpc_estimatefees 
[gw3] [100%] ERROR test_sauron.py::test_rpc_getutxout 
[gw4] [100%] ERROR test_sauron.py::test_rpc_estimatefees 
```

The tests run fine together locally, but in GitHub action `lightningd` doesn't seem to not shutdown properly after a test. This is why in the logs you then see
```
  lightningd-0 2024-10-01T00:02:01.508Z **BROKEN** connectd: Failed to listen on socket 127.0.0.1:9735: Address already in use
  Time-out: can't find [re.compile('Server started with public key')] in logs
```

ca-ruz:
@sip-21 Thanks!! I recreated the bug and now it's fixed! Can you please run the tests again and let me know?

chrisguida:
Waiting on @sip-21's PR here https://github.com/ca-ruz/plugins/pull/2#issuecomment-2405818832

chrisguida (approved):

.ci/test.py
```diff
@@ -241,13 +241,14 @@ def run_one(p: Plugin, workflow: str) -> bool:
 
     logging.info(f"Virtualenv at {vpath}")
 
+    num_workers = 1 if p.name == "sauron" else 5
```
chrisguida, line 244: @sip-21 didn't you fix the concurrency? We can get rid of this I think?

chrisguida: If removing this breaks the tests, we should add the changes to this file back, but under the other commit (the one that adds the tests).

sauron/requirements.txt
```diff
@@ -1,2 +1,2 @@
-pyln-client>=23.2
+pyln-client>=23.2,<=24.5
```
chrisguida, line 1: We should create an issue to fix this

chrisguida (changes requested):
We're almost there! Just a couple of minor issues.

.ci/test.py
ca-ruz: Tried removing these lines from the code. Tests are all failing now.

chrisguida:
@sip-21

After adding back concurrency, the tests are failing because they're trying to listen on port 9735... that actually doesn't seem necessary. Can you fix the tests not to require listening on port 9735? (or maybe just listen on a random port)

Screenshot 2024-10-24 at 16 54 40

sauron/requirements.txt
ca-ruz: Created Issue and linked to this PR.

sauron/tests/test_sauron_esplora_bitcoin.py
```diff
@@ -0,0 +1,159 @@
...
+
+class LightningNode(utils.LightningNode):
+    def __init__(self, *args, **kwargs):
+        pyln.testing.utils.TEST_NETWORK = "bitcoin"
+        utils.LightningNode.__init__(self, *args, **kwargs)
+        lightning_dir = args[1]
+
+        self.daemon = LightningD(lightning_dir, None)  # noqa: F405
+        options = {
+            "disable-plugin": "bcli",
```
cdecker, line 22: I don't think you need to override this class at all. Just set the options appropriately when using the `NodeFactory.get_node()`. That'll obviate the need for the subclass and the fixtures.

cdecker (approved):

cdecker:
Please don't override the fixtures and the node class unless you absolutely need to. `options` passed to `NodeFactory.gey_node()` is where you can disable one and enable the other. Furthermore, the tests differ only in the backend used, so you could use pytest's parametrized Tests to reuse the implementation but get the backend as a parameter (they'd run once for each backend then). Finally, you are doing format checking which is great, but we could go one further and actually use the JSON schemas from CLN to ensure all fields are present and types match up.

chrisguida:
@cdecker do you have an example of any pytests that use the JSON schema from CLN to check the fields and types? That sounds amazing!

cdecker:
It's done automatically by instrumenting the pyln-client library to load and check the schemas. Anything going through pyln-client in a test environment should be checked afaik. I don't think notifications and hooks are checked yet (there are no schemas yet), but that'd be an amazing extension covering the docs vs implementation too

ca-ruz:
@chrisguida @cdecker @sip-21 Made some changes to the tests.

chrisguida:
@ca-ruz some tests still failing, looks like blockstream is ratelimiting us

ca-ruz:
@chrisguida  Guess we need to work on a workaround for that!

ca-ruz:
@chrisguida @sip-21 Skipped the sendrawtransaction_invalid test across all backends since we were intentionally sending an invalid transaction to test the API call and response. However, it seems that Esplora is rate-limiting us, likely due to the high frequency of invalid requests.

sauron/tests/test_sauron_esplora_bitcoin.py
ca-ruz: Hey @cdecker it looks like we actually do need to override both the `LightningNode` class and the `LightningD` class. 

We need to override the `LightningNode` class because it doesn't allow us to override `self.daemon` as you can see in this code here: https://github.com/ElementsProject/lightning/blob/e66653fa1d847e28bd812356a762225a44bfae83/contrib/pyln-testing/pyln/testing/utils.py#L780-L783

In addition, we have to override the `LightningD` class because there is no way to instantiate it without the `bcli` options which are set here: https://github.com/ElementsProject/lightning/blob/e66653fa1d847e28bd812356a762225a44bfae83/contrib/pyln-testing/pyln/testing/utils.py#L618-L622

These options if set, crash `lightningd` when `bcli` is disabled.

On top of this, the `LightningD.start`method requires `bitcoin-rpcuser` to be set (which as above crashes `lightningd` when we disable `bcli`). 

So unfortunately we need to leave these classes as they are unless I'm missing something.

chrisguida:
@ca-ruz just make an issue to refactor these tests, and clean up the commits, then let's merge

ca-ruz:
I've created an issue to address this: #622.

chrisguida (approved):
ACK https://github.com/lightningd/plugins/pull/584/commits/569dabb01aef16dea01f39519c3e6da7cb6e540b

#618 and #622 still open from this feature

PR #586, opened 2024-09-03
remove ssh requirement for torq plugin

endothermicdev:
This was causing reckless to fail when users don't have a github credential with ssh access. See
https://github.com/ElementsProject/lightning/issues/7635


daywalker90:
ACK cb4705a

chrisguida:
ACK https://github.com/lightningd/plugins/pull/586/commits/cb4705ab2569ee7397d8d55c5352b34ffac75fcd

PR #591, opened 2024-09-23
summars: update to v3.4.0

daywalker90:
# Changelog

### Added

- nix flake (thanks to @RCasatta)
- New column for the main channels table: ``PERC_US``: the percentage of funds in the channel that belong to us.
- New column ``TOTAL_SATS``: the total size of the channel in sats
- New column ``MIN_HTLC``: the minimum size of an outgoing htlc for that channel in sats
- New column ``IN_BASE``: the base fee in msats set by the channel's peer
- New column ``IN_PPM``: the ppm fee set by the channel's peer

### Changed

- Can only sort by enabled columns, since we don't collect data on some disabled columns for performance reasons
- updated dependencies

### Fixed

- Improved performance if invoices or forwards tables don't have anything to show in the requested time window


chrisguida:
ACK https://github.com/lightningd/plugins/pull/591/commits/ed538465abc73dc1871f4ca3520bbe7c944482d4

PR #588, opened 2024-09-12
ci: update to latest upstream CLN

chrisguida:
Add 24.08
Remove 23.11

PR #594, opened 2024-09-24
CI: upgrade to CLN 24.08, drop 23.11

daywalker90:
replaces #588


chrisguida:
Hmm, what does this mean? 24.08 failing like this:

image

daywalker90:
The CI would check the version of lightningd with `lightningd --version` and then checkout that ref. Unfortunately the release binaries for 24.08.1 have the -modded affix. I've added a commit where we just use the version string from the CI file directly.

chrisguida (approved):
ACK https://github.com/lightningd/plugins/pull/594/commits/ff6370a58188d903608be4742fd1a3961d17186d

PR #598, opened 2024-09-26
clnrest-rs

daywalker90:
ontop of #594 so i don't have to do a follow up

# Changelog

### Added
- log rune id, method and params when access was granted

### Changed
- ``clnrest-cors-origins`` is now a multi option, meaning you have to specify it multiple times if you have multiple values, you must no longer feed it multiple values in a single string. This change was missing to be a drop-in replacement with clnrest.py


chrisguida:
Tested on my nixos mainnet box with Zeus, works great!

@daywalker90 I would add docs for the config in the README, especially `clnrest-port`, which the plugin cannot start without.

ACK https://github.com/lightningd/plugins/pull/598/commits/09baba5b2139fde12d8c503b4e715e0c76e8e7bc

chrisguida:
@cdecker @m-schmoock @vincenzopalazzo merging this soon unless I hear otherwise! I think it's a fantastic plugin.

PR #626, opened 2024-12-10
CI: don't upgrade generic env's python dependencies

daywalker90:
We don't upgrade python dependencies in the other types (python/poetry). This would cause the generic type to upgrade `pyln-testing` beyond the tested CLN version and `pyln-testing` is not always backwards compatible.

Example run with fix: https://github.com/daywalker90/plugins/actions/runs/12252218795

Issue reported by @chrisguida on discord


daywalker90:
The failing tests on cln 24.02 is a different issue for clnrest-rs

daywalker90:
Same with torq. Both require pyln-testing >=24.05.

chrisguida (approved):
Excellent, thanks for the fix!

PR #630, opened 2024-12-16
clnrest-rs: update to v0.2.2

daywalker90:
### Fixed

- wait for other plugins to generate certs

### Changed
- logging: added offending method name if verify_rune fails
- upgraded dependencies


chrisguida (approved):
Looks good!

PR #632, opened 2024-12-16
holdinvoice: update to v3.1.1

daywalker90:
### Changed

- upgrade dependencies

Also fixes the test failures with 24.11 by using decode instead of now deprecated decodepay.


chrisguida (approved):
LGTM!

PR #633, opened 2024-12-16
sling: update to v3.0.0

daywalker90:
### Added

- ``sling-inform-layers``: CLN 24.11+ will inform these layers about sling's liquidity information gained by failed rebalances. Can be stated multiple times. Defaults to ``xpay`` which will help the ``xpay`` command.

### Changed

- ``sling-reset-liquidity-interval`` defaults to 60m now and checks every 2m (previously: 360m/10m). This is to be in sync with the aging of ``xpay``'s layer

### Fixed

- ``sling-refresh-peers-interval`` was not actually dynamic
- ``sling-refresh-aliasmap-interval`` was not actually dynamic
- ``sling-refresh-gossmap-interval`` was not actually dynamic
- ``sling-reset-liquidity-interval`` was not actually dynamic

### Removed

- ``sling-utf8``, it was not used

### Fixed
- rebalances between private channels would fail after the first if only private channels are candidates


chrisguida (approved):
Looks great!

PR #634, opened 2024-12-16
summars: update to v4.0.1

daywalker90:
### Fixed

- pays: CLN 24.11+: payments that were started (but not yet finished) earlier than the first payment shown by summars would not show up once paid
- invoices: invoices created before the first invoice shown by summars would not show up once paid
- forwards: offered and not yet settled forwards that were created before the first forward shown by summars would not show up once settled

### Added

- pays: CLN 24.11+ will use newly added indexing in ``listpays`` to speed up building the pays table on subsequent summars calls

### Changed

- pays: summars can't guarantee that it can find/calculate ``sats_requested``,``msats_requested`` and therefore also can't guarantee ``fee_sats`` and ``fee_msats``. If this is the case these will be shown as ``N/A`` and not included in the totals summary at the end and in json mode the fields will be omitted
- pays: summars can't guarantee that it can find the ``destination``. If this is the case it will be shown as ``N/A`` and in json mode this field will be omitted
- pays: description field now included in json output mode if a description was found

### Fixed

- no longer panic on missing payment ``destination`` or ``amount_msat``

### Added

- Total values summary for forwards, pays and invoices table at the bottom in normal view and as ``totals`` object in json mode
- Show timeframe and limits you are currently using for each table of forwards, pays and invoices
- New options to limit output of forwards, pays and invoices tables: ``summars-forwards-limit``, ``summars-pays-limit`` and ``summars-invoices-limit``. Defaults to 0 (off) and limits the outputs to the last x entries, useful if you are setting high time values for ``summars-forwards``, ``summars-pays`` or ``summars-invoices``
- New ``summars-forwards-columns`` column: ``eff_fee_ppm`` showing the effective fee ppm


chrisguida (approved):
Very cool

PR #635, opened 2024-12-21
CI: we decide the pyln-testing version to be installed

daywalker90:
We can't let plugins dictate which version of pyln-testing will be installed for testing. Due to missing backward/forward compatibility and no way for plugin repos to tell which CLN versions they are being tested on we must set the correct pyln-testing version. For this i added/moved this to the last step of setting up the pyton environments.


daywalker90:
This fixes a few tests currently failing for 24.11

chrisguida:
Cool, thanks for catching that!

chrisguida (approved):
LGTM! Thanks!!

PR #636, opened 2024-12-21
datastore tests: handle cln v24.11+ default non-empty listdatastore

daywalker90:
By default listdatastore in cln 24.11 contains an "askrene" object. So i changed these lines to only check for inclusion and not equalness.


chrisguida:
Nice catch!

chrisguida (approved):
ACK

PR #640, opened 2025-01-03
CI: install pyln-testing before running generic setup

daywalker90:
In the generic setup the python dependencies installation should be done before custom code from the setup.sh is executed. Since `install_dev_pyln_testing` installs things like `protobuf` the version can change between generating grpc bindings and running the tests which causes errors (see holdinvoice erroring on nightly but not 24.11)


chrisguida (approved):
LGTM, thanks for catching!

PR #641, opened 2025-01-03
donations: upgrade dependencies

daywalker90:
The recent PR to install pyln-testing after a plugin's dependencies causes issue for this plugin because some dependencies were incompatible with the cln dev env we install with pyln-testing. The easiest fix for me is to just update the dependencies of donations. They were updated in a way that they are still compatible with the python versions we test.


chrisguida (approved):
Looks good!

PR #642, opened 2025-01-06
Poetry deprecation fixes

daywalker90:
I saw some fresh problems in the CI with poetry in python 3.11:

```
The command "export" does not exist.
```

```
Warning: The current project could not be installed: No file/folder found for package cln-meta-project
If you do not want to install the current project use --no-root.
```

I remembered this deprecation warning:

```
Warning: poetry-plugin-export will not be installed by default in a future version of Poetry.
In order to avoid a breaking change and make your automation forward-compatible, please install poetry-plugin-export explicitly. See https://python-poetry.org/docs/plugins/#using-plugins for details on how to install a plugin.
```

So i added the dependency manually aswell as the --no-root. Let's hope it doesn't break the other versions.

### DIFF
--- a/.ci/test.py
+++ b/.ci/test.py
@@ -16,6 +16,7 @@ global_dependencies = [
     "pytest",
     "pytest-xdist",
     "pytest-timeout",
+    "poetry-plugin-export",
 ]
 
 pip_opts = ["-qq"]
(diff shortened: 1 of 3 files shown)


chrisguida (approved):
Thanks!

chrisguida:
Maybe we could update to poetry 2.0? https://python-poetry.org/blog/announcing-poetry-2.0.0/

PR #651, opened 2025-02-23
currencyrate: switch to new coindesk api

daywalker90:
The old coindesk api is no longer working. I'm not sure this is the best index but the values seem in line with the other sources.


chrisguida:
This appears to fix the CI failure [here](https://github.com/lightningd/plugins/actions/runs/13535595835/job/37826605809#step:7:1563) on line 1560:

`pyln.client.lightning.RpcError: RPC call failed: method: currencyrates, payload: ['USD'], error: {'code': -32600, 'message': 'Error while processing currencyrates: HTTPSConnectionPool(host=\'api.coindesk.com\', port=443): Max retries exceeded with url: /v1/bpi/currentprice/USD.json (Caused by NameResolutionError("<urllib3.connection.HTTPSConnection object at 0x7f9012cdaf00>: Failed to resolve \'api.coindesk.com\' ([Errno -5] No address associated with hostname)"))', `.

ACK https://github.com/lightningd/plugins/pull/651/commits/51c204602b3fd691df50ab342e0b692eefc8af11

Thanks for the fix @daywalker90 !!

PR #657, opened 2025-03-13
holdinvoice: update to v4.0.0

daywalker90:
### Changed

- differentiate between "soft" and "hard" expire. Previously holdinvoice would cancel HTLC's even if it was still possible to settle
- less usage of cln's datastore by not storing expiry value for ACCEPTED state and instead read it from plugin state
- after plugin start wait for 10s before processing rpc commands to give the plugin a chance to process HTLC's during a node restart
- upgraded dependencies

### Fixed

- use a constant amount of rpc connections instead of a proportional amount to the number of HTLC's held, fixes os error 11 crash when holding more HTLC's than the network connections limit


chrisguida (approved):
ACK

PR #666, opened 2025-04-08
payany: update to v0.2.4

daywalker90:
### Changed
- removed extra RRSIG query for BIP-353 since some DNS servers don't respond to them, instead rely on hickory's proof status alone

### Changed
- use DNS over HTTPS with included root certificates, so you won't get censored by your router or ISP

### Fixed
- don't use ANY dns query type, some servers refuse those, use specific ones instead
- support multiline TXT records

### Fixed
- don't panic on wrong user inputs


daywalker90:
Thanks @chrisguida for helping me test this on more real life systems!

chrisguida (approved):
I have finally run out of ways to break this thing!

ACK 05a5edf372253f21cb3c5d78b150ef4e9100b616

PR #669, opened 2025-04-24
Historian update

endothermicdev:
This brings the historian plugin up to date:
- updates the dependencies of historian
- handles dns addresses in node announcements
- ignores the dying flag (closing channels that may in fact be in splice) in gossip store message headers
- filters out all private gossip types (for older CLN nodes)

It also adds the ability to publish to a message queue if the RABBITMQ_URL environment variable is set. (This is how Greenlight consumes gossip, so it's really useful to have it built into the plugin.)

### DIFF
--- a/archived/historian/historian.py
+++ b/archived/historian/historian.py
@@ -297,11 +297,16 @@ class Flusher(Thread):
                          gossipd.ChannelAnnouncement,
                          gossipd.NodeAnnouncement]:
             if isinstance(msg, msg_type):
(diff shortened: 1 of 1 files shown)


chrisguida:
@endothermicdev great!! So to clarify, I should treat you as the new maintainer, correct? (The person who unarchives the plugin is the default maintainer on all the other plugins).

endothermicdev:
Yep, sounds good.

PR #670, opened 2025-04-26
rebalance.py: Fixing setup_routing_fees() and peer_from_scid() for unannounced channels

21M4TW:
rebalance.py did not work with unannounced channels, as the setup_routing_fees() and peer_from_scid() functions were not using listpeerchannels for local channels. This PR fixes these issues.


chrisguida:
@gallizoltan please review

cc @m-schmoock @cdecker

gallizoltan (approved):
LGTM, but untested.
Yeah, I never really thought about private channels before. 😅

PR #672, opened 2025-04-28
Add new plugin bumpchannelopen

ca-ruz:
A Core Lightning plugin to bump an open channel transaction by creating a CPFP, useful when the funding transaction is stuck in the mempool or has a low fee.

https://github.com/ca-ruz/bumpchannelopen


chrisguida:
@ca-ruz looks great so far! I think the testing around the emergency reserve could be improved.. there is an edge case where the user creates a channel and then their change output becomes the emergency reserve, and if they bump this output with too high a fee, the emergency reserve could end up getting inadvertently spent. Please either add a very loud warning about this in the README and in the RPC help strings, or add some logic that removes this footgun.

PR #678, opened 2025-07-07
payany: upgrade to v0.3.0

daywalker90:
### Removed
- :warning: ``payany-dns``: option removed (you have to remove it from your config if you have set it!) in favor of trying them all one by one and using CLN's `proxy` if it is configured to `always-use-proxy`

### Changed
- All lookups are proxied through CLN's `proxy` if it is configured and `always-use-proxy` is set to `true`
- If one DNS fails the next one is tried instead of immediately giving up

### Added
- Explicit timeout of 30s for all lookups

### Fixed
- ``payany-xpay-handle-pay``: wallets sending ``maxfeepercent`` as a string now work as expected


chrisguida (approved):
LGTM

ACK https://github.com/lightningd/plugins/pull/678/commits/da5d41c18b00c610f029bc73deb93c6839a3709c

There are some flaky tests due to GH ratelimiting its API, @daywalker90 is working on a workaround.

PR #686, opened 2025-07-23
summars,payany: remove github api usage in test setup

daywalker90:
This should fix the flakes.


chrisguida (approved):
Tests are passing! Let's hope they keep passing :crossed_fingers: 

ACK

PR #687, opened 2025-07-24
Unarchive historian

endothermicdev:
This mostly just moves historian out of the archive directory, but I also fixed a minor bug for the greenlight use case.  Finally, I'm trying to add support for the uv package manager while maintaining compatibility with poetry.  If all goes well, the intent is to migrate fully to uv, but for now I'd like to allow testing without breaking existing installation methods (reckless is getting a uv installer but existing nodes won't have that yet.)


chrisguida (approved):
Awesome, thanks so much for submitting this, @endothermicdev!

Psyched to see `historian` rise again :rocket: 

ACK https://github.com/lightningd/plugins/pull/687/commits/b839311a917c7d97eb7daed4421fcf83498bf200

PR #693, opened 2025-08-11
CI: switch nightly to uv and add python 3.13

daywalker90:
CLN master has switched to uv instead of poetry. It also supports python 3.13 so lets add it aswell.


daywalker90:
You can inspect a run of this PR here: https://github.com/daywalker90/plugins/actions/runs/16880249309

Failures come from falke in bolt12-prism and for python3.13 the torq-plugin needs to fix its dependencies.

cdecker (approved):

chrisguida (approved):
Thanks for the fix!!

ACK

chrisguida:
Link to broken CI run https://github.com/lightningd/plugins/actions/runs/16870448412

Link to an explanation as to why `uv` is cool https://github.com/ElementsProject/lightning/issues/7665#issuecomment-2419044125

PR #644, opened 2025-01-28
add smaug plugin

chrisguida:
This plugin sends bkpr-compatible events to the bkpr plugin for external on-chain wallet movements. This allows bkpr to maintain a full picture of a merchant's bitcoin inflows and outflows, vastly improving bitcoin accounting and financial analysis.

PR #694, opened 2025-08-21
CI: update main and add uv support

daywalker90:
- Support uv projects, refactored the explicit tmp directory creation to not run for uv projects
- Update python 3.12 to 3.13 in the matrix (torq-plugin needs to update dependencies for 3.13)
- Detect poetry and uv projects by their .lock files to differentiate them (if both are present due to migration prioritize uv)
- Don't check out lightning repo and pull the published pyln-* packages instead with the matching version to CLN
- Save cache more often, they get deleted after a week anyways and it speeds up PR runs

### DIFF
--- a/.ci/test.py
+++ b/.ci/test.py
@@ -1,12 +1,15 @@
 import logging
 import os
+import re
 import subprocess
 import sys
 import tempfile
 import time
+import shutil
 import json
 from itertools import chain
 from pathlib import Path
+from typing import Tuple
 
 from utils import Plugin, configure_git, enumerate_plugins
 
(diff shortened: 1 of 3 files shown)


daywalker90:
clnrest-rs sometimes runs very long i don't know why. The timeout then exceeds the action timeout, and the job gets cancelled. We will archive clnrest-rs with v25.09 anyways so that's how i'd fix it.

chrisguida:
To add some context:

We used to explicitly detect Rust plugins in order to install them using setup.sh and avoid attempting to install the plugin itself using poetry if the repo had poetry tests. This functionality was added in #644.

This PR changes the flow a bit. Now we just install the plugin using setup.sh if it exists, and use the detected framework just for the tests.

To summarize the new flow:

1) we try to detect what framework the plugin is by looking at various project files.
2) we install the plugin using its setup.sh file if it exists, and based on the framework we found if not.
3) we run the tests based on the detected framework

In the logs, we list the plugins based on the detected framework. For now, this puts Smaug into the "poetry" category because its tests are poetry, but the plugin itself is Rust and uses a setup.sh to install.

daywalker90:
The rust plugin detection was just this extra log line:
`Rust plugins: c-lightning-pruning-plugin, cln-nip47, cln-ntfy, clnaddress, clnrest-rs, clnrod, consolidator, hold, holdinvoice, payany, sling, smaug, summars, watchtower-client`
There never was a rust specific category they were always in "generic".

setup.sh was previously only ran on generic plugins now that rust plugins get sorted by their test environment we run setup.sh whenever we detect it regardless of rust/test framework.

In the logs the plugins are listed by test framework, yes.

PR #696, opened 2025-09-08
datastore: archive obsolete plugin

daywalker90:
datastore is part of CLN for a very long time now, so there is no point to keep it in the active plugins directory imo


chrisguida (approved):
Yep, was thinking of doing this myself!

ACK

PR #775, opened 2026-01-19
sauron: add cache and global fetch lock

daywalker90:
In an effort to get the tests passing again i had to either mock the esplora responses or make the plugin and test framework respect esplora's rate limit.

I decided to do the latter and add a global file based cache to reduce requests for the same URL in the cache's TTL.

The global fetch lock is needed for the race condition of same URL requests that happen often during multiple CLN node startups.

Please let me know what you think.

I'm tagging some of the original sauron authors (looking at the commit history):
@darosior  @Amperstrand @chrisguida @russnes @ca-ruz @iorch @sip-21

### DIFF
--- a/sauron/pyproject.toml
+++ b/sauron/pyproject.toml
@@ -5,7 +5,11 @@ description = "A Bitcoin backend plugin relying on Esplora"
 readme = "README.md"
 requires-python = ">=3.9.2"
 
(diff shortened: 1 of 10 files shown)


daywalker90:
No review in 4 months so i will be archiving sauron instead: https://github.com/lightningd/plugins/pull/841

PR #797, opened 2026-04-27
add lnradar v0.1.1

Lagrang3:
I would like to add a link to `lnradar` a plugin to probe the network for liquidity allocation.
It can be used to improve payment reliability.
Still beta but usable.


daywalker90:
I added the missing referece+link at the bottom of the markdown files.

PR #845, opened 2026-05-15
bumpit: update submodule for CLN 25.12 newaddr fix

ca-ruz:
Updates the bumpit submodule to include the CLN 25.12+ fix for `newaddr`.

CLN 25.12 changed the default address type returned by `newaddr`, so bumpit now requests `bech32` explicitly before reading the `bech32` field.


daywalker90:
LGTM, btw python 3.9 is EOL: https://devguide.python.org/versions/ so you might want to bump that to 3.10

PR #888, opened 2026-06-28
rebalance: handle channels without short_channel_id

bittylicious:
The rebalance plugin can crash in peer_from_scid when listpeerchannels returns entries without a short_channel_id, such as channels that are not yet announced/locked in.

This changes the lookup to use ch.get("short_channel_id") before comparing, matching the defensive pattern already used elsewhere in the plugin.

Observed against Core Lightning v26.06 while running an explicit rebalance; the RPC failed with KeyError: 'short_channel_id'.

Validation:\n- python3 -m py_compile rebalance/rebalance.py rebalance/test_rebalance.py
- Applied equivalent one-line fix on a running node and confirmed explicit rebalance commands complete successfully.


daywalker90:
Good catch , thank you @bittylicious !

PR #891, opened 2026-07-20
rebalance: support custom CLN version strings

bittylicious:
## Summary

- accept custom Core Lightning build identifiers that are not dotted release numbers
- infer the legacy or modern `getroutes` schema from the returned route fields
- preserve the minimum-version check for parseable release versions

## Background

Core Lightning exposes its build identifier verbatim through `getinfo.version`, so custom builds may return a branch or package name instead of a release such as `v26.06.2`. The rebalance plugin currently parses that value as dotted integers during initialization, which raises `ValueError` and leaves the plugin only partially initialized.

This change treats an unparseable identifier as an unknown release version. Route handling no longer needs that release number: it detects the pre-26.06 or 26.06+ schema from the actual `getroutes` path. Parseable versions older than `v25.09` are still rejected as before.

## Compatibility

The repository CI matrix runs the rebalance integration tests against CLN `25.12`, `26.04`, and `26.06.1` on Python 3.10 and 3.12, covering both route schemas. An opaque build identifier necessarily skips only the startup release-number check; unavailable RPCs still fail normally when called.

## Tests

- `pytest -q rebalance/test_clnutils.py` (`7 passed`)
- `ruff check` and `ruff format --check`
- `python3 -m py_compile`
- `git diff --check`

### DIFF
--- a/rebalance/clnutils.py
+++ b/rebalance/clnutils.py
@@ -9,6 +9,9 @@ def cln_parse_rpcversion(string):
     to ubuntu style with version 22.11  `yy.mm[.patch][-mod]`
     make sure we can read all of them for (the next 80 years).
     """
+    if not isinstance(string, str):
+        return None
+
     rpcversion = string
     if rpcversion.startswith("v"):  # strip leading 'v'
         rpcversion = rpcversion[1:]
@@ -19,5 +22,10 @@ def cln_parse_rpcversion(string):
     if rpcversion.count(".") == 1:  # imply patch version 0 if not given
         rpcversion = rpcversion + ".0"
 
-    # split and convert numeric string parts to actual integers
-    return list(map(int, rpcversion.split(".")))
+    # Custom builds may use a branch or package name instead of a release.
+    try:
+        version = list(map(int, rpcversion.split(".")))
+    except ValueError:
+        return None
+
+    return version if len(version) >= 2 else None
--- a/rebalance/rebalance.py
+++ b/rebalance/rebalance.py
@@ -36,76 +36,52 @@ def get_thread_id_str():
     return f"{plugin.threadids.get(threading.get_ident(), 0):{'0' + str(max_digits)}}"
 
 
-# The route msat helpers are needed because older versions of cln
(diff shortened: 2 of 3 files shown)


bittylicious:
A real-world data point: this affected us in production, rather than being a theoretical compatibility issue. Our live CLN node reports a tweaked version string because we carry a few local customisations. That caused the rebalance plugin to fail during initialisation when it tried to parse the version as dotted integers. We have run this change on the live node and confirmed that it restores normal plugin startup and operation, while ordinary release version strings still retain the minimum-version check.

daywalker90:
Oof. Does your version string atleast include the original version string somewhere? Half my plugins use that to determine features...

bittylicious:
No, unfortunately it did not at the time. `getinfo.version` was literally `opencode/spenderp-defer-awaiting-recovery`, with no upstream release embedded. You are right: that is unfriendly to plugin feature detection, and our build metadata should have preserved the original version.

We have since corrected that side too. The live node is still running our customised image line, but the binary now reports `v26.06.6`. I see this patch as a defensive fallback for unexpected/custom strings, not as an endorsement of replacing the upstream version with an opaque label.

bittylicious:
It looks like there is already a de facto convention: normal builds use git describe, which preserves the nearest upstream tag in values such as v26.06.6-1-gabcdef0, and the existing parser deliberately ignores the hyphenated suffix. Our earlier v26.06.6-custom values therefore worked correctly.
 
I could not find a formally documented or enforced grammar for custom VERSION overrides, though: the build accepts arbitrary values and getinfo.version is specified only as a string. So following the existing parseable-prefix convention is the primary fix for custom builds, while this PR provides a defensive fallback so an unexpected value warns and uses response-schema detection rather than crashing the plugin.

Your call really - it does handle what most people do in reality, but the right answer might be some documentation detailing the version string (or it could be this PR, or both).

daywalker90 (approved):

daywalker90:
Alright, thank you for the changes, lgtm

PR #953, opened 2026-09-15
bolt12-prism: archive plugin

daywalker90:
@gudnuf

I want to archive this plugin since all tests are failing in CI and the PR with a fix has no response in almost 2 months.


gudnuf:
go for it @daywalker90 I have no plans on maintaining.

